PR #28, opened 2019-09-05
Read MAC address from EEPROM

### LINKED ISSUE #23: Read MAC address from EEPROM
The MAC address should be read from the EEPROM, which has built in EUI48.

I need this and am happy to implement it.
My initial plan would be to print the MAC address found on startup via semihosting - this requires the user have an STLink. Another option would be to dump the MAC address over the debug UART, which comes out to a header. 

@jordens any preferences on how to get the MAC address out?

cjbe:
Try and read the unique MAC address from EEPROM.
If the EEPROM can not be read (e.g. on the unpatched hardware version 1.0 where the EEPROM is unpowered) a hardcoded default of 10:E2:D5:00:03:00 is used.

The MAC address found / used is logged. At the moment the only way of accessing this log is via semihosting.

Closes #23


jordens:
Why not just use the hal? The only reason I didn't use it was that it didn't exist yet.

cjbe:
@jordens the HAL I looked at (https://github.com/stm32-rs/stm32h7xx-hal) is in pretty rough shape - it does not handle I2C NACKs / retries yet, so locks up if the EEPROM is busy or not present.

I originally planned to use the HAL, and improve its I2C implementation, but I could not get the HAL to work with RTFM (I only tried briefly).

src/main.rs
```diff
@@ -611,21 +629,34 @@ const APP: () = {
...
+        let hardware_addr = match eeprom::read_eui48(c.resources.i2c) {
+            Err(_) => {
+                info!("Could not read EEPROM, using default MAC address");
+                net::wire::EthernetAddress([0x10, 0xE2, 0xD5, 0x00, 0x03, 0x00])
+            },
+            Ok(raw_mac) => net::wire::EthernetAddress(raw_mac)
+        };
+        info!("MAC: {}", hardware_addr);
+
         let hardware_addr = net::wire::EthernetAddress([0x10, 0xE2, 0xD5, 0x00, 0x03, 0x00]);
```
jordens: I guess you didn't test this because you'd have to hack your hardware. But with this line it won't work.

cjbe: Oops - I did test this on my hardware, but I must have screwed up the final rebase.

cjbe:
@jordens I propose merging this now as a stopgap - the code has no extra dependencies, so should require negligible maintenance (which I am happy to do). When I have time I will upstream a more robust I2C driver to stm32h7-hal and move Stabilizer over to this.

PR #49, opened 2019-11-12
fix memory safety issue in ethernet interface (closes #33)

### LINKED ISSUE #33: Ping requests only responded to after two requests have arrived
Running the current master firmware, Stabilizer seems to only respond to ping requests after two requests have arrived. Here the default ping period is 1s, so the first ping request is only responded to after the second has been sent, etc.
```
$ ping 10.255.6.56
PING 10.255.6.56 (10.255.6.56) 56(84) bytes of data.
64 bytes from 10.255.6.56: icmp_seq=1 ttl=64 time=1026 
64 bytes from 10.255.6.56: icmp_seq=2 ttl=64 time=0.303 
64 bytes from 10.255.6.56: icmp_seq=3 ttl=64 time=1015 
64 bytes from 10.255.6.56: icmp_seq=4 ttl=64 time=0.296 
64 bytes from 10.255.6.56: icmp_seq=5 ttl=64 time=1000 
64 bytes from 10.255.6.56: icmp_seq=6 ttl=64 time=0.411 ms
```
This can be clearly seen in [this packet dump](https://github.com/quartiq/stabilizer/files/3597890/dump.zip).

This is not problem in itself, but suggests that something in the ethernet interface is not quite right.

cjbe:
The CPU is allowed to implement normal memory writes out-of-order. Here
the write to the OWN flag in the DMA descriptor (normal memory) was
placed after the DMA tail pointer advance (in device memory, so not
reorderable). This meant the ethernet DMA engine stalled as it saw
a descriptor it did not own, and only restarted and sent the packet
when the next packet was released.

This fix will work as long as the CPU data cache is disabled. If we
want to enable the cache, the simplest method would be to mark SRAM3
as uncacheable via the MPU.

### DIFF
--- a/src/eth.rs
+++ b/src/eth.rs
@@ -245,7 +245,13 @@ impl RxRing {
 
         let addr = &self.desc_buf[self.cur_desc] as *const _ as u32;
         assert_eq!(addr & 0x3, 0);
+
         let dma = unsafe { pac::Peripherals::steal().ETHERNET_DMA };
+
+        // Ensure changes to the descriptor (in particular, the OWN flag) are
+        // committed before DMA engine sees tail pointer store.
+        cortex_m::asm::dsb();
+
         dma.dmacrx_dtpr.write(|w| unsafe { w.bits(addr) });
 
         self.cur_desc = self.next_desc();
@@ -314,7 +320,13 @@ impl TxRing {
 
         let addr = &self.desc_buf[self.cur_desc] as *const _ as u32;
         assert_eq!(addr & 0x3, 0);
+
         let dma = unsafe { pac::Peripherals::steal().ETHERNET_DMA };
+
+        // Ensure packet contents as well as changes to the descriptor have been
+        // committed before DMA engine sees the tail pointer store.
+        cortex_m::asm::dsb();
+
         dma.dmactx_dtpr.write(|w| unsafe { w.bits(addr) });
     }
 }
@@ -489,6 +501,10 @@ impl Device {
                 .te().bit(true) // Transmiter Enable
         });
         eth_mtl.mtltx_qomr.modify(|_, w| w.ftq().set_bit());
+
+        // Ensure ring buffer descriptors have been set up in memory before
+        // enabling DMA engine.
+        cortex_m::asm::dsb();
 
         // Manage DMA transmission and reception
         eth_dma.dmactx_cr.modify(|_, w| w.st().set_bit());


src/eth.rs
```diff
@@ -241,11 +241,15 @@ impl RxRing {
     fn buf_release(&mut self) {
         let addr = &self.pkt_buf[self.cur_desc] as *const _;
         self.desc_buf[self.cur_desc][0] = addr as u32 & EMAC_DES0_BUF1AP;
+        // Ensure change to desc0 occurs before setting desc3 OWN flag
+        cortex_m::asm::dmb();
```
dnadlinger: Not sure this is required as long as the tail pointer is only updated once both are written.

cjbe: Yep - you are right.

src/eth.rs
```diff
@@ -315,6 +317,8 @@ impl TxRing {
         let addr = &self.desc_buf[self.cur_desc] as *const _ as u32;
         assert_eq!(addr & 0x3, 0);
         let dma = unsafe { pac::Peripherals::steal().ETHERNET_DMA };
+        // Ensure change to desc3 OWN flag occurs before tail pointer adjustment
```
dnadlinger: To be picky, I think it's not just the OWN flag, but also the packet buffer contents. (It is implied from context, i.e. source code order, though.)

src/eth.rs
```diff
@@ -482,6 +486,7 @@ impl Device {
 
         self.rx.init(eth_dma);
         self.tx.init(eth_dma);
+        cortex_m::asm::dsb();
```
jordens: That barrier should be 10 lines further down before the DMAC[RT]X_CR enables to indicate the deadline.

cjbe:
> Before enabling the data cache, some thought should go into replacing `DSB` with less brutal synchronization constructs.

I am not sure there is any way around this, as the problem is that the ethernet peripheral address range has the 'device' attribute, and the RAM typically has a 'normal memory' attribute (hence reorderable). As I understand it the options we have are:
1) No DCACHE for SRAM3, 'normal memory' type attribute (as we have now, or via marking SRAM3 as uncachable, then enabling cache)
2) Enable DCACHE for SRAM3, but invalidate/clean cache at appropriate times. Then a DSB is required as part of the standard ARM cache sequence.
3) Mark SRAM3 as device type - this forces the CPU to respect order, but will reduce efficiency.

For (1) the performance hit comes at the DSB which forces a couple of memory accesses to complete before hitting the next instruction.
For (2) the performance may be a little better, but there is still a DSB in the cache invalidate ([see here](https://docs.rs/cortex-m/0.6.1/src/cortex_m/peripheral/scb.rs.html#485-509)).
For (3) I expect the performance hit to be pretty brutal, and has much the same effect as having DSBs all over the SRAM3 accesses.

dnadlinger:
> Before enabling the data cache, some thought should go into replacing DSB with less brutal synchronization constructs.

But this isn't something to worry about here, is it? Right now the dcache is disabled, and remains so after this PR, but even so you need a fence because of store reordering. (A non-trivial amount of extra attention would be required to make the implementation compatible with caching.)

As Chris points out, another way to address this problem _might_ be to configure the SRAM3 region used for network buffers as write-through/non-cachable according to the link I posted in #33. This _might_ allow us to downgrade the `atomic::fence`s to `atomic::compiler_fence`s – that is, if my understanding of the Cortex-M7 memory subsystem is correct –, but I wouldn't want to bet on this without seeing more explicit documentation first.

@cjbe: I noticed you are using `dsb`, whereas I was using `atomic::fence(Ordering::SeqCst)`, which I would expect to generate a `dmb` (didn't check). The latter appeared to work, but it's not clear to me whether the definition of observability would extend to the DMA peripheral. Do you know what the actual behaviour guarantees are here? I found [an ARM document which shows DMB being used to set up DMA buffers](http://infocenter.arm.com/help/topic/com.arm.doc.genc007826/Barrier_Litmus_Tests_and_Cookbook_A08.pdf), but also [an LKML thread suggesting that DSB might be required](http://lists.infradead.org/pipermail/linux-arm-kernel/2011-April/047708.html). I'd suggest just leaving it a DSB for now.

By the way, the Rust Embedonomicon seems to be wrong in suggesting that barriers are only required if the dcache is used here: https://docs.rust-embedded.org/embedonomicon/dma.html#dont-we-need-a-memory-barrier

cjbe:
> As Chris points out, another way to address this problem might be to configure the SRAM3 region used for network buffers as write-through/non-cachable

I don't think this is the case - this is still a 'memory' type region, so the CPU can reorder the accesses past other accesses (such as the 'device' type DMA tail pointers)

cjbe:
> @cjbe: I noticed you are using dsb, whereas I was using atomic::fence(Ordering::SeqCst), which I would expect to generate a dmb (didn't check).

I agree - for ARM7M the DMB should /probably/ work, as (per the architecture reference manual) the ARM7M "only supports system-wide barriers with no Shareability domain or access type limitations", so all LDR / STR's should be completed by the barrier, but I am not fully confident on this.

src/eth.rs
jordens: That barrier should be 10 lines further down before the DMAC[RT]X_CR enables to indicate the deadline.

jordens:
> But this isn't something to worry about here, is it? Right now the dcache is disabled, and remains so after this PR, but even so you need a fence because of store reordering. (A non-trivial amount of extra attention would be required to make the implementation compatible with caching.)

Yes. I only want to ensure that this gets solved comprehensively in that case.

jordens:
re `DSB/DMB/SeqCst`: I'm also unsure as to the line between ordering across different buses and memories/mmaped regions vs "merely" instructing the compiler to not reorder access instructions.

dnadlinger:
@jordens: I moved the barrier in init as you requested, although I didn't mind Chris's location choice either (before any hardware register setup seemed just as clean).

---

> re DSB/DMB/SeqCst: I'm also unsure as to the line between ordering across different buses and memories/mmaped regions vs "merely" instructing the compiler to not reorder access instructions.

The first half of this is a bit tricky to answer; the interaction of DMB with the various buses seems underdocumented. Thankfully, though, it shouldn't really matter here; I can't really see DSB becoming a performance problem on the Cortex-M7, but we can always have a closer look at DMB if it ever does.

Regarding atomic barriers and compiler reordering, Rust more or less just imported the memory model from C++11 (which isn't a bad idea), so I'm happy to help clear up any questions regarding that (and LLVM optimisations…). The wrinkle here is just that different peripherals/buses/… aren't really in the purview of the abstract memory model – or, equivalently, the hardware behaviour isn't specified in terms of its abstract machine.

A compiler barrier would have been enough here if Cortex-M7 didn't have a (non-trivial) store buffer – Cortex-M4 only had a one-entry write buffer, so the compiler barrier would have worked there (with the dcache disabled).

PR #51, opened 2019-11-13
factor CPU and board-specific initialisation out of main

cjbe:


jordens:
Just to confirm: Is this a 1:1 move?

cjbe:
Yes - there is no functional change. I will amend the commit message to make this clear

jordens:
It's ok
bors r+

jordens:
This wasn't a 1:1 move. You converted the entire thing to dos line endings and shuffled the log setup.

PR #62, opened 2019-11-23
setup TIM2 after interrupts enabled (fixes #55)

### LINKED ISSUE #55: Ch1 IIR state diagnostics not updating
Using the current master, the IIR state diagnostics for ch1 (broadcast on port 1234) is not updating, instead it has a fixed constant (plausible) value. The diagnostics for ch0 are correctly updating.

I see this both running via a debug session, and after booting from flash, with both the nightly and stable compiler.
(c.f. #32)

cjbe:
If TIM2 is setup before interrupts are enabled, the first interrupt can be
serviced late, leading to the sampling of one or both channels halting

### DIFF
--- a/src/board.rs
+++ b/src/board.rs
@@ -522,11 +522,19 @@ pub fn init() {
     let dbgmcu = dp.DBGMCU;
     dbgmcu.apb1lfz1.modify(|_, w| w.tim2().set_bit());
 
-    tim2_setup(&dp.TIM2);
-
     let i2c2 = dp.I2C2;
     i2c::setup(&rcc, &i2c2);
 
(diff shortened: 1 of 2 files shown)


jordens:
This was triggered by your board refactoring and adding the I2C setup late.

Don't move the entire configuration of the timer out of `init()`. `Peripheral::steal()` is unsafe for a reason. Just configure and enable the timer late enough: Move `tim2_setup` to the end of `init()` and add a comment to ensures that it stays there.

A couple other things that slipped through the review:

* `init_log()` should be as early as possible (as it was before).

jordens:
Closing in favor of 1140b4f

PR #89, opened 2020-04-22
Stabilizer v1.1 Update

ryan-summers:
This PR introduces support for the Stabilizer v1.1 hardware. This PR is intended to replace https://github.com/quartiq/stabilizer/pull/89, but exclude any pounder-specific support.

This change set includes the following:
* Addition of AFE control for channels 1 and 2 (no control interface yet provided)
* Re-organizes repository into a Cargo workspace with `stm32h7xx-hal` and `stabilizer` crates
* Utilizes a custom build of the `stm32h7xx-hal` for device abstraction
* Utilizes the `stm32h7-ethernet` crate for ethernet support

Testing:
* ADC1, ADC2, DAC1, and DAC2 SPI CS signals were probed at 500KHz update configuration to verify that interrupt timing was measuring nominal and their was no resource contention.


src/main.rs
```diff
@@ -93,46 +102,299 @@ macro_rules! create_socket {
...
+            let config = hal::spi::Config::new(hal::spi::Mode{
+                    polarity: hal::spi::Polarity::IdleHigh,
+                    phase: hal::spi::Phase::CaptureOnSecondTransition,
+                })
+                .communication_mode(hal::spi::CommunicationMode::Transmitter)
+                .manage_cs()
+                .frame_size(16)
+                .swap_mosi_miso();
+
+            dp.SPI4.spi((spi_sck, spi_miso, hal::spi::NoMosi), config, 25.mhz(), &clocks)
```
jordens, line 230: The DACs can and should do 50 MHz (div2). As in the old firmware.
https://www.analog.com/media/en/technical-documentation/data-sheets/AD5512A_5542A.pdf

src/main.rs
```diff
@@ -93,46 +102,299 @@ macro_rules! create_socket {
...
+        // Start a SPI transaction on ADC0 and ADC1
+        c.resources.adc1.lock(|adc| adc.spi.cr1.modify(|_, w| w.cstart().set_bit()));
+        c.resources.adc2.lock(|adc| adc.spi.cr1.modify(|_, w| w.cstart().set_bit()));
+
+        c.resources.dbg_pin.set_low().unwrap();
+    }
+
+    #[task(binds = SPI2, resources = [adc1, dac1, adc2, dac2, iir_state, iir_ch, dac_pin], priority = 2)]
+    fn adc_spi(c: adc_spi::Context) {
+        #[cfg(feature = "bkpt")]
```
jordens, line 360: I used this for cycle counting (read out the cycle counter in gdb at each bkpt and take the difference, there is a little snippet in the gdb config for that).

src/main.rs
```diff
@@ -93,46 +102,299 @@ macro_rules! create_socket {
...
+
+        c.resources.dac_pin.set_high().unwrap();
+
+        let output_ch1 = {
+            let a: u16 = c.resources.adc1.read().unwrap();
+            let x0 = f32::from(a as i16);
+            let y0 = c.resources.iir_ch[0].update(&mut c.resources.iir_state[0], x0);
+            y0 as i16 as u16 ^ 0x8000
+        };
+        c.resources.adc1.spi.ifcr.write(|w| w.eotc().set_bit());
```
jordens, line 371: In the old firmware I did this very early (before the read+processing+write) because of the SPI silicon erratum.

src/main.rs
```diff
@@ -93,46 +102,299 @@ macro_rules! create_socket {
...
+
+        let output_ch2 = {
+            let a: u16 = nb::block!(c.resources.adc2.read()).unwrap();
+            let x0 = f32::from(a as i16);
+            let y0 = c.resources.iir_ch[1].update(&mut c.resources.iir_state[1], x0);
+            y0 as i16 as u16 ^ 0x8000
+        };
+        c.resources.adc2.spi.ifcr.write(|w| w.eotc().set_bit());
+
+        c.resources.dac1.send(output_ch1).unwrap();
```
jordens, line 381: This now has different latency adc-dac for both channels. Important to write to the DACs as early as possible.

src/main.rs
```diff
@@ -93,46 +102,299 @@ macro_rules! create_socket {
...
+//            ethernet_periph: (
+//                dp.ETHERNET_MAC,
+//                dp.ETHERNET_DMA,
+//                dp.ETHERNET_MTL,
+//            ),
+        }
+    }
+
+    #[task(binds = TIM2, resources = [dbg_pin, timer, adc1, adc2])]
+    fn tim2(mut c: tim2::Context) {
```
jordens, line 347: Note that this costs about 30-40 cycles (100 ns per servo cycle) compared to DMA. I.e. 5-10% more load. That's significant. OK for the transition phase but should be changed later.

src/main.rs
```diff
@@ -93,46 +102,299 @@ macro_rules! create_socket {
...
+        c.resources.dac_pin.set_low().unwrap();
+        #[cfg(feature = "bkpt")]
+        cortex_m::asm::bkpt();
+    }
+
+    #[idle]
+    fn idle(_c: idle::Context) -> ! {
+        // TODO Implement and poll ethernet interface.
+        loop {
+            asm::nop();
```
jordens, line 393: `wfe()`/`wfi()`?

jordens:
@cjbe @dnadlinger if you are interested, please feel free to review as well.

src/main.rs
jordens: The best solution would be to implement load monitoring (basically `top`) for the different tasks in RTFM. I think that's on their roadmap.

Cargo.toml
```diff
@@ -47,6 +53,7 @@ nightly = ["cortex-m/inline-asm"]
 [profile.dev]
 codegen-units = 1
 incremental = false
+opt-level = 3
```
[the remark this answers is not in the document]
ryan-summers: The size is likely small enough because all of the ethernet work is commented out

src/main.rs
```diff
@@ -93,46 +102,299 @@ macro_rules! create_socket {
...
+            .per_ck(100.mhz())
+            .pll2_p_ck(100.mhz())
+            .pll2_q_ck(100.mhz())
+            .freeze(vos, &dp.SYSCFG);
+
+        clocks.rb.rsr.write(|w| w.rmvf().set_bit());
+
+        clocks.rb.d2ccip1r.modify(|_, w| w.spi123sel().pll2_p().spi45sel().pll2_q());
+
+        let gpioa = dp.GPIOA.split(&mut clocks.ahb4);
```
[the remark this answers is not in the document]
ryan-summers: Ideally, yeah I think it should. It doesn't really make sense that we need to provide the specific AHB bus. I can update our HALs API. This was just using the existing HAL API.

src/main.rs
ryan-summers: Yeah, I currently have been using debug pin toggling to visualize all of the interrupts in relation to each other on a logic analyzer. I can characterize it without intrusive pin toggling as development nears completion.

src/main.rs
ryan-summers: This actually is done to make an effort to ensure the channel latency is nearly identical for each output. Because the sampling is triggered near instantaneously for each channel, the theory here is to generate the outputs simultaneously as well, which will result in a matched latency. If you update CH1 first, CH2 will have a latency of 400uS longer while it performs the IIR filter.

src/main.rs
ryan-summers: Yeah, there's a TODO in the original PR indicating conversion to utilizing DMA again. I was seeing some bizarre behavior at first, so I went for the simpler solution for the time being.

src/main.rs
jordens: Sure.
Lower latency (even on only one channel) has priority over matched latency.
The better solution for the future will be to space the two ADC xfers by about one IIR processing time so that they both have matched and low latency. (Would need two channels on that timer).
I think that "uS" should be "ns". And is it really that much? I seem to remember less.

src/main.rs
jordens: Yes. Good approach.

dnadlinger:
I am interested in placing the Stabilizer firmware on a solid footing, although at this point, I'm not sure I have much to add, except for a few general points:
 1. Supports for multiple different hardware configurations (bare, with our current sense board, with Pounder, etc.) will be critical, hopefully from one uniform code base. I suppose all the Pounder-specific things should be hidden behind a feature at the very least, although there might also be design implications from trying to make that ergonomic.
 2. Another thing to keep in mind is that even on that fairly fast CPU, we are not far from being throughput-limited, with two channels running at the design 1 MHz sample rate, possibly some more complex demodulation, etc., and full-rate network streaming (the latter barely works for one half-rate channel, although we haven't seriously optimised the code yet). I didn't necessarily see any non-zero-cost abstractions in the PR, but it's something to keep in mind.
 3. Following from 2., DMA is pretty essential, as Robert already pointed out.

In case you are curious, the latest firmware a student in our group used to get the magnetic field stabilisation (+feedforward) up and running is here: https://github.com/OxfordIonTrapGroup/stabilizer/tree/ff_on_fb_offset. The code quality is, to put it politely, a bit all over the place, as this is the result of a few people's first foray into embedded/Rust programming, but it might still be an illustration of what will need to be ported on top of whatever newly refactored framework.

dnadlinger:
Also cf. https://github.com/quartiq/stabilizer/pull/39.

jordens:
> 1. Supports for multiple different hardware configurations (bare, with our current sense board, with Pounder, etc.) will be critical, hopefully from one uniform code base. I suppose all the Pounder-specific things should be hidden behind a feature at the very least, although there might also be design implications from trying to make that ergonomic.

Building firmware that fits all these different hardware setups is not the goal here but it will help a lot if you are interested in working on that.
The nice thing about Pounder is that in the current basic functionality it is orthogonal to the rest of the Stabilizer functionality. I.e. you can use firmware for Stabilizer+Pounder on a board without Pounder just fine. I don't know whether this is true for your firmware.
Once we get to anything more complicated than the SISO IIR filter between ADC and DAC (i.e. feedback onto phase/frequency, MIMO loops, etc.) I'm unsure whether it will be possible to represent that in a single monolithic design. But in any case, this PR makes it easier, not harder.

> 2. Another thing to keep in mind is that even on that fairly fast CPU, we are not far from being throughput-limited, with two channels running at the design 1 MHz sample rate, possibly some more complex demodulation, etc., and full-rate network streaming (the latter barely works for one half-rate channel, although we haven't seriously optimised the code yet). I didn't necessarily see any non-zero-cost abstractions in the PR, but it's something to keep in mind.

In my measurements, IIRC, the CPU load at 500 kHz, two loops, was some 20-30%. 1 MHz would still be fine. And when I did full rate streaming for both ADCs and both DACs simultaneously at 500 kHz (~50 Mb/s) and both loops running at the same time, that wasn't anywhere near the CPU limit.

src/main.rs
ryan-summers: You are correct in the uS being nS. I'll delve more into the IIR filter time using the cycle counter later to verify

src/pounder/mod.rs
```diff
@@ -0,0 +1,186 @@
...
+const ATT_LE3_PIN: u8 = 8 + 3;
+
+pub struct QspiInterface {
+    pub qspi: hal::qspi::Qspi,
+}
+
+impl ad9959::Interface for QspiInterface {
+    type Error = Error;
+
+    fn configure_mode(&mut self, mode: ad9959::Mode) -> Result<(), Error> {
```
jordens, line 40: Only `FourBit` can be made to work reasonably. No need to bother with the other modes.
c.f.
https://github.com/quartiq/ad9959/commit/33a90e2e271cc4a5d96b2a4d19ea181bfc2be288

src/pounder/mod.rs
```diff
@@ -0,0 +1,186 @@
...
+
+        Ok(())
+    }
+}
+
+impl<DELAY> AttenuatorInterface for PounderDevices<DELAY>
+{
+    fn reset(&mut self) -> Result<(), Error> {
+        self.mcp23017.digital_write(ATT_RST_N_PIN, 1).map_err(|_| Error::I2c)?;
+        // TODO: Delay here.
```
jordens, line 134: We can assume (and document) that the second I2C write is slow enough.

src/pounder/mod.rs
```diff
@@ -0,0 +1,186 @@
...
+    fn latch(&mut self, channel: DdsChannel) -> Result<(), Error> {
+        let pin = match channel {
+            DdsChannel::Zero => ATT_LE0_PIN,
+            DdsChannel::One => ATT_LE1_PIN,
+            DdsChannel::Two => ATT_LE2_PIN,
+            DdsChannel::Three => ATT_LE3_PIN,
+        };
+
+        self.mcp23017.digital_write(pin, 1).map_err(|_| Error::I2c)?;
+        // TODO: Delay here.
```
jordens, line 149: We can assume (and document) that the second I2C write is slow enough.

src/pounder/rf_power.rs
```diff
@@ -0,0 +1,13 @@
...
+use super::InputChannel;
+
+pub trait PowerMeasurementInterface {
+    fn sample_converter(&mut self, channel: InputChannel) -> Result<f32, Error>;
+
+    fn measure_power(&mut self, channel: InputChannel) -> Result<f32, Error> {
+        let analog_measurement = self.sample_converter(channel)?;
+
+        // The AD8363 with VSET connected to VOUT provides an output voltage of 52mV / dB.
+        Ok(analog_measurement / 0.052)
```
jordens, line 11: Let's take the 51.7 mV/dB factor spec'd at 100 MHz here.

src/pounder/mod.rs
```diff
@@ -0,0 +1,186 @@
...
+            adc2,
+            adc1_in_p,
+            adc2_in_p,
+        };
+
+        // Configure power-on-default state for pounder. All LEDs are on, on-board oscillator
+        // selected, attenuators out of reset.
+        devices.mcp23017.write_gpioa(0xF).map_err(|_| Error::I2c)?;
+        devices.mcp23017.write_gpiob(1_u8.wrapping_shl(5)).map_err(|_| Error::I2c)?;
+        devices.mcp23017.all_pin_mode(mcp23017::PinMode::OUTPUT).map_err(|_| Error::I2c)?;
```
jordens, line 108: The mcp23017 crate really needs a couple of fixes.
I think [`all_pin_mode`](https://github.com/lucazulian/mcp23017/blob/09ff49d6b7ffd20d21d31430917aac922c130b05/src/lib.rs#L164) is broken. It set's just the LSB.
And `digital_write` and `_read` appear to be badly defined.
Also https://github.com/lucazulian/mcp23017/issues

src/pounder/mod.rs
```diff
@@ -0,0 +1,186 @@
...
+        self.mcp23017.digital_write(ATT_RST_N_PIN, 1).map_err(|_| Error::I2c)?;
+        // TODO: Delay here.
+        self.mcp23017.digital_write(ATT_RST_N_PIN, 0).map_err(|_| Error::I2c)?;
+
+        Ok(())
+    }
+
+    fn latch(&mut self, channel: DdsChannel) -> Result<(), Error> {
+        let pin = match channel {
+            DdsChannel::Zero => ATT_LE0_PIN,
```
jordens, line 142: The mapping attenuator to dds channel is not 1:1. DDS 0,1,2,3 correspond to Attenuators 1,0,3,2. Attenuator byte order in the daisy chain and latch mapping is 0,1,2,3.

src/afe.rs
```diff
@@ -0,0 +1,71 @@
...
+    {
+        let mut afe = Self { a0: a0, a1: a1};
+
+        afe.set_gain(Gain::G1);
+
+        afe
+    }
+
+    pub fn set_gain(&mut self, gain: Gain) {
+        match gain {
```
jordens, line 33: Since the Enum values already have the right bits, maybe this can be done without the match and with less code.
Same for `get_gain` below.

src/pounder/attenuators.rs
```diff
@@ -0,0 +1,46 @@
...
+use super::DdsChannel;
+
+pub trait AttenuatorInterface {
+    fn modify(&mut self, attenuation: f32, channel: DdsChannel) -> Result<f32, Error> {
+        if attenuation > 31.5 {
+            return Err(Error::Bounds);
+        }
+
+        // Calculate the attenuation code to program into the attenuator.
+        let attenuation_code = (attenuation * 2.0) as u8;
```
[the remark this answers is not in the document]
ryan-summers: Good catches here. I've implemented a fix to properly shift by 2 bits and invert the code.

src/pounder/mod.rs
ryan-summers: I agree that the crate is a bit clunky and has a few bugs present. I'll fork the repository and fix some bugs (and try to contribute them back to mainline).

ryan-summers: There's actually a PR already that cleans things up: https://github.com/mrd0ll4r/mcp23017
I'm changing our dependency to point to that PR for the time being.

src/afe.rs
ryan-summers: This can be done for `set_gain()`, but it would require the `TryFrom` trait for `get_gain()`, which is (unfortunately) a `std` component. This could definitely be cleaner, but is generally restricted from being so based on the API provided by the `embedded-hal`. Namely, we can't do something like `set_state(true|false)` and instead have to use `set_high()` and `set_low()`.

src/pounder/mod.rs
ryan-summers: Added a TODO to perform and document this measurement.

src/pounder/mod.rs
ryan-summers: Added a TODO to perform and document this measurement.

src/pounder/mod.rs
ryan-summers: I've reverted the referenced changes out of the `ad9959` driver and moved them into the Qspi interface implementation in this crate. This keeps the `ad9959` device-independant, since the problems mentioned in those comments is related to the STM32H7 we're using and not the AD9959.

src/pounder/mod.rs
```diff
@@ -183,21 +183,22 @@ where
                                     1_u8.wrapping_shl(5)).map_err(|_| Error::I2c)?;
         devices.mcp23017.all_pin_mode(mcp23017::PinMode::OUTPUT).map_err(|_| Error::I2c)?;
 
-        devices.select_onboard_clock()?;
+        // Select the on-board clock with a 5x prescaler (500MHz).
```
jordens, line 186: `s/prescaler/multiplier/g` throughout.

src/pounder/mod.rs
```diff
@@ -32,34 +32,109 @@ const ATT_LE3_PIN: u8 = 8 + 3;
...
     }
 
     fn read(&mut self, addr: u8, mut dest: &mut [u8]) -> Result<(), Error> {
         if (addr & 0x80) != 0 {
             return Err(Error::InvalidAddress);
         }
+
+        // It is not possible to read data from the AD9959 in single bit two wire mode because the
+        // QSPI interface assumes that data is always received on IO1.
+        if self.mode == ad9959::Mode::SingleBitTwoWire {
```
jordens, line 134: I think the only mode that works for reads is `FourBitSerial`.

```suggestion
        if self.mode != ad9959::Mode::FourBitSerial {
```

src/pounder/mod.rs
jordens: I think it is a incompatibility between the two, not solely attributable to either.
But `QspiInterface` is certainly the right spot to address it. Excellent.

src/pounder/mod.rs
```diff
@@ -32,34 +32,109 @@ const ATT_LE3_PIN: u8 = 8 + 3;
 
 pub struct QspiInterface {
     pub qspi: hal::qspi::Qspi,
+    mode: ad9959::Mode,
+}
+
+impl QspiInterface {
+    pub fn new(mut qspi: hal::qspi::Qspi) -> Result<Self, Error> {
+        qspi.configure_mode(hal::qspi::QspiMode::FourBit).map_err(|_| Error::Qspi)?;
+        Ok(Self { qspi: qspi, mode: ad9959::Mode::SingleBitTwoWire })
```
jordens, line 41: This apparent contradiction and the lack of `qspi.configure_mode()` anywhere else should probably get a comment.

ryan-summers:
@jordens I've removed all pounder-specific code from this pull request and updated this PR to be an update to support Stabilizer v1.1 hardware. I'm now bringing it out of draft as things should be ready for more formal testing and review before merge.

src/main.rs
ryan-summers: Debug pins have now been removed.

src/main.rs
ryan-summers: This has been moved to the start of the transaction (although I do not believe the errata is cause for concern).

src/main.rs
ryan-summers: This has been corrected by updating the two channels to occur at a 50% duty cycle offset.

src/main.rs
ryan-summers: DMA has now been implemented for triggering channel updates.

Cargo.toml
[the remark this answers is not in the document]
ryan-summers: Current text size of this PR is ~200KB when using `cargo size --bin stabilizer`

dnadlinger:
> Building firmware that fits all these different hardware setups is not the goal here […]

Agreed, it is not your job to build something that caters to different use cases. I'd argue that a bit of time spent on making sure other configurations can be added seamlessly in-tree now (when there are large-scale refactorings anyway) is certainly in your interest, though. The configurations you are maintaining will benefit from fixes and additions to the common base (network stack, command interface, …) just as much as other projects, and source-compatibility greatly increases the chance you'll get a PR.

(I haven't looked at the latest changes in that respect yet, beyond noticing that Pounder is gone completely – what's the plan for that?)

> And when I did full rate streaming for both ADCs and both DACs simultaneously at 500 kHz (~50 Mb/s) and both loops running at the same time, that wasn't anywhere near the CPU limit.

Ah, that's very interesting to hear. Students here have been having some trouble with continuous transfer of a single 500 ks/s stream for more than a few seconds (uncongested GbE link, buffered into RAM on the PC side), but this was on modified firmware, and I haven't gotten around to profiling this yet.

ryan-summers:
@dnadlinger 

> Agreed, it is not your job to build something that caters to different use cases. I'd argue that a bit of time spent on making sure other configurations can be added seamlessly in-tree now (when there are large-scale refactorings anyway) is certainly in your interest, though. The configurations you are maintaining will benefit from fixes and additions to the common base (network stack, command interface, …) just as much as other projects, and source-compatibility greatly increases the chance you'll get a PR.
> 
> (I haven't looked at the latest changes in that respect yet, beyond noticing that Pounder is gone completely – what's the plan for that?)

The current implementation plan should allow different configurations to be easily supported with minor updates to the code. This PR has been updated to only support Stabilizer V1.1 hardware. I'm going to be doing two follow-on PRs to revamp the ethernet interface and add pounder support.

> 
> > And when I did full rate streaming for both ADCs and both DACs simultaneously at 500 kHz (~50 Mb/s) and both loops running at the same time, that wasn't anywhere near the CPU limit.
> 
> Ah, that's very interesting to hear. Students here have been having some trouble with continuous transfer of a single 500 ks/s stream for more than a few seconds (uncongested GbE link, buffered into RAM on the PC side), but this was on modified firmware, and I haven't gotten around to profiling this yet.

I'm currently in the testing state of the revamped ethernet interface, so I'll look into throughput/connection issues.

dnadlinger:
> I'm currently in the testing state of the revamped ethernet interface, so I'll look into throughput/connection issues.

Thanks – this would be relevant mostly to streaming, though, which didn't exist in master last time I looked. As long as DMA is still used for ADC reads and the MAC ring buffer handling didn't materially change, it should perform roughly the same also after the refactorings. I'll need to look at (and clean up) our hacked-up code base to see what went wrong there, if it worked in steady-state for @jordens (did you use TCP?).

Issue #108, opened 2020-06-16
make pounder optional


jordens:
Currently this firmware will not boot without Pounder.
Pounder should be made optional. Either and preferrably at runtime (if AD9959 or I2C init fail, remove the pounder endpoints from the API) or by a `#[cfg]` option.

jordens:
#115

PR #115, opened 2020-06-21
Feature/optional pounder

ryan-summers:
This PR updates the initialization function to measure the PGOOD output on pounder to detect whether or not pounder is installed.

This also re-enables the high-speed external oscillator for Stabilizer.

**Note**: No timing was conducted on firmware and power-supply management for monitoring PGOOD. It is currently assumed that PGOOD is steady-state before the firmware measures it.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -172,13 +172,23 @@ const APP: () = {
 
         timer: hal::timer::Timer<hal::stm32::TIM2>,
 
-        // Note: Do not rustfmt the following definition, as it appears to cause a bug in GDB that
-        // results in failing to set breakpoints at correct locations.
+        // Note: It appears that rustfmt generates a format that GDB cannot recognize, which
+        // results in GDB breakpoints being set improperly. To debug, redefine the following
+        // definition to:
+        //
+        // ```rust
+        // net_interface: net::iface::EthernetInterface<
+        //     'static,
+        //     'static,
+        //     'static,
+        //     ethernet::EthernetDMA<'static>>,
+        // ```
         net_interface: net::iface::EthernetInterface<
             'static,
             'static,
             'static,
-            ethernet::EthernetDMA<'static>>,
+            ethernet::EthernetDMA<'static>,
+        >,
         eth_mac: ethernet::EthernetMAC,
         mac_addr: net::wire::EthernetAddress,
 
--- a/src/main.rs
+++ b/src/main.rs
@@ -33,7 +33,7 @@ use rtic::cyccnt::{Instant, U32Ext};
 use stm32h7xx_hal as hal;
 use stm32h7xx_hal::prelude::*;
 
-use embedded_hal::digital::v2::OutputPin;
+use embedded_hal::digital::v2::{InputPin, OutputPin};
 
 use smoltcp as net;
 use stm32h7_ethernet as ethernet;
@@ -175,8 +175,7 @@ const APP: () = {
             'static,
             'static,
             'static,
-            ethernet::EthernetDMA<'static>,
-        >,
+            ethernet::EthernetDMA<'static>>,
         eth_mac: ethernet::EthernetMAC,
         mac_addr: net::wire::EthernetAddress,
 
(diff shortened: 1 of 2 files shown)


jordens:
LGTM.
Prioritize formatting over working around that gdb bug if bors fails.

bors r+

jordens:
This closes #108

ryan-summers:
@jordens I've expanded the comment to include the definition required for debug and run rust-fmt on `main.rs`

dnadlinger:
Hmm… This doesn't seem the way to go for multiple possible add-on boards. (There are already two out in the wild.) Wouldn't a build-time feature be easier?

jordens:
I don't think that would be easier. It would certainly be more inconvenient than runtime discovery. And regarding your board i don't think a compile time option would be practical. Happy to be proven wrong though.

dnadlinger:
> It would certainly be more inconvenient than runtime discovery. 

The current "runtime discovery" strategy depends on one bit of accidentally present information being enough to distinguish between two hardware variants (while there are three now, and definitely more in the future). Elsewhere, you'd object to this being a valid solution on engineering grounds (and I'd agree with you).

> And regarding your board i don't think a compile time option would be practical.

How else do you envision Stabilizer to be maintained going forward? One divergent fork per board, with patches being exchanged either way once in a while if somebody can be bothered? Seems short-sighted.

dtcallcock:
Could unused GPA6, GPA7 and GPB4 of I2C IO expander be used as hardware revision bits?

jordens:
> The current "runtime discovery" strategy depends on one bit of accidentally present information being enough to distinguish between two hardware variants (while there are three now, and definitely more in the future).

What do you mean? It only distinguishes between Pounder present and not. And that firmware only ever targets v1.1. Please explain.

The PGOOD signal is not "accidentally present". It's there by design. It might not be designed to detect Pounder presence, but I agree with Ryan that will work. I can't see a case where it will not.

> How else do you envision Stabilizer to be maintained going forward? One divergent fork per board, with patches being exchanged either way once in a while if somebody can be bothered? Seems short-sighted.

Definitely. v1.0 and v1.1 are fully diverged already on a hardware level and I don't think it's practicable to maintain them in parallel.
The short-sightedness started with the v1.0 hardware design being a prototype that was bound to be incompatible to future revisions. And it continued with people designing mezzanine boards without any discussion about how to support multiple of them. And finally it was crowned by people forking the entire code tree to unidirectionally support their mezzanines without a workable effort to merge back and without an effort to discuss a strategy to maintain multiple use configurations. AFAICT that short-sightedness lies with your Oxford group.

To enable a manageable codebase, we actually did the footwork to refactor the code, rewrite most of it, and move from the initial basic codebase to maintainable structures like the HAL. That effort did not come from any user group and it exemplifies the problem we are facing where everybody wants features but nobody wants to think about code management and sustained maintenance.

dnadlinger:
> I can't see a case where it will not.

Another mezzanine which happens to pull that pin high.

Regarding run-time detection of mezzanines over compile-time versioning, I'm also not sure how feasible building a single executable, for, say, ten different mezzanine boards would be, given the code size constraints.

> Definitely. v1.0 and v1.1 are fully diverged already on a hardware level and I don't think it's practicable to maintain them in parallel.

As you say, v1.0 was a prototype and is going the way of the dodo, so no need to care about maintaining that.

> without a workable effort to merge back 

You got several PRs, representing many hours of development effort from Chris and me, as they addressed "hard" issues (data corruption, …) that had made the initial firmware inadequate for real-world use.

Yes, merging back the other changes is indeed still to be done. It got a bit delayed because
 * much of the work done was people's first foray into Rust, or even embedded programming in general, so the code needs a thorough once-over before going in,
 * the MSc student who did the B field stabilisation work is off for exams (before coming back later this year), and
 * Chris, Tom, and I got sidetracked (main-tracked?) by various deadlines and new exciting roles, plus
 * merging things back would now be quite a bit more involved because of all the refactorings.

Nevertheless, it *is* on the to-do list, because upstreaming the additions is clearly the right thing to do (cf. the DHCP PR, which is still open).

> […] AFAICT that short-sightedness lies with your Oxford group.

I'm not interesting in playing blame games here.

> nobody wants to think about code management and sustained maintenance

Not true; I have been thinking about, and pointing out, the particular issue discussed here since the first signs of an impending refactoring appeared. :)

[unsolicited soapbox rant]

Now, I certainly agree with your wider point, i.e. that new features being easier to sell than less visible work is a challenge: To me, it seems obvious that the real value of this whole project – that is, over everybody building their own one-off boards/software as has been the norm in our community – is precisely to be able to collaborate on high-quality, well-maintained code; aspects which would usually fall by the wayside in typical physicist projects. But that challenge is hardly specific to our case here. In fact, most open-source projects face similar issues, whether in terms of the nature of drive-by contributions, or interaction with in-house development of commercial users. There is a fairly obvious strategy to this: Actively welcome and encourage contributions from new contributors, and then make sure to steer them in a productive direction, whether in issue discussions/code review or through technological nudges, like a project-provided CI platform. Sure, this won't always be very fun to do, but you won't ever get good contributions if you don't get _any_ contributions.

It somewhat pains me to see so little effort spent on this here (ARTIQ proper has a similar problem), as this tends to greatly help project success – and, by extension, future contract work/consulting gigs. You don't need to do all the work yourself, just set up the framework for it to be possible (and encouraged). But short of a fork/hostile takeover, as the steward of the Stabilizer firmware effort only you are in the position to do that. Sure, there is a significant part of our situation in addition to usual OSS concerns that is only indirectly in our purview – what with the conception of software development/engineering in a field emerging from experimental physics in general –, but if we neglect the OSS part, the issue certainly isn't going to fix itself.

[/unsolicited soapbox rant]

Would a PR integrating further mezzanines via compile-time versioning be of interest (code quality issues/… notwithstanding)?

PR #121, opened 2020-06-23
spi nss: very high speed

jordens:
* checked the datasheet for required and recommended OSPEEDR values.
* [ethernet](https://www.st.com/resource/en/datasheet/stm32h743vi.pdf#page=206) interestingly specifies `OSPEEDR=0b10` only. Maybe worth trying.


ryan-summers:
In regard to ethernet, my guess would be that we save a small amount of power during normal operation - the datasheets didn't go into depth about the output drivers. It's worth a try, but I doubt it will yield much.

PR #138, opened 2020-09-11
Cargo.toml: fix stm32h7xx-hal override

sbourdeauducq:
Doing it via dependencies.stm32h7xx-hal causes cargo vendor to fail with:

error: failed to sync

Caused by:
  found duplicate version of package `stm32h7xx-hal v0.5.0` vendored from two sources:

        source 1: registry `https://github.com/rust-lang/crates.io-index`
        source 2: https://github.com/quartiq/stm32h7xx-hal.git?branch=feature/pounder-support#ff00e938

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -58,9 +58,10 @@ branch = "master"
 features = ["stm32h743v"]
 
 [dependencies.stm32h7xx-hal]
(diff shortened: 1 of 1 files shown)


jordens (approved):
Bors r+

PR #142, opened 2020-09-15
Make stabilizer.py work on v0.4.0+

HarryMakes:
This is a "dirty" fix to have `stabilizer.py` working with the new API since v0.4.0, by applying the new JSON format for the IIRs as follows:

```json
{
    "req": "Write",
    "attribute": "stabilizer/iir?/state",
    "value": {
        "channel":1,
        "iir":{
            "ba":[b0,b1,b2,a1,a2],
            "y_offset":?,
            "y_min":?,
            "y_max":?
         }
    }
}
```

The actual ASCII data sent will convert the innermost `"` to `'`, then wrap the curly-bracketed tree of `value` with `"`.

### DIFF
--- a/stabilizer.py
+++ b/stabilizer.py
@@ -17,11 +17,16 @@ class StabilizerConfig:
         self.reader, self.writer = await asyncio.open_connection(host, port)
 
     async def set(self, channel, iir):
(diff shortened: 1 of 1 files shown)


jordens:
Should all disappear with MQTT. But let's be pragmatic, this works for now.
bors r+

Issue #141, opened 2020-09-14
Eth PHY reset time might not be long enough


HarryMakes:
When powering the board with 12V, the Ethernet PHY LAN8742A won't have enough time to reset sometimes. If the reset is done and long enough, the two LEDs on the RJ45 plug should blink briefly upon power-cycling the board - this can't be observed when the PHY hasn't been reset properly.

A solution is to lengthen the deassertion of the nRST signal to **25ms** by replacing [this line](https://github.com/quartiq/stabilizer/blob/master/src/main.rs#L551) in src/main.rs. According to the [PHY's datasheet](http://ww1.microchip.com/downloads/en/DeviceDoc/DS_LAN8742_00001989A.pdf), this should be sufficient for the duration of "external power supplies at operational level to nRST deassertion".

jordens:
t_RSTIA is 100 µs.

![image](https://user-images.githubusercontent.com/1338946/95864640-f482c180-0d65-11eb-971c-944563298494.png)

But arguably there should be sufficient delay between power supply ramp up and that reset. The setup before that may not take long enough.

jordens:
Changing it in software doesn't appear to alter behavior. Changing it in hardware is sinara-hw/Stabilizer#84

PR #143, opened 2020-09-16
Revert rtic from 0.5.5 to 0.5.3

HarryMakes:
This is to fix a panicking upon initialization problem that can be resolved by reverting the update earlier taken on the rtic crate.

Since d5916c550177ba6c3483f31a95b74b01f4a7ee9a where rtic is bumped from 0.5.3 to 0.5.5, the compiled firmware will cause the Stabilizer panick upon powering up. With GDB, the backtrace looks like this:
```
(gdb) backtrace
#0  0x08020c90 in rust_begin_unwind (_info=<optimized out>)
#1  0x08017532 in core::panicking::panic_fmt () at library/core/src/panicking.rs:85
#2  0x0801f0d0 in core::option::expect_failed () at library/core/src/option.rs:1213
#3  0x08014674 in stabilizer::init (c=...)
    at /rustc/6af1bdda54abc9e919fc1137411dfc4311e05649/library/core/src/option.rs:370
#4  0x0801741c in stabilizer::APP::main () at src/main.rs:160
```

Doing `next` on GDB from the first PC will put the CPU to the panicking loop after the following code:
```
...
halted: PC: 0x08000334
537	                *SCB_CPACR | SCB_CPACR_FPU_ENABLE | SCB_CPACR_FPU_USER,
(gdb) n
Info : halted: PC: 0x08000336
halted: PC: 0x08000336
Info : halted: PC: 0x0800033a
halted: PC: 0x0800033a
535	            core::ptr::write_volatile(
(gdb) n
Info : halted: PC: 0x0800033c
halted: PC: 0x0800033c
550	            trampoline()
(gdb) n
Info : halted: PC: 0x08020c88
halted: PC: 0x08020c88
^C
Program received signal SIGINT, Interrupt.
0x08020c96 in rust_begin_unwind (_info=<optimized out>)
```

On a side note, a similar symptom where the Ethernet PHY does not get reset, as discussed in #141, might have also arisen from this new version of rtic. However, the reset delay still seems a bit too short to me.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -145,9 +145,9 @@ dependencies = [
 
 [[package]]
 name = "cortex-m-rtic"
-version = "0.5.5"
+version = "0.5.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b30efcb6b7920d9016182c485687f0012487032a14c415d2fce6e9862ef8260e"
+checksum = "04cd388b154c7e7d212c5af7541ee1f174f29ccb0c22e9117f8d13a5aad233b6"
 dependencies = [
  "cortex-m",
  "cortex-m-rt",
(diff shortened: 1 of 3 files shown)


HarryMakes:
I have just opened an issue over cortex-m-rtic as well: https://github.com/rtic-rs/cortex-m-rtic/issues/361

HarryMakes:
Closed since a new solution is needed after migrating from 0.5.3 to 0.5.5: #144

Issue #140, opened 2020-09-14
`server::json_reply()` raises error after receiving packets on the port


HarryMakes:
When I send a packet to the port (1235 by default), the firmware will panic at the [`socket.write_str()`](https://github.com/quartiq/stabilizer/blob/master/src/server.rs#L149) line of the `json_reply()` function in src/server.rs. The most observable symptom is that Pinging will stop to work after I send the packet. Using GDB, whenever I just send a packet to the TCP socket, the firmware seems to panic. Here is the backtrace:

```
^C
Program received signal SIGINT, Interrupt.
0x08024b28 in rust_begin_unwind (_info=<optimized out>)
(gdb) backtrace
#0  0x08024b28 in rust_begin_unwind (_info=<optimized out>)
#1  0x0801b3ca in core::panicking::panic_fmt () at library/core/src/panicking.rs:85
#2  0x08022f7c in core::option::expect_none_failed () at library/core/src/option.rs:1221
#3  0x0800ec38 in stabilizer::server::json_reply (socket=0x2001ac00, msg=<optimized out>)
    at /rustc/5099914a16a215794ad243df0cc7a05d91d168e0/library/core/src/result.rs:973
#4  0x08011aa4 in stabilizer::idle (c=...) at src/server.rs:127
#5  0x0801b32c in stabilizer::APP::main () at src/main.rs:160
```

I'd like to stress that this issue was not happening for the first few times I started using my Stabilizer v1.1 board. So far, I can't tell how to replicate this same issue on another Stabilizer.

HarryMakes:
@jordens May I know if any person has found this issue and working on it? I've just talked to one of our Stabilizer v1.1 customers, and they also face the same symptom. My temporary patch is disabling `json_reply()` and having `stabilizer.py` to immediately disconnect after issuing the Request packet from the client, which works for both the customer and myself. So I still believe `json_reply()` is causing the issue.

jordens:
I don't know if anybody has seen it. This is also very experimental software with very few active users with no funding to fix these issues. Could you investigate and maybe propose a fix?

jordens:
Also there is some work on stabilizer (mqtt interface) in the planning. So ultimately we'll look at it as well.

ryan-summers:
[...]

PR #161, opened 2020-10-29
Fix server crashing & stabilizer.py hanging upon JSON reply

HarryMakes:
This is to fix the symptoms as described in #140, and shall act as an immediate remedy for current customers before MQTT gets implemented.

With my own experimentation, the client will not get back any replies from the server until it closes its transmission half, which is now done with the additional `write_eof()` call in `stabilizer.py`. 

In addition, if multiple valid JSON requests (or mixed with bytes in an unrecognized format), the server will respond with multiple replies accordingly. This can be tested with the [`socat`](https://linux.die.net/man/1/socat) tool and a plain text file containing lines of the JSON strings:

```shell
cat your_jsons.txt | socat stdio tcp4-connect:<stabilizer-ip>:1235
```

### DIFF
--- a/src/server.rs
+++ b/src/server.rs
@@ -1,7 +1,5 @@
 use heapless::{consts::*, String, Vec};
 
-use core::fmt::Write;
-
 use serde::{Deserialize, Serialize};
 
 use serde_json_core::{de::from_slice, ser::to_string};
@@ -144,9 +142,10 @@ pub struct Status {
 }
 
 pub fn json_reply<T: Serialize>(socket: &mut net::socket::TcpSocket, msg: &T) {
-    let mut u: String<U512> = to_string(msg).unwrap();
-    u.push('\n').unwrap();
-    socket.write_str(&u).unwrap();
+    if socket.can_send() {
+        let u: String<U256> = to_string(msg).unwrap();
+        socket.send_slice(&*u.as_bytes()).unwrap();
+    }
 }
 
 pub struct Server {
@@ -172,7 +171,7 @@ impl Server {
     where
         F: FnMut(&Request) -> Response,
     {
-        while socket.can_recv() {
+        if socket.may_recv() {
             let found = socket
                 .recv(|buf| {
                     let (len, found) =
@@ -222,6 +221,8 @@ impl Server {
                 }
                 self.data.clear();
             }
+        } else if socket.may_send() {
+            socket.close();
         }
     }
 }
--- a/stabilizer.py
+++ b/stabilizer.py
@@ -27,6 +27,7 @@ class StabilizerConfig:
         assert "\n" not in s
         logger.debug("send %s", s)
         self.writer.write(s.encode("ascii") + b"\n")
+        self.writer.write_eof()
         r = (await self.reader.readline()).decode()
         logger.debug("recv %s", r)
         ret = json.loads(r, object_pairs_hook=OD)
--- a/src/main.rs
+++ b/src/main.rs
@@ -757,9 +757,7 @@ const APP: () = {
             {
                 let socket =
                     &mut *sockets.get::<net::socket::TcpSocket>(tcp_handle);
-                if socket.state() == net::socket::TcpState::CloseWait {
-                    socket.close();
-                } else if !(socket.is_open() || socket.is_listening()) {
+                if !socket.is_active() {
                     socket
                         .listen(1235)
                         .unwrap_or_else(|e| warn!("TCP listen error: {:?}", e));


ryan-summers:
@HarryMakes The stabilizer I currently have doesn't have a working ethernet interface, but I have a new one being delivered in the near future. When I get the new hardware, I'll give these changes a test. Thanks for opening the PR!

stabilizer.py
```diff
@@ -27,6 +27,7 @@ async def set(self, channel, iir):
         assert "\n" not in s
         logger.debug("send %s", s)
         self.writer.write(s.encode("ascii") + b"\n")
+        self.writer.write_eof()
```
ryan-summers, line 30: Why write EOF here? We use the newline as "end-of-message" indications - this shouldn't be necessary.

src/server.rs
```diff
@@ -222,6 +221,8 @@ impl Server {
                 }
                 self.data.clear();
             }
+        } else if socket.may_send() {
```
ryan-summers, line 224: I think this would make more sense from a reader's perspective if we wrote it as a separate conditional. Pull it down a line and change it to 
```rust
if socket.may_send() && !socket.may_recv() {
    socket.close();
}
```

That makes it more explicit why we're closing the socket.

src/server.rs
```diff
@@ -144,9 +142,10 @@ pub struct Status {
 }
 
 pub fn json_reply<T: Serialize>(socket: &mut net::socket::TcpSocket, msg: &T) {
-    let mut u: String<U512> = to_string(msg).unwrap();
-    u.push('\n').unwrap();
-    socket.write_str(&u).unwrap();
+    if socket.can_send() {
```
ryan-summers, line 145: This may result in us silently dropping responses, which I don't think is what we want. What's the reason that the socket is not able to send a response in the first place? Is the TX buffer full?

If so, we need to implement a more robust means of flow control here. Silently discarding the reply doesn't seem like the right approach to me.

ryan-summers (changes requested):
In general, I'm hesitant to approve these changes because it doesn't seem like the root cause of the failure has been found.

The changes seem to just be a work-around to make a single stabilizer board start working, but it's not immediately clear why the code wasn't working in the first place or why these changes help.

@HarryMakes Would you mind providing some more context as to what was going wrong and how these changes fix it?

Are you sending more than one JSON request at once? That might be causing the internal buffer handler to overflow. Try doing the requests 1 at a time and see if it still occurs.

src/server.rs
HarryMakes: Thanks for the advice! May I know how `can_send()` (or `may_send()`) is somehow violating the current Request/Response system? Because `may_send()` checks whether or not TX has been closed on the server side, and `can_send()` checks, on top of `may_send()`, whether or not the TX buffer is full, I'm confused why replies would still need to be written to the buffer if the client wouldn't be able to receive anything. I added the `can_send()` check after looking into our other smoltcp projects like [Thermostat](https://git.m-labs.hk/M-Labs/thermostat/src/branch/master/src/main.rs#L205), so I'd like to know why `can_send()` isn't needed in the first place.

I'm also confused why `String<U512>` was used instead of `String<U256>`.

On the other hand, if I keep `json_reply()`'s code unchanged, the Request/Response mechanism still works fine.

stabilizer.py
ryan-summers: Stabilizer should not require the TX channel to be closed to be able to send a reply. That's the root of the problem here.

src/server.rs
ryan-summers: > I'm also confused why String<U512> was used instead of String<U256>.

The template time (`U256` or `U512`) indicates the size of the buffer backing the string (e.g. 256 or 512 bytes). It's an arbitrary selection.

> May I know how can_send() (or may_send()) is somehow violating the current Request/Response system?

Placing a check in `json_reply()` is not the right approach because we are actively trying to respond to a request that we received. `can_send()` may return false simply because the TX buffer is full. With the code as written, we silently ignore this failure, but the client never receives a reply to the request that they sent and may hang forever.

I get that there's no real way around it, but we should at least log a warning here if we can't send the message. Additionally, this is actually a redundant check. We can tell if the socket was able to send the reply by checking the status of `socket.write_str(&u)` (instead of unwrapping it) and use the resulting `Result` to detect a failure to transmit the response.

HarryMakes:
As per 659a6879f7b78495bb4b3dbd3f7047298ea7c1d0, the server no longer crashes while the connection is alive and can send JSON replies properly. Closing this PR.

PR #162, opened 2020-10-30
added gateway

### LINKED ISSUE #158: Not pingable through gateway
Stabilizer and a computer are connected to different networks. Pinging doesn't work although the computer has a connection to the other network through gateway. 

Pinging from a computer that is in stabilizers network works.

nkuh:
Fixes #158

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -36,6 +36,8 @@ use stm32h7xx_hal::prelude::*;
 use embedded_hal::digital::v2::{InputPin, OutputPin};
 
 use smoltcp as net;
+use smoltcp::wire::Ipv4Address;
+use smoltcp::iface::Routes;
 use stm32h7_ethernet as ethernet;
 
 use heapless::{consts::*, String};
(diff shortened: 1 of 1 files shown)


src/main.rs
```diff
@@ -85,6 +88,8 @@ static mut NET_STORE: NetStorage = NetStorage {
     )],
 
     neighbor_cache: [None; 8],
+
+    routes_storage: [None; 2],
```
ryan-summers, line 92: Why are we adding two routes? I only see a single route added to the list below

src/main.rs
```diff
@@ -617,13 +622,18 @@ const APP: () = {
                 24,
             );
 
+            let default_v4_gw = Ipv4Address::new(10,0,16,1);
```
ryan-summers, line 625: spaces between the IP:
`IPv4Address::new(10, 0, 16, 1)`

ryan-summers (changes requested):

src/main.rs
ryan-summers: Please run `cargo fmt --all` as well

ryan-summers (approved):

PR #165, opened 2020-11-05
Stabilizer asynchronous batch sampling support

ryan-summers:
This PR updates the stabilizer firmware to utilize batch-based sampling of the ADCs. This essentially offloads the task of triggering SPI reads and gathering ADC samples into a buffer into hardware DMA. This PR also updates DAC outputs to utilize a DMA trigger to periodically write DAC codes to the DAC SPI as well.

A single timer drivers DAC + ADC DMA transfers - four comparison channels (each configured to compare against timer value 0) are used to generate 4 DMA requests, which initiate DAC writes and ADC samples simultaneously.

To generally describe the new ADC implementation, there are four DMA streams and one timer used to offload sampling. The ADC SPI interfaces are configured in infinite transaction mode with auto-suspend CS when no data is available. To trigger an ADC sample, a constant word is written to the SPI TXFIFO by the trigger DMA, which is repeatedly requested by a timer channel. This effectively controls the sample rate on both channels.

Two other DMA streams are configured to continuously read from the SPI RX FIFOs to gather data. The DMA will only read when data is present in the RX FIFO, so these will wait until the desired number of samples are acquired.

Once the DMA completes gathering N samples, an interrupt is generated for the DMA stream and firmware retrieves the buffers provided to DMA, which are now filled with ADC samples. It then schedules the next DMA transaction for additional samples before continuing to perform DSP on the acquired ADC samples.

Once DSP has been completed on the samples, they are enqueued into a FIFO for transmission on the DAC. The DAC DMA transfers are configured and started. The DACs are then asynchronously updated.

**TODO**
- [x] ~~Wait until https://github.com/stm32-rs/stm32h7xx-hal/pull/153 is merged into the mainline HAL~~
    - We will instead track the `dma` branch, as it may take some time before merge is completed
- [x] Document new modules
- [x] Merge https://github.com/quartiq/stabilizer/pull/159
- [x] Update ADC0 and ADC1 trigger DMA channels to use distinct DMA requests


Cargo.toml
```diff
@@ -54,7 +54,8 @@ path = "ad9959"
 
 [dependencies.stm32h7xx-hal]
 features = ["stm32h743v", "rt", "unproven", "ethernet", "quadspi"]
-git = "https://github.com/stm32-rs/stm32h7xx-hal"
+git = "https://github.com/quartiq/stm32h7xx-hal"
+branch = "feature/stabilizer-dma"
```
ryan-summers, line 58: We should document why this is using a branch. My current reasons are:
* incorporates `dma` branch from official repo
* Removes the `Drop` trait from DMA streams

We can likely use the `dma` branch from the source repository directly as well, which may be easier.

Cargo.toml
```diff
@@ -64,7 +65,7 @@ nightly = ["cortex-m/inline-asm"]
 [profile.dev]
 codegen-units = 1
 incremental = false
-opt-level = 3
+opt-level = 1
```
ryan-summers, line 68: Should revert the opt-level back to 3

openocd.gdb
```diff
@@ -26,3 +26,6 @@ set var $t0=*$cc
 continue
 end
 #set var $t0=*$cc
+
+source ../../PyCortexMDebug/cmdebug/svd_gdb.py
+svd_load ~/Downloads/STM32H743x.svd
```
ryan-summers, line 31: Remove local debug-code

ryan-summers:
@irwineffect After our discussion yesterday, Robert and I discovered that DMA transfers behave differently than documentation would lead you to believe.

When a peripheral generates a DMA request, this results in a single data transfer of the request, not a completion of the entire request.

This means that it's possible to use the DMA channels to also update the DAC output codes regularly. I have since updated the PR to support DMA-driven DAC updates as well.

jordens:
Would macros or generics for the timer channels, the ADC code, and the DAC code shrink the code?

jordens (approved):
Looks good. The refactoring with macros/generics to reduce code footprint can also be done later.

src/adc.rs
```diff
@@ -0,0 +1,359 @@
...
+
+    /// Handle a transfer completion.
+    ///
+    /// # Returns
+    /// A reference to the underlying buffer that has been filled with ADC samples.
+    pub fn transfer_complete_handler(&mut self) -> &[u16; SAMPLE_BUFFER_SIZE] {
+        let next_buffer = self.next_buffer.take().unwrap();
+
+        // Wait for the transfer to fully complete before continuing.
+        while self.transfer.get_transfer_complete_flag() == false {}
```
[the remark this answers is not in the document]
irwineffect: I suppose we'll get the SPI Rx FIFO overrun interrupt if we did get stuck here, so we don't really care.

src/dac.rs
```diff
@@ -0,0 +1,279 @@
...
+        Self {}
+    }
+}
+
+unsafe impl TargetAddress<MemoryToPeripheral> for SPI4 {
+    /// SPI2 is configured to operate using 16-bit transfer words.
+    type MemSize = u16;
+
+    /// SPI4 DMA requests are generated whenever TIM2 CH3 comparison occurs.
+    const REQUEST_LINE: Option<u8> = Some(DMAReq::TIM2_CH3 as u8);
```
irwineffect, line 40: Why not use a different timer so that we could have different rates for the DACs vs the ADCs? I suppose we could  adjust the output compare values to let the DAC run at some fraction of the ADCs if needed.

src/dac.rs
```diff
@@ -0,0 +1,279 @@
...
+
+        // Copy data into the next buffer
+        next_buffer.copy_from_slice(data);
+
+        // If the last transfer was not complete, we didn't write all our previous DAC codes.
+        // Wait for all the DAC codes to get written as well.
+        if self.first_transfer {
+            self.first_transfer = false
+        } else {
+            while self.transfer.get_transfer_complete_flag() == false {}
```
irwineffect, line 177: Do we want to stall here? What if we need to need to service the other DAC?

src/main.rs
```diff
@@ -256,207 +264,183 @@ const APP: () = {
...
+                    .set_speed(hal::gpio::Speed::VeryHigh);
 
-            spi.listen(hal::spi::Event::Rxp);
+                let config = hal::spi::Config::new(hal::spi::Mode {
+                    polarity: hal::spi::Polarity::IdleHigh,
+                    phase: hal::spi::Phase::CaptureOnSecondTransition,
+                })
+                .manage_cs()
+                .suspend_when_inactive()
+                .cs_delay(220e-9);
```
irwineffect, line 302: Should we create a single definition for this delay value?

src/main.rs
```diff
@@ -256,207 +264,183 @@ const APP: () = {
...
 
-            spi
-        };
+                let spi: hal::spi::Spi<_, _, u16> = dp.SPI2.spi(
+                    (spi_sck, spi_miso, hal::spi::NoMosi),
+                    config,
+                    50.mhz(),
```
irwineffect, line 307: Single definition for this too.

src/main.rs
```diff
@@ -771,30 +743,33 @@ const APP: () = {
...
-        };
+    #[task(binds=DMA1_STR3, resources=[adcs, dacs, iir_state, iir_ch], priority=2)]
+    fn adc_update(c: adc_update::Context) {
+        let (adc0_samples, adc1_samples) =
+            c.resources.adcs.transfer_complete_handler();
 
-        c.resources.dac1.send(output).unwrap();
-    }
+        let mut dac0: [u16; SAMPLE_BUFFER_SIZE] = [0; SAMPLE_BUFFER_SIZE];
+        let mut dac1: [u16; SAMPLE_BUFFER_SIZE] = [0; SAMPLE_BUFFER_SIZE];
```
irwineffect, line 752: Should we use [core::mem::MaybeUninit](https://doc.rust-lang.org/core/mem/union.MaybeUninit.html) to save the time initializing these buffers since we write to them just below here?

irwineffect: Actually, could we add a method to the DAC module that returns a reference to the buffer the DMA is not currently  using, so we could write directly into that buffer? This would save us the time of copying from these buffers into the internal DAC buffers in `dacs.next_data()`

src/dac.rs
```diff
@@ -0,0 +1,279 @@
...
+///! results in DAC update for both channels.
+use super::{
+    hal, sampling_timer, DMAReq, DmaConfig, MemoryToPeripheral, TargetAddress,
+    Transfer, SAMPLE_BUFFER_SIZE,
+};
+
+// The following global buffers are used for the DAC code DMA transfers. Two buffers are used for
+// each transfer in a ping-pong buffer configuration (one is being prepared while the other is being
+// processed). Note that the contents of AXI SRAM is uninitialized, so the buffer contents on
+// startup are undefined.
```
irwineffect, line 14: Should we explicitly initialize the DAC buffers to some "safe" value? I'm thinking the DACs will spit out garbage when they are first turned on as the DMA reads from the very first buffer.

src/sampling_timer.rs
```diff
@@ -0,0 +1,147 @@
...
+pub struct Timer2Channel4 {}
+
+impl Timer2Channel4 {
+    /// Allow CH4 to generate DMA requests.
+    pub fn listen_dma(&self) {
+        let regs = unsafe { &*hal::stm32::TIM2::ptr() };
+        regs.dier.modify(|_, w| w.cc4de().set_bit());
+    }
+
+    /// Operate CH2 as an output-compare.
```
irwineffect, line 136: CH4

src/sampling_timer.rs
```diff
@@ -0,0 +1,147 @@
...
+pub struct Timer2Channel3 {}
+
+impl Timer2Channel3 {
+    /// Allow CH4 to generate DMA requests.
+    pub fn listen_dma(&self) {
+        let regs = unsafe { &*hal::stm32::TIM2::ptr() };
+        regs.dier.modify(|_, w| w.cc3de().set_bit());
+    }
+
+    /// Operate CH2 as an output-compare.
```
irwineffect, line 113: CH3

src/adc.rs
```diff
@@ -0,0 +1,359 @@
...
+        // occurs.
+        trigger_channel.listen_dma();
+        trigger_channel.to_output_compare(0);
+
+        // The trigger stream constantly writes to the TX FIFO using a static word (dont-care
+        // contents). Thus, neither the memory or peripheral address ever change. This is run in
+        // circular mode to be completed at every DMA request.
+        let trigger_config = DmaConfig::default()
+            .memory_increment(false)
+            .peripheral_increment(false)
```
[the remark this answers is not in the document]
ryan-summers: While it's default for now, if that changed in the future, it may inadvertently break our code. For the more obscure parameters, the defaults likely won't change. However, I could logically see `memory_increment()` being changed to default-true in the future.

src/adc.rs
```diff
@@ -0,0 +1,359 @@
...
+    type MemSize = u16;
+
+    /// SPI2 DMA requests are generated whenever TIM2 CH1 comparison occurs.
+    const REQUEST_LINE: Option<u8> = Some(DMAReq::TIM2_CH1 as u8);
+
+    /// Whenever the DMA request occurs, it should write into SPI2's TX FIFO to start a DMA
+    /// transfer.
+    fn address(&self) -> u32 {
+        let regs = unsafe { &*hal::stm32::SPI2::ptr() };
+        &regs.txdr as *const _ as u32
```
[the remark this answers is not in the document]
ryan-summers: The SPI registers are owned by the `hal::spi::Spi` structure, which is then owned by the `data_transfer`. Hence, it's not possible to move the singleton into this struct and why we have to use unsafe here.

src/sampling_timer.rs
```diff
@@ -0,0 +1,147 @@
...
+        Self {
+            timer,
+            channels: Some(TimerChannels::new()),
+        }
+    }
+
+    /// Get the timer capture/compare channels.
+    pub fn channels(&mut self) -> TimerChannels {
+        self.channels.take().unwrap()
+    }
```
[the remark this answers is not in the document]
ryan-summers: This holds because the HAL is not currently making use of any of the timer channels. Ideally, this type of support would be added to the HAL in the future, although you make a good point.

If the HAL were updated in the future to start using the underlying timer channels, invariants may break. I don't see a way around this beyond documenting it in the code.

src/sampling_timer.rs
```diff
@@ -0,0 +1,147 @@
...
+            ch3: Timer2Channel3 {},
+            ch4: Timer2Channel4 {},
+        }
+    }
+}
+
+/// Representation of CH1 of TIM2.
+pub struct Timer2Channel1 {}
+
+impl Timer2Channel1 {
```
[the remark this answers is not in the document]
ryan-summers: Macro implemented using `paste`.

src/sampling_timer.rs
```diff
@@ -0,0 +1,147 @@
...
+    /// Allow CH1 to generate DMA requests.
+    pub fn listen_dma(&self) {
+        let regs = unsafe { &*hal::stm32::TIM2::ptr() };
```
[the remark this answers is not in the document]
ryan-summers: Added docs throughout about safety.

src/sampling_timer.rs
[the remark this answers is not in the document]
ryan-summers: Added docs throughout about safety.

src/adc.rs
```diff
@@ -0,0 +1,359 @@
...
+            // Allow the SPI FIFOs to operate using only DMA data channels.
+            spi.enable_dma_rx();
+            spi.enable_dma_tx();
+
+            // Enable SPI and start it in infinite transaction mode.
+            spi.inner().cr1.modify(|_, w| w.spe().set_bit());
+            spi.inner().cr1.modify(|_, w| w.cstart().started());
+        });
+
+        trigger_transfer.start(|_| {});
```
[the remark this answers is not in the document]
ryan-summers: Because we need to actually start the SPI peripheral there. The trigger transfer start just starts allowing the timer overflow to schedule SPI transfers. If the SPI isn't enabled, it won't do anything.

src/dac.rs
ryan-summers: This is likely something we may want in the future, but is omitted in this PR for simplicity.

src/main.rs
ryan-summers: Any reason why you're advocating for moving it out to a singular definition?

For context, the 50MHz is the maximum clock frequency allowed by the DACs.

src/main.rs
ryan-summers: Any reason you're advocating for moving this out? Is it for the sake of documenting the magic number?

src/adc.rs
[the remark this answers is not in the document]
jordens: The (maybe unwritten) rule is to stick with the peripheral reset values as defaults unless the functionality exposed implies deviating from the default. But always better to specify it.

src/dac.rs
ryan-summers: The DAC transfers aren't enabled at first. They are only started after the first `next_data()` call, so there's no uninitialized data ever written to the DACs.

src/adc.rs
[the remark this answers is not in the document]
ryan-summers: With plans to change the overrun ISRs with incrementing counters, we should likely implement a more robust check here.

src/dac.rs
ryan-summers: Stalling here is the simpler solution from a code standpoint. We won't be stalling very long

src/main.rs
ryan-summers: Great recommendation here, as it removes the last unnecessary data operation from the read->dsp->write standpoint. It has now been removed.

src/main.rs
ryan-summers: Moved to a new `design_parameters.rs` file containing other constants. Will likely move things out there too to clean up `main.rs` in the future.

src/main.rs
ryan-summers: Moved to a new `design_parameters.rs` file containing other constants. Will likely move things out there too to clean up `main.rs` in the future.

src/adc.rs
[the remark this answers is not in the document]
ryan-summers: This isn't possible for the `adc` since we're using both RX and TX channels, but it _is_ possible for the DACs and eliminates some unsafe code, so I've updated that module.

src/adc.rs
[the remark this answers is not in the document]
ryan-summers: I think the right approach here would be instead to rely on the watchdog catching us hanging.

* This flag should essentially always be set by the time we get here
* Adding a counter for number of viable checks is essentially just giving us a few clock cycles of lee-way and is not a reliable mechanism to handle bus slowdown or bus-contention related issues

I think viable solutions are as follows:
1. Panic immediately if the flag is not set
2. Loop until the flag _is_ set and rely on a watchdog or external mechanism to catch us

We can defer adding watchdog functionality for the future, so I'm of the opinion that this loop is acceptable for now. Let me know what you think

src/adc.rs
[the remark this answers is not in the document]
ryan-summers: After some discussion - this is safe as is, but it would be nice to expand the DMA API of the HAL so that in the future, full-duplex SPI communication could be supported without needing to use unsafe code.

src/adc.rs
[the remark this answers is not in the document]
irwineffect: Ah, also  `trigger_transfer` doesn't have ownership of the SPI registers, so we can't do it in that closure.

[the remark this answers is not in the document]
ryan-summers: After further discussion, this isn't possible for the DACs because SPI is owned by the transfer. For symmetry, we will follow the API from the DACs and keep it as-is.

src/adc.rs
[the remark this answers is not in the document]
ryan-summers: I've now updated to remove setting the default values.

src/adc.rs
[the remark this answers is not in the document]
ryan-summers: After some discussion, we have added comments and increased priority on the overrun interrupts.

src/dac.rs
ryan-summers: Added comment + priority change to these loops to indicate they are not checked formaximum times for speed purposes.

irwineffect (approved):
Looks good!

Issue #170, opened 2020-11-16
Timestamping


jordens:
### External reference on digital input

* ADC/DAC sampling timer period `t`, integer in units of some timer clock period `τ` (e.g 10 ns) locked to the CPU PLL, and thus to the CPU XO. For example `t = 128 τ ~ 1.28 µs`. `τ` is a design constant, `t` is a compile-time constant. We should aim at making `t` a power of two.
* `n` DAC and ADC samples per batch,  batch period `n*t`, `n` being a power of two compile-time constant, e.g. `n = 4`.
* Timestamping timer overflow/reload period is `j*n*t`. Locked through the CPU PLL to the CPU XO. Not yet/necessarily deterministic (across reboots, resets, timer reconfigurations) phase w.r.t. ADC/DAC sampling timer. `j >= 1`, a power of two, as large as possible within the timer counter size.
* Digital input edges capture the timestamping timer counter: `s` (`u16 as i16`), Digital input frequency typically anywhere from 50 Hz (mains) to 500 kHz (less than ADC Nyquist typically). These are samples of the batch phase at zero reference phase.
* Processing routine receives `m` timestamps for the current batch. Number of timestamps `0 <= m <= n` depends on reference frequency, no new timestamps is explicitly possible.
* Processing routine scales the timestamps from `mod j*n*t` to natural `i32` `mod (1 << 32)`. Pure left shift.
* Feed all received scaled `s` into the PLL en-bloc and at the end of the batch, store most recent phase `y`, frequency `f` (both `i32`) and the time `k` in units of full timer overflow periods between the two last PLL updates (i.e. `k = 0` necessarily if `m >= 2`, though `k` can also be `0` if `m < 2`). `j > 1` will help with that.
* Converted frequency and phase: `g = (1 << 64)/(n*f + n*(k << 32))`, `v = -(y << 32)/f`. (**TODO**: distribute exponents or just `i64` math).
* Harmonic index `u: i32`.
* Each ADC sample is assigned the phase `p_i: i32 = u*(i*g + v)`.
* Demodulate with the negative of that phase.

### 125 MHz from Pounder on ETR

* Timestamping timer with counter driven from Pounder period `σ`, `σ = r*4*2 ns`, typically the timer ETR prescaler is `r = 4` or `r = 8`, compile-time constant. `σ` is a design constant.
[...]

Issue #60, opened 2019-11-20
lock-in


jordens:
* Should work well to >= 100 kHz modulation frequency
* Modulation styles with varying complexity:
  * external modulation and a soft PLL
  * the digital input and timed ADC DMA sampling or time tagging of samples/blocks
  * internal modulation on a separate DAC channel
  * internal modulation on the PI-controlled DAC channel
  * internal modulation on a digital output
* 1f/2f, adjustable phase
* re-use existing IIR filter

Old prototype code:
https://github.com/quartiq/queenmod

![image](https://user-images.githubusercontent.com/1338946/106497939-fa401680-64be-11eb-8591-eff138d4cd2d.png)

jordens:
Implemented, no characterization and battle testing yet.

PR #173, opened 2020-11-19
Lock-in

matthuszagh:
This PR adds lock-in amplifier functionality to stabilizer. There are 3 primary public functions that a processing sequence in stabilizer might use. `prefilt` performs demodulation and decimation. `postfilt_iq` performs the same steps as `prefilt`, but additionally performs biquad IIR filtering before decimation. This generates the in-phase (I) and quadrature (Q) signal components (also frequently called X and Y). `postfilt_at` performs the same steps as `postfilt_iq`, but additionally converts the I and Q components to magnitude and phase values. There is an additional public function, `arr`, which is a const function for computing the ARR register overflow value from the ADC sampling period and number of ADC samples in a batch. There is also a public struct `TimeStamp`. A mutable array of 2 of these is used to keep track of the last 2 external reference clock counter values.

**Notable Changes**
1. I separated the IIR code into its own separate subproject to use it in the lock-in code.

**TODO**
- [ ] Currently, the `postfilt` functions filter before decimating. However, @jordens discussed performing a boxcar average and then filtering. The code is in place (but commented out) to do this. It also yields significant (~30% for 16 ADC batch samples) performance gains in terms of total processing latency since filtering only needs to be performed on 1 sample rather than the full batch.
- [ ] `low_pass.rs` provides integration tests that test the lock-in functionality in a traditional lock-in low-pass filtering case. Part of this test involves computing acceptable tolerances for the amplitude and phase values for various configurations (noise frequencies and amplitudes, filter cutoff frequency, etc.). Although these computations seem to provide reasonable tolerance values, I'm not absolutely convinced that they're technically correct. Any feedback on this is appreciated.
- [ ] I've included annotations giving timing latencies for major contributors in the DSP processing chain (based on a GPIO toggle). I've left these in because I expect they may be helpful in the review process. However, they'll probably be removed before the actual merge.
- [ ] I've added several global constants (e.g., for the ADC batch size). These will be defined as part of [PR #165](https://github.com/quartiq/stabilizer/pull/165) and can be removed from the lockin code at that point.

Finally, it's worth noting that this code would benefit from const generics. This would allow testing for different batch and output sizes and would no longer require using global constants. Unfortunately, we have to wait several months before const generics will be stabilized.

Related:
- [Issue #60 (lock-in)](https://github.com/quartiq/stabilizer/issues/60)
- [Issue #170 (timestamping)](https://github.com/quartiq/stabilizer/issues/170)


ryan-summers (commented):
Only did a structural pass at this point, but it seems like it may make more sense to make a single `dsp`/`digital_signal_processing` crate at the root level that contains `iir` and the `lockin` functionality.

lockin/src/lib.rs
```diff
@@ -0,0 +1,973 @@
...
+use trig::{atan2, cos, sin};
+
+// TODO temporary, pending new DMA code
+pub const SAMPLE_BUFFER_SIZE: usize = 16;
+pub const TSTAMP_BUFFER_SIZE: usize = SAMPLE_BUFFER_SIZE / 2;
+pub const OUTPUT_BUFFER_SIZE: usize = 1;
+
+/// Slow external reference edge timestamps.
+#[derive(Copy, Clone)]
+pub struct TimeStamp {
```
jordens, line 19: This can just be a `t: i32`. The age increment per batch for an old timestamp would be `t -= something`.

lockin/src/lib.rs
```diff
@@ -0,0 +1,973 @@
...
+
+/// Compute the reference signal phase increment for each ADC sample
+/// and use them to demodulate the ADC inputs.
+macro_rules! prefilt_no_decimate {
+    ( $adc_samples:expr, $ref_tstamps:expr, $valid_tstamps:expr, $phi:expr, $tadc:expr, $fscale:expr, $tstamps_mem:expr) => {{
+        record_new_tstamps($ref_tstamps, $valid_tstamps, $tstamps_mem);
+        let ts_valid_count = tstamps_valid_count($tstamps_mem);
+
+        if ts_valid_count < 2 {
+            return ([0.; OUTPUT_BUFFER_SIZE], [0.; OUTPUT_BUFFER_SIZE]);
```
jordens, line 74: This incorrect. Just return either an `Option` or a `Result` if you don't have enough data to do processing yet.

lockin/src/lib.rs
```diff
@@ -0,0 +1,973 @@
...
+    // one that has not yet been set).
+    pub sequences_old: i32,
+}
+
+impl TimeStamp {
+    // Initialize a new TimeStamp.
+    pub fn new() -> Self {
+        TimeStamp {
+            count: 0,
+            sequences_old: -1,
```
jordens, line 36: No magic numbers.
Use `Option` in these cases.

lockin/src/lib.rs
```diff
@@ -0,0 +1,973 @@
...
+    /// called at the end of processing a sequence of ADC samples.
+    pub fn new_sequence(&mut self) {
+        if self.sequences_old != -1 {
+            self.sequences_old += 1;
+        }
+    }
+
+    /// Returns true if TimeStamp is valid.
+    pub fn is_valid(&self) -> bool {
+        self.sequences_old != -1
```
jordens, line 50: `.is_some()`.

lockin/src/lib.rs
```diff
@@ -0,0 +1,973 @@
...
+///
+/// # Arguments
+///
+/// `i` - In-phase signal.
+/// `q` - Quadrature signal.
+fn iq_to_t(i: f32, q: f32) -> f32 {
+    atan2(q, i)
+}
+
+/// Demodulation phase values corresponding to each ADC sample.
```
jordens, line 332: `Compute ...`

lockin/src/lib.rs
```diff
@@ -0,0 +1,973 @@
...
+        self.count = newval as u32;
+        self.sequences_old = 0;
+    }
+}
+
+/// Compute the reference signal phase increment for each ADC sample
+/// and use them to demodulate the ADC inputs.
+macro_rules! prefilt_no_decimate {
+    ( $adc_samples:expr, $ref_tstamps:expr, $valid_tstamps:expr, $phi:expr, $tadc:expr, $fscale:expr, $tstamps_mem:expr) => {{
+        record_new_tstamps($ref_tstamps, $valid_tstamps, $tstamps_mem);
```
jordens, line 70: This shifts the timestamps into a two-deep fifo. Maybe with more concise terminology this can be formulated better.

lockin/src/lib.rs
```diff
@@ -0,0 +1,973 @@
...
+/// instance, they can be from different processing periods and this
+/// will still count the number of counts between them, accounting for
+/// overflow wrapping. Both TimeStamps must be valid for this to be
+/// used.
+///
+/// # Arguments
+///
+/// * `tstamps_mem` - Recorded TimeStamps.
+/// * `overflow_count` - Max timestamp value.
+fn tstamps_diff(tstamps_mem: &[TimeStamp; 2], overflow_count: u32) -> u32 {
```
jordens, line 396: If these were just `i32`, it would be a simple difference, right? Inline then.

lockin/src/lib.rs
```diff
@@ -0,0 +1,973 @@
...
+    if tstamps_mem[0].sequences_old == tstamps_mem[1].sequences_old {
+        return tstamps_mem[0].count - tstamps_mem[1].count;
+    }
+
+    let rem0: u32 = tstamps_mem[0].count;
+    let rem1: u32 = overflow_count - tstamps_mem[1].count;
+    let empty_sequences =
+        tstamps_mem[1].sequences_old - tstamps_mem[0].sequences_old - 1;
+
+    rem0 + rem1 + overflow_count * empty_sequences as u32
```
jordens, line 408: Also, if you write that out on paper and simplify it, this becomes clear and easier to read. It also becomes obvious that here you are implementing a number representation and math on it.

lockin/src/lib.rs
```diff
@@ -0,0 +1,973 @@
...
+}
+
+/// Add new timestamps to the TimeStamp memory.
+///
+/// # Arguments
+///
+/// * `ref_tstamps` - New timestamp values.
+/// * `valid_tstamps` - Number of valid timestamps.
+/// * `tstamps_mem` - Last 2 recorded timestamps.
+fn record_new_tstamps(
```
jordens, line 283: The two-timestamp storage should probably be just `[Option<i32>; 2]`.

lockin/src/lib.rs
```diff
@@ -0,0 +1,973 @@
...
+            &[
+                TimeStamp {
+                    count: 0,
+                    sequences_old: 0,
+                },
+                tstamps_mem[0],
+            ],
+            overflow_count,
+        );
+    }
```
jordens, line 373: Could you elaborate/document this? I have a hard time understanding.

jordens (changes requested):

lockin/src/lib.rs
```diff
@@ -0,0 +1,973 @@
...
+        let thetas = adc_phases(
+            $ref_tstamps[0] as u32,
+            $tstamps_mem,
+            $phi,
+            $fscale,
+            $tadc,
+        );
+        let mut sines: [f32; SAMPLE_BUFFER_SIZE] = [0.; SAMPLE_BUFFER_SIZE];
+        let mut cosines: [f32; SAMPLE_BUFFER_SIZE] = [0.; SAMPLE_BUFFER_SIZE];
+        for i in 0..SAMPLE_BUFFER_SIZE {
```
ryan-summers, line 86:
```rust
for theta in thetas.iter() {
    // ...
}
```

should be sufficient, shouldn't it?

lockin/src/lib.rs
```diff
@@ -0,0 +1,973 @@
...
+/// instance, 2 would demodulate with the first harmonic of the reference
+/// frequency.
+/// * `tstamps_mem` - Last two external reference timestamps (i.e., recorded
+/// values of `ref_tstamps`.)
+///
+/// # Latency (ADC batch size = 16): 5.6us
+///
+/// * 2.0us to compute ADC phase values
+/// * 3.4us to compute sin and cos
+pub fn prefilt(
```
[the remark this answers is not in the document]
ryan-summers: I think this would actually benefit a lot by creating a `DspFilter` object and then `processing()` ADC samples to generate I/Q.

Sample API:
```rust
pub struct DspFilter {
    phase_offset: u32,
    adc_cycles: u32,
    frequency_scale: u32,
    timestamps: [TimeStamp; 2].
}

impl DspFilter {
    pub fn prefilter(adc_samples: &[i16], timestamps: &[u16]) -> Result<([f32; OUTPUT_BUFFER_SIZE], [f32; OUTPUT_BUFFER_SIZE]), Error> {
        // ...
    }

    // Etcetera
}
```

I believe Rust may be able to analyze things better and recognize that certain parameters are actually `const` as well.

lockin/src/lib.rs
```diff
@@ -0,0 +1,973 @@
...
+            || (SAMPLE_BUFFER_SIZE / OUTPUT_BUFFER_SIZE) % 2 == 0
+    );
+
+    let n_k: usize = SAMPLE_BUFFER_SIZE / OUTPUT_BUFFER_SIZE;
+    let mut res_i: [f32; OUTPUT_BUFFER_SIZE] = [0.; OUTPUT_BUFFER_SIZE];
+    let mut res_q: [f32; OUTPUT_BUFFER_SIZE] = [0.; OUTPUT_BUFFER_SIZE];
+
+    let mut n: usize = 0;
+    let mut k: usize = 0;
+    while n < SAMPLE_BUFFER_SIZE {
```
ryan-summers, line 510: I think we could do something like

```rust
for (src, dest) in i.iter().step_by(n_k).zip(res_i.iter_mut()) {
    *dest = src;
}
```

lockin/src/lib.rs
```diff
@@ -0,0 +1,973 @@
...
+/// # Arguments
+///
+/// `i` - In-phase signals.
+/// `q` - Quadrature signals.
+/// `iir` - IIR filters for the in phase (index 0) and quadrature
+/// (index 1) signals.
+/// `iirstate` - State of each IIR filter.
+///
+/// Latency (ADC batch size = 16): 3.5us
+fn filter(
```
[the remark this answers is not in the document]
ryan-summers: I agree - I think we could actually just do this in-place with a `map()`. It would be cleaner and likely more performant. Just mutably borrow the original array and modify it in-place.

lockin/src/lib.rs
```diff
@@ -0,0 +1,973 @@
...
+/// Decimate from `SAMPLE_BUFFER_SIZE` to `OUTPUT_BUFFER_SIZE`
+/// samples. `SAMPLE_BUFFER_SIZE`/`OUTPUT_BUFFER_SIZE` is assumed to
+/// be equal to 2**n, where n is some non-negative integer. Decimates
+/// the in-phase and quadrature signals separately and returns the
+/// result as (i, q).
+///
+/// # Arguments
+///
+/// `i` - In-phase signal.
+/// `q` - Quadrature signal.
```
ryan-summers, line 494: I would prefer if `Return` was under it's own heading just like `Arguments`.

ryan-summers: This is true for all functions, I just have only placed this comment here

lockin/src/lib.rs
```diff
@@ -0,0 +1,973 @@
...
+///
+/// # Arguments
+///
+/// `i` - In-phase signals.
+/// `q` - Quadrature signals.
+/// `iir` - IIR filters for the in phase (index 0) and quadrature
+/// (index 1) signals.
+/// `iirstate` - State of each IIR filter.
+///
+/// Latency (ADC batch size = 16): 3.5us
```
ryan-summers, line 467: When specifying latency like this, there's a _lot_ of test parameters these numbers depend on and they could very easily change (e.g. by adjusting clock speeds, code getting ejected from the cache, etc.). It would be good to add context as to how these values were determined as well.

lockin/src/lib.rs
```diff
@@ -0,0 +1,973 @@
...
+    fscale: u32,
+    tadc: u32,
+) -> [f32; SAMPLE_BUFFER_SIZE] {
+    let overflow_count: u32 = tadc * SAMPLE_BUFFER_SIZE as u32;
+    let tref_count: u32 = tstamps_diff(tstamps_mem, overflow_count);
+    let mut thetas: [f32; SAMPLE_BUFFER_SIZE] = [0.; SAMPLE_BUFFER_SIZE];
+    let mut theta_count: u32;
+
+    // 68ns
+    if tstamps_mem[0].sequences_old == 0 {
```
ryan-summers, line 360:
```rust
let mut theta_count: u32 = if (condition) {
} else {
}
```

lockin/src/lib.rs
```diff
@@ -0,0 +1,973 @@
...
+fn adc_phases(
+    first_t: u32,
+    tstamps_mem: &mut [TimeStamp; 2],
+    phi: u32,
+    fscale: u32,
+    tadc: u32,
+) -> [f32; SAMPLE_BUFFER_SIZE] {
+    let overflow_count: u32 = tadc * SAMPLE_BUFFER_SIZE as u32;
+    let tref_count: u32 = tstamps_diff(tstamps_mem, overflow_count);
+    let mut thetas: [f32; SAMPLE_BUFFER_SIZE] = [0.; SAMPLE_BUFFER_SIZE];
```
ryan-summers, line 356: Place variable declarations immediately where they're used so referencing what they are is simpler.

lockin/src/lib.rs
```diff
@@ -0,0 +1,973 @@
...
+                tstamps_mem[0],
+            ],
+            overflow_count,
+        );
+    }
+
+    thetas[0] = real_phase(theta_count, fscale, tref_count, phi);
+    for i in 1..SAMPLE_BUFFER_SIZE {
+        theta_count += tadc;
+        thetas[i] = real_phase(theta_count, fscale, tref_count, phi);
```
ryan-summers, line 378:
```rust
thetas.iter().enumerate().map(|(i, theta)| {
    real_phase(theta_count + i * tadc, fscale, tref_count, phi)
});
```

or something similar

lockin/src/lib.rs
```diff
@@ -0,0 +1,973 @@
...
+                    count: 0,
+                    sequences_old: 0,
+                },
+                tstamps_mem[0],
+            ],
+            overflow_count,
+        );
+    }
+
+    thetas[0] = real_phase(theta_count, fscale, tref_count, phi);
```
ryan-summers, line 375: I would recommend a different name than `thetas`. Instead, why not make it `sample_phase` or something similar, so it's clear what it's referring to?

lockin/src/trig.rs
```diff
@@ -0,0 +1,146 @@
...
+#[path = "sinf.rs"]
+mod sinf;
+#[path = "k_sinf.rs"]
+mod k_sinf;
+#[path = "cosf.rs"]
+mod cosf;
+#[path = "k_cosf.rs"]
+mod k_cosf;
+#[path = "rem_pio2f.rs"]
+mod rem_pio2f;
```
ryan-summers, line 19: I think making a `trig/` folder with all of the routines in it would be more idiomatic and help to keep a clean directory structure as well.

lockin/src/trig.rs
```diff
@@ -0,0 +1,146 @@
...
+use cosf::cosf;
+
+/// Computes the sine of a value `theta` in the range [0, 2*PI).
+///
+/// # Arguments
+///
+/// * `theta` - Angle for which sine is computed. Must be in range
+/// [0,2*PI).
+pub fn sin(theta: f32) -> f32 {
+    return sinf(theta);
```
ryan-summers, line 31: Why all the wrappers? If you don't like the `sinf` name, you can just do 
```rust
use sinf::sinf as sin;
```

ryan-summers (changes requested):

lockin/src/lib.rs
```diff
@@ -0,0 +1,973 @@
...
+    pub fn new_count(&mut self, newval: u16) {
+        self.count = newval as u32;
+        self.sequences_old = 0;
+    }
+}
+
+/// Compute the reference signal phase increment for each ADC sample
+/// and use them to demodulate the ADC inputs.
+macro_rules! prefilt_no_decimate {
+    ( $adc_samples:expr, $ref_tstamps:expr, $valid_tstamps:expr, $phi:expr, $tadc:expr, $fscale:expr, $tstamps_mem:expr) => {{
```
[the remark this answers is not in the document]
matthuszagh: This is a macro because it can return prematurely when `ts_valid_count < 2`. That return needs to be from the function in which the macro is invoked. Is there a better way to handle this?

lockin/src/lib.rs
```diff
@@ -0,0 +1,973 @@
...
+/// # Latency: 138ns
+fn real_phase(
+    theta_count: u32,
+    fscale: u32,
+    period_count: u32,
+    phase_count: u32,
+) -> f32 {
+    let total_angle =
+        modulo::<u32>(theta_count * fscale + phase_count, period_count);
+    2. * PI * (total_angle as f32 / period_count as f32)
```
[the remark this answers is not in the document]
matthuszagh: Would you mind further explaining the use of integer phase here? It seems to me that this would only be possible if the external reference period were fixed and *known ahead of time*. This would allow creating a LUT with the same number of indices as the reference period (worth noting: this has the downside of requiring large LUTs for slow reference frequencies). However, I was under the impression that the external reference period is unknown and must be computed at runtime based on the timestamps. Can I treat the reference period as a compile-time input? Or, is there another way to accomplish this result using integer phase that I'm not considering?

lockin/src/lib.rs
matthuszagh: Makes sense. Do you think these are worth even documenting at all for this reason? I was planning to include them initially so you could see the major contributors to performance, but then remove them before this ultimately gets merged.

lockin/src/lib.rs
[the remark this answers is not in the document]
ryan-summers: I would highly recommend insted using a `Result()` return value. The function that calls this can then check the return value for an `Ok(data)` or an `Err(Error::NotReady)`. It cleans up the code a lot here.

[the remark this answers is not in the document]
matthuszagh: I've done a major structural rewrite of this code (in line with your suggestions) and so this (and a bunch of other parts of this code) have been changed. I'm still cleaning it up, but hopefully some commits will be ready to push soon.

lockin/src/cosf.rs
```diff
@@ -0,0 +1,86 @@
+/* origin: FreeBSD /usr/src/lib/msun/src/s_cosf.c */
```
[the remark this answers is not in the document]
matthuszagh: I've raised an [issue in libm](https://github.com/rust-lang/libm/issues/248) about this.

lockin/src/lib.rs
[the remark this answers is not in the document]
jordens: Several things:

* It doesn't make sense to ever do f32 modulo since `sin()` and `cos()` (when implemented correctly) will do that.
* Integer modulo is always exact. Contrast that with float modulo (i.e. convert to float first and then do modulo something) where you kill accuracy quickly.
* Take a look at the two cases and work through the math. One of them has know and fixed reference period.
* You don't need a big LUT. Interpolation is cheap (compared to the polynomial approach in the existing impls). Math can be used to estimate LUT size from accuracy requirement.
* Also since we know that `sin/cos` will be a performance factor, just take a look around how that's implemented usually when performance is key and accuracy is given. E.g. cmsis dsp has a simple 512-LUT with linear interpolation. That also matches what would typically be implemented in FPGAs.

jordens:
As mentioned in the review, I'd suggest the following:

* Prepare a minimal PR moving the IIR code into a dsp crate.
* Rewrite your DSP code using the review comments above. That should shrink it dramatically. Split it into manageable parts if possible. Make a (or multiple) PR adding it, using libm for trig.
* Factor out the libm/trig impl discussion into a separate issue.

lockin/src/lib.rs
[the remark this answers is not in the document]
jordens: Also I am a bit uncertain about the cmsis quality and I don't understand some design decisions there either. Why don't they remap into the first quadrant (is it really better to save a few sign flips an pay for 4 times the storage)? And they don't seem to handle values outside [0, 2pi) at all.
If I were to implement it with a LUT, I'd try copying the usual [FPGA implementation](https://github.com/m-labs/misoc/blob/master/misoc/cores/cossin.py). But I'm a bit inexperienced to predict whether this is faster than the polynomial approach.
From what I can see, because the phase handling should probably be integer, we'd want `cossin(i32) -> (f32, f32)`.
And in general I'd really like to know whether using integer math for all the DSP (not just phases/PLL) might be better alltogether. But let's just keep that in mind, not act on it now.

matthuszagh:
Thanks for the additional information @jordens. I've created a PR moving the iir module to a dsp crate [here](https://github.com/quartiq/stabilizer/pull/174).

I've done a full rewrite of this PR based on the feedback. As you expected, this shrinks the code quite a bit. I'll push this soon, once I've had some more time to clean it up and review it. This will use real phase and libm.

Once I've pushed the changes to this PR, I'll open a separate issue to discuss the demodulation phase + trig issue (integer vs real phase, trig implementation considerations, and whether a PLL might be cheaper), with associated analysis.

matthuszagh:
Closed, in favor of https://github.com/quartiq/stabilizer/pull/177.

PR #177, opened 2020-11-24
Lock-in

matthuszagh:
I'm closing https://github.com/quartiq/stabilizer/pull/173 and opening this PR, since I'm not keeping any commits from that PR.

# Usage

The first step is to initialize a `Lockin` instance with `Lockin::new()`. This provides the lock-in algorithms with necessary information about the demodulation and filtering steps, such as whether to demodulate with a harmonic of the reference signal and the IIR biquad filter to use. There are then 4 different processing blocks that can be used (these blocks are mutually independent):

* `demodulate` - Computes the phase of the demodulation signal corresponding to each ADC sample, uses this phase to compute the in-phase and quadrature demodulation signals, and multiplies these demodulation signals by the ADC-sampled signal. This is a method of `Lockin` since it requires information about how to modify the reference signal for demodulation.
* `filter` - Performs IIR biquad filtering of in-phase and quadrature signals. This is commonly performed on the in-phase and quadrature components provided by the demodulation step, but can be performed at any other point in the processing chain or omitted entirely. `filter` is a method of `Lockin` since it must hold onto the filter configuration and state.
* `decimate` - This decimates the in-phase and quadrature signals to reduce the load on the DAC output. It does not require any state information and is therefore a normal function (i.e., non-method).
* `magnitude_phase` - Computes the magnitude and phase of the component of the ADC-sampled signal whose frequency is equal to the demodulation frequency. This does not require any state information and is therefore a normal function.

# Considerations

* [As discussed](https://github.com/quartiq/stabilizer/pull/173#issuecomment-731751931), this uses real (i.e., non-integral) phase values and libm. Since libm's `sincosf` uses `f64`, it is not currently usable on stabilizer, which only supports `f32`. I will open a separate issue investigating alternatives.


dsp/src/lockin.rs
```diff
@@ -0,0 +1,534 @@
...
+        adc_samples: [i16; ADC_SAMPLE_BUFFER_SIZE],
+        timestamps: [u16; TIMESTAMP_BUFFER_SIZE],
+        valid_timestamps: u16,
+    ) -> Option<([f32; ADC_SAMPLE_BUFFER_SIZE], [f32; ADC_SAMPLE_BUFFER_SIZE])>
+    {
+        // update old timestamps for new ADC batch
+        let sample_period = self.sample_period as i32;
+        self.timestamps.iter_mut().for_each(|t| match *t {
+            Some(i) => {
+                *t = Some(i - ADC_SAMPLE_BUFFER_SIZE as i32 * sample_period);
```
ryan-summers, line 158: Replace `i` with a variable name like `timestamp` - it took me a moment to figure out what was being calculated here, so it may be worth adding a comment as well describing the calculation.

dsp/src/lockin.rs
```diff
@@ -0,0 +1,534 @@
...
+        });
+
+        // record new timestamps
+        timestamps
+            .iter()
+            .take(valid_timestamps as usize)
+            .rev()
+            .take(2)
+            .rev()
+            .for_each(|t| self.timestamps.push(Some(*t as i32)));
```
ryan-summers, line 170: Won't this infinitely grow `self.timestamps`? I'm not seeing a truncation step anywhere.

dsp/src/lockin.rs
```diff
@@ -0,0 +1,534 @@
...
+    /// `demodulate` expects that the timestamp counter value is equal
+    /// to 0 when the ADC samples its first input in a batch. This can
+    /// be achieved by configuring the timestamp counter to overflow
+    /// at the end of the ADC batch sampling period.
+    pub fn demodulate(
+        &mut self,
+        adc_samples: [i16; ADC_SAMPLE_BUFFER_SIZE],
+        timestamps: [u16; TIMESTAMP_BUFFER_SIZE],
+        valid_timestamps: u16,
+    ) -> Option<([f32; ADC_SAMPLE_BUFFER_SIZE], [f32; ADC_SAMPLE_BUFFER_SIZE])>
```
ryan-summers, line 152: I personally think this would make more sense as a `Result`, but not too opinionated

dsp/src/lockin.rs
ryan-summers: I guess I'm not following what `push` does here? Isn't `self.timestamps` just an array? How can we "push" to it?

dsp/src/lockin.rs
```diff
@@ -0,0 +1,534 @@
...
+    );
+
+    let mut in_phase_decimated = [0f32; DECIMATED_BUFFER_SIZE];
+    let mut quadrature_decimated = [0f32; DECIMATED_BUFFER_SIZE];
+
+    in_phase_decimated
+        .iter_mut()
+        .zip(quadrature_decimated.iter_mut())
+        .zip(in_phase.iter().step_by(n_k))
+        .zip(quadrature.iter().step_by(n_k))
```
ryan-summers, line 249: Is it possible to create a new zipped iterator of `in_phase` and `quadrature` simultaneously? E.g.

```rust
let real_signals = in_phase.iter().step_by(n_k)
                            .zip(quadrature.iter().step_by(n_k));

in_phase_decimated.iter_mut().zip(quadrature_decimated.iter_mut())
                                 .zip(real_signals)
                                 .for_each(|((i_d, q_d), (i, q))| {
    // ...
});
```

dsp/src/lockin.rs
```diff
@@ -0,0 +1,534 @@
...
+            let new_q = libm::atan2f(*q, *i);
+            *i = new_i;
+            *q = new_q;
+        });
+}
+
+/// Treat the 2-element array as a FIFO. This allows new elements to
+/// be pushed into the array, existing elements to shift back in the
+/// array, and the last element to fall off the array.
+trait Fifo2<T> {
```
ryan-summers, line 280: Move this to a point before it's used - I was very confused by what the `push()` was doing on the array at first, but seeing this makes it all clear.

ryan-summers (changes requested):
Some minor changes, but in general the code looks much better!

dsp/src/lockin.rs
ryan-summers: I found the trait below - would recommend moving the definition above so that others aren't as confused as I was

matthuszagh:
Thanks for the review @ryan-summers! I've made the changes. Let me know if you find anything else.

dsp/src/lockin.rs
matthuszagh: I like this - it looks cleaner.

ryan-summers (approved):
LGTM from a rust perspective. Will need to defer to @jordens on the math aspect of things

dsp/src/lockin.rs
```diff
@@ -0,0 +1,544 @@
...
+    /// # Assumptions
+    ///
+    /// `demodulate` expects that the timestamp counter value is equal
+    /// to 0 when the ADC samples its first input in a batch. This can
+    /// be achieved by configuring the timestamp counter to overflow
+    /// at the end of the ADC batch sampling period.
+    pub fn demodulate(
+        &mut self,
+        adc_samples: [i16; ADC_SAMPLE_BUFFER_SIZE],
+        timestamps: [u16; TIMESTAMP_BUFFER_SIZE],
```
ryan-summers, line 176: For simplicity, just have this function accept a slice of timestamps. That way, you don't need two arguments and you can accept a variable length array.

`timestamps: [u16]` is the API, then the number of valid timestamps is `timestamps.len()`

dsp/src/lockin.rs
```diff
@@ -0,0 +1,544 @@
...
+    ///
+    /// # Assumptions
+    ///
+    /// `demodulate` expects that the timestamp counter value is equal
+    /// to 0 when the ADC samples its first input in a batch. This can
+    /// be achieved by configuring the timestamp counter to overflow
+    /// at the end of the ADC batch sampling period.
+    pub fn demodulate(
+        &mut self,
+        adc_samples: [i16; ADC_SAMPLE_BUFFER_SIZE],
```
ryan-summers, line 175: This should be a reference, not ownership, because the ADC temporarily borrows out data buffers to prevent copying. We can't transfer ownership from the ADC module to the demodulate function.

dsp/src/lockin.rs
```diff
@@ -0,0 +1,544 @@
...
+    (in_phase_decimated, quadrature_decimated)
+}
+
+/// Compute the magnitude and phase from the in-phase and quadrature
+/// signals. The in-phase and quadrature arrays are modified in place.
+///
+/// # Arguments
+///
+/// * `in_phase` - In-phase signal.
+/// * `quadrature` - Quadrature signal.
```
ryan-summers, line 300: It's not clear where magnitude and phase are stored. Which magnitude and which phase?

I think this would be simpler if you took ownership fo the `in_phase` and `quadrature` arrays and then returned them with the new values as (magnitude, phase) arrays. It would still be in-place updates, but ownership would make it clear that the data has transformed.

E.g.
```rust
let (i, q) = //...;
let (mag, phase) = dsp::lockin::magnitude_phase(i, q);
```

dsp/src/lockin.rs
ryan-summers, line 175: Samples are acquired in a `u16` buffer and need to be re-interpretted as i16. The function should accept a borrowed array of u16 data.

jordens: That reinterpretation should be hidden in the ADC part. It's common for all use cases

jordens (changes requested):
I'd suggest a few changes:

1. Use (`num-complex`) or create a `Complex<f32>` type. Arrays/slices/iterators over it come automatically.
2. Use one IIR filter for both quadratures.
3. Let's try boiling down the iterators further. Being generic over iterators is probably going to be a bit hairy. Let's not try it. Maybe with the changes above the functions become so small that just chaining iterators in user code is actually the easiest and most flexible way. Could you give something like this a try (just a suggestion, untested, maybe some of these don't work)?

```rust
// feed timestamps into pll
timestamps.map(|t| pll.update(&mut pll_state, t)).last();
let last_phase = adc_samples.iter()
    .enumerate()
    .map(|j, x| {
        let phi = pll.get_sample_phase(j);
        let (cos_phi, sin_phi) = cossinf(phi);
        // demodulate
        let (xcp, xsp) = (x*cos_phi, x*sin_phi);
        // filter
        let i = iir.update(&mut iir_state.0, xcp);
        let q = iir.update(&mut iir_state.1, xsp);
        let wraps = unwrap_phase((i, q), &mut wrap_state);  // The processing should later also handle phase unwrapping (which needs two subsequent samples and a state). That unwrapping can and should be done before decimate, maybe even before filtering.
        ((i, q), wraps)
    })
    .step_by(n_k)  // decimate
    .map(|(i, q), wraps| atan2f(q, i) + wraps)  // compute the phase
    .last();  // just look at the last phase in this case
dac_samples.iter_mut()
    .enumerate()
    .for_each(|j, d| *d = last_phase + modulate_waveform[j]);  // update the DAC samples with both the feedback phase and the modulation waveform
```

dsp/src/lockin.rs
matthuszagh: The difficulty with this is that this could be performed on arrays of various sizes. For instance, `magnitude_phase` might be called after `decimate`. Alternatively `decimate` might be omitted.

dsp/src/lockin.rs
```diff
@@ -0,0 +1,544 @@
...
+    in_phase
+        .iter_mut()
+        .zip(quadrature.iter_mut())
+        .for_each(|(i, q)| {
+            let new_i = libm::sqrtf([*i, *q].iter().map(|i| i * i).sum());
+            let new_q = libm::atan2f(*q, *i);
+            *i = new_i;
+            *q = new_q;
+        });
+}
```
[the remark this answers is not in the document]
matthuszagh: Is the goal here to hold onto/pass around the iterator? I don't believe it's possible to `collect()` into an array.

dsp/src/lockin.rs
matthuszagh: Based on Robert's feedback, it looks like this will go away anyway.

dsp/src/lockin.rs
```diff
@@ -0,0 +1,544 @@
...
+            None => (),
+        });
+
+        // record new timestamps
+        timestamps
+            .iter()
+            .take(valid_timestamps as usize)
+            .rev()
+            .take(2)
+            .rev()
```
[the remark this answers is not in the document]
matthuszagh: The change won't work because there's no guarantee that we have at least 2 timestamps. `take(2)` takes up to 2 timestamps. However, the other complaint is a fair one. I did it this way because I thought we could assume a stable reference frequency and because using the closest 2 timestamps complicates the logic somewhat. But, I can change this to interpolate based on the closest 2 timestamps.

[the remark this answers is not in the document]
matthuszagh: I've changed this to:

```
let start_index: usize = if timestamps.len() < 2 {
   0
} else {
   timestamps.len() - 2
};
timestamps[start_index..]
   .iter()
   .for_each(|t| self.timestamps.push(Some(*t as i32)));
```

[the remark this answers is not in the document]
matthuszagh: The implementation now uses the closest 2 timestamps to the current ADC sample (rather than the most recent 2 timestamps) to compute the reference period.

matthuszagh:
@jordens I've implemented most of your suggested revisions. Currently, however, I'm still using arrays rather than iterators as function arguments and return values. Moreover, I haven't provided a user code solution that chains iterators. The reason for this is that computing the demodulation phase for each ADC sample by using the timestamps closest to the ADC sample complicates the demodulation logic somewhat, and the required user code would be somewhat lengthy. In other words, although some of the changes did decrease the amount of code in each function (e.g., using arrays of complex values instead of individual `f32` arrays), computing the ADC phases more than offset these decreases. It's worth noting that `demodulate` is the only non-trivial function at this point.

I expect that it will be worth revisiting the idea of chaining iterators in user code when the PLL is implemented.

.github/workflows/ci.yml
```diff
@@ -78,6 +78,25 @@ jobs:
...
+      - uses: actions/checkout@v2
+      - name: Install Rust ${{ matrix.toolchain }}
+        uses: actions-rs/toolchain@v1
+        with:
+          toolchain: ${{ matrix.toolchain }}
+      - name: cargo test
+        uses: actions-rs/cargo@v1
+        with:
+          command: test
+          args: --manifest-path dsp/Cargo.toml --target=x86_64-unknown-linux-gnu
```
jordens, line 98: `--package dsp`

dsp/src/testing.rs
```diff
@@ -0,0 +1,54 @@
...
+    f32_is_close(a.0, b.0) && f32_is_close(a.1, b.1)
+}
+
+pub fn complex_array_is_close(a: &[Complex<f32>], b: &[Complex<f32>]) -> bool {
+    let mut result: bool = true;
+    a.iter().zip(b.iter()).for_each(|(i, j)| {
+        result &= complex_is_close(*i, *j);
+    });
+    result
+}
```
jordens, line 17: Don't repeat yourself. ;)
These are use cases of the more general code below (`relative_tolerance = f32::EPS` and `fixed_tolerance = 0.`).

dsp/src/testing.rs
```diff
@@ -0,0 +1,54 @@
...
+
+pub fn within_tolerance(
```
jordens, line 19: Can we stay closer to existing namings, e.g. from numpy: `allclose()`, `rtol`, `atol` (not "fixed").

jordens (commented):
On top of the previous changes and the ones pending with the PLL code, a couple things.

dsp/src/lockin.rs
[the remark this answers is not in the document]
jordens: As mentioned above, passing iterators around might be tricky because the types become a bit complicated.
Let's just provide the building blocks for single-sample processing and then have the user use iterators in `process()` (the ISR in `main.rs`) where and when convenient. Like https://github.com/quartiq/stabilizer/pull/177#pullrequestreview-538698163

jordens:
> @jordens I've implemented most of your suggested revisions. Currently, however, I'm still using arrays rather than iterators as function arguments and return values. Moreover, I haven't provided a user code solution that chains iterators. The reason for this is that computing the demodulation phase for each ADC sample by using the timestamps closest to the ADC sample complicates the demodulation logic somewhat, and the required user code would be somewhat lengthy. In other words, although some of the changes did decrease the amount of code in each function (e.g., using arrays of complex values instead of individual `f32` arrays), computing the ADC phases offset these decreases. It's worth noting that `demodulate` is the only non-trivial function at this point. I've also left `magnitude_phase` in the code for the time being. This can, of course, easily be removed.

Ack. It's typically very good to reduce the amount of code. Remove trivial functions and reduce the interface of the non-trivial ones to make the universal.
Then with the PLL now this can be solved. Something like https://github.com/quartiq/stabilizer/pull/177#pullrequestreview-538698163 should be possible verbatim in `process()`. The `get_sample_phase()` will use the batch size and the frequency/phase from the pll `update()`.

I think we will also want to use the `IIR` instances in `main.rs` to do the filtering of the demodulated signal. They serve the same purpose AFAICS.

> I expect that it will be worth revisiting the idea of chaining iterators in user code when the PLL is implemented.

Yes.

> I've used a custom `Complex` implementation instead of `num-complex`. `num-complex` consists of a type-generic struct with `re` and `im` fields. It additionally provides many functions frequently used with complex numbers, which are either provided by `std` or `libm`. Whether we should use `num-complex` depends on how we decide to compute the trig functions (sincos and atan2) and the sqrt function. If we use floating point implementations contributed to `libm`, then it would probably make sense to use `num-complex`. Currently, though, the custom implementation is trivial (it is simply `pub type Complex<T> = (T, T);`). I felt it made sense to wait for a better understanding of these trig functions before adding the `num-complex` dependency.

Ack. Sounds good!

jordens:
Bors r+

jordens:
Bors retry

jordens:
bors cancel

jordens:
bors retry

PR #178, opened 2020-11-25
Feature/iir tweaks

jordens:
Some minor IIR tweaks and a couple more relevant compiler flags.

**Untested on hardware.**

### DIFF
--- a/.cargo/config
+++ b/.cargo/config
@@ -2,8 +2,14 @@
 runner = "gdb-multiarch -q -x openocd.gdb"
 rustflags = [
     "-C", "link-arg=-Tlink.x",
+# The target (below) defaults to cortex-m4
+# There currently are two different options to go beyond that:
+# 1. cortex-m7 has the right flags and instructions (FPU) but no instruction schedule yet
     "-C", "target-cpu=cortex-m7",
+# 2. cortex-m4 with the additional fpv5 instructions and a potentially
+# better-than-nothing instruction schedule
     "-C", "target-feature=+fp-armv8d16",
+# When combined they are equivalent to (1) alone
 ]
 
 [build]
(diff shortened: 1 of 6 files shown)


jordens:
@ryan-summers could you review and give it a quick test on hardware (see whether it behaves as it did before passing through your 1 kHz test signal -> ADC -> DAC -> scope)?

ryan-summers:
@jordens Programming this branch onto a stabilizer with a 1KHz 1V amplitude sinusoid into ADC0 shows an identical sinusoid on DAC0 output. I don't see any distortion or inconsistencies introduced as a result of these changes.

.cargo/config
```diff
@@ -1,6 +1,10 @@
 [target.'cfg(all(target_arch = "arm", target_os = "none"))']
 runner = "gdb-multiarch -q -x openocd.gdb"
-rustflags = ["-C", "link-arg=-Tlink.x"]
+rustflags = [
+    "-C", "link-arg=-Tlink.x",
+    "-C", "target-cpu=cortex-m7",
+    "-C", "target-feature=+fp-armv8d16",
```
ryan-summers, line 6: Why are we allowed to include amv8 floating point instructions for the Cortex-M7, which is an Armv7 CPU? Is this valid?

ryan-summers: It would be worth adding comments about why we can use the ARMv8 flag on an ARMv7 processor, since this is actually due to the underlying FPv5 implementation on the STM32H743. https://reviews.llvm.org/D5537

ryan-summers (changes requested):
Minor documentation request - was very confused about why we were enabling ARMv8 features until I did a deep dive in documentation and LLVM reviews.

.cargo/config
ryan-summers: Actually, I think what we should actually be doing is making sure that the compiler is getting the right definition of the FPU: `fpv5-dp-d16`. I think this may be a work-around because this isn't getting properly defined.

jordens: I don't think that exists (anymore).

ryan-summers: https://code.woboq.org/llvm/llvm/lib/Support/ARMTargetParser.cpp.html#192 indicates you may be right. Mind adding documentation about this here for posterity?

PR #180, opened 2020-11-26
Feature/adc dac io macros

### LINKED ISSUE #183: DAC1 DMA stream operates in circular mode
The DAC1 DMA stream is configured in `circular_buffer` mode, which means that it will loop once the transfer completes. This is not the desired behavior, as the DMA transfer should end after all data is transfered.

For existing applications, I do not believe this generates incorrect output because a new DMA transfer is scheduled before the looped data can be transmitted, but this could cause faults in the future.

CC @jordens who discovered this in https://github.com/quartiq/stabilizer/pull/180

jordens:
I wanted to try macros.
This moves the ADC and DAC DMA setup into macros reducing code footprint. Hopefully no functional changes there.
I didn't test this on hardware and I may have missed differences between `Adc0`/`Adc1` and `Dac0`/`Dac1`.

It removes the `AdcInputs` and `DacOutputs` structs and replaces them with tuples as they were just fan-outs/fan-ins.
It also does some minor tweaks in the `process()` ISR towards higher flexibility enforces some data patterns to help the compiler.

Differences missing:
* [x] `.transfer_complete_interrupt(true)` for `Adc1` only: needed
* [x] `.circular_buffer(true);` for `Dac1` only: close #183

### DIFF
--- a/src/adc.rs
+++ b/src/adc.rs
@@ -27,358 +27,184 @@ static mut SPI_START: [u16; 1] = [0x00];
 // The following global buffers are used for the ADC sample DMA transfers. Two buffers are used for
 // each transfer in a ping-pong buffer configuration (one is being acquired while the other is being
 // processed). Note that the contents of AXI SRAM is uninitialized, so the buffer contents on
-// startup are undefined.
+// startup are undefined. The dimension are `ADC_BUF[adc_index][ping_pong_index][sample_index]`.
 #[link_section = ".axisram.buffers"]
-static mut ADC0_BUF0: [u16; SAMPLE_BUFFER_SIZE] = [0; SAMPLE_BUFFER_SIZE];
+static mut ADC_BUF: [[[u16; SAMPLE_BUFFER_SIZE]; 2]; 2] =
+    [[[0; SAMPLE_BUFFER_SIZE]; 2]; 2];
 
-#[link_section = ".axisram.buffers"]
-static mut ADC0_BUF1: [u16; SAMPLE_BUFFER_SIZE] = [0; SAMPLE_BUFFER_SIZE];
-
-#[link_section = ".axisram.buffers"]
-static mut ADC1_BUF0: [u16; SAMPLE_BUFFER_SIZE] = [0; SAMPLE_BUFFER_SIZE];
-
-#[link_section = ".axisram.buffers"]
-static mut ADC1_BUF1: [u16; SAMPLE_BUFFER_SIZE] = [0; SAMPLE_BUFFER_SIZE];
-
-/// SPI2 is used as a ZST (zero-sized type) for indicating a DMA transfer into the SPI2 TX FIFO
-/// whenever the tim2 update dma request occurs.
-struct SPI2 {
-    _channel: sampling_timer::tim2::Channel1,
-}
-impl SPI2 {
-    pub fn new(_channel: sampling_timer::tim2::Channel1) -> Self {
-        Self { _channel }
-    }
-}
-
-// Note(unsafe): This structure is only safe to instantiate once. The DMA request is hard-coded and
-// may only be used if ownership of the timer2 channel 1 compare channel is assured, which is
-// ensured by maintaining ownership of the channel.
-unsafe impl TargetAddress<MemoryToPeripheral> for SPI2 {
-    /// SPI2 is configured to operate using 16-bit transfer words.
-    type MemSize = u16;
-
-    /// SPI2 DMA requests are generated whenever TIM2 CH1 comparison occurs.
-    const REQUEST_LINE: Option<u8> = Some(DMAReq::TIM2_CH1 as u8);
-
-    /// Whenever the DMA request occurs, it should write into SPI2's TX FIFO to start a DMA
-    /// transfer.
-    fn address(&self) -> u32 {
-        // Note(unsafe): It is assumed that SPI2 is owned by another DMA transfer and this DMA is
-        // only used for the transmit-half of DMA.
-        let regs = unsafe { &*hal::stm32::SPI2::ptr() };
-        &regs.txdr as *const _ as u32
-    }
-}
-
-/// SPI3 is used as a ZST (zero-sized type) for indicating a DMA transfer into the SPI3 TX FIFO
-/// whenever the tim2 update dma request occurs.
-struct SPI3 {
-    _channel: sampling_timer::tim2::Channel2,
-}
-impl SPI3 {
-    pub fn new(_channel: sampling_timer::tim2::Channel2) -> Self {
-        Self { _channel }
-    }
-}
-
-// Note(unsafe): This structure is only safe to instantiate once. The DMA request is hard-coded and
-// may only be used if ownership of the timer2 channel 2 compare channel is assured, which is
-// ensured by maintaining ownership of the channel.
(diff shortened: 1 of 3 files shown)


jordens:
Differences missing:
* [ ] `.transfer_complete_interrupt(true)` for `Adc1` only: needed
* [ ] `.circular_buffer(true);` for `Dac1` only. Why is that? @ryan-summers

ryan-summers:
> * [ ]  `.circular_buffer(true);` for `Dac1` only. Why is that? @ryan-summers

I believe this is a bug in the DAC outputs - these DMA streams should not be operating in circular mode. I believe it's not currently an issue because we schedule a new DMA transfer as soon as the last one completes, so the circular transfer configuration never gets a chance to generate incorrect outputs. However, this should likely be removed. I'll open an issue.

jordens:
Then that's a copy-paste bug par excellence, found and fixed by "don't repeat yourself" ;)

src/adc.rs
```diff
@@ -27,358 +27,195 @@ static mut SPI_START: [u16; 1] = [0x00];
 // The following global buffers are used for the ADC sample DMA transfers. Two buffers are used for
 // each transfer in a ping-pong buffer configuration (one is being acquired while the other is being
 // processed). Note that the contents of AXI SRAM is uninitialized, so the buffer contents on
-// startup are undefined.
+// startup are undefined. The dimension are `ADC_BUF[adc_index][ping_pong_index][sample_index]`.
```
ryan-summers, line 30: Nit: `dimensions`

src/adc.rs
```diff
@@ -27,358 +27,195 @@ static mut SPI_START: [u16; 1] = [0x00];
...
-        trigger_transfer.start(|_| {});
-
-        Self {
-            // Note(unsafe): The ADC0_BUF1 is "owned" by this peripheral. It shall not be used
-            // anywhere else in the module.
-            next_buffer: unsafe { Some(&mut ADC0_BUF1) },
-            transfer: data_transfer,
-            _trigger_transfer: trigger_transfer,
+static mut ADC_BUF: [[[u16; SAMPLE_BUFFER_SIZE]; 2]; 2] =
+    [[[0; SAMPLE_BUFFER_SIZE]; 2]; 2];
```
ryan-summers, line 33: If you type this as `0u16`, you can entirely omit the type signature of `ADC_BUF` I believe, which would make this a one-liner and may make it easier to read.

src/adc.rs
```diff
@@ -27,358 +27,195 @@ static mut SPI_START: [u16; 1] = [0x00];
...
+
+macro_rules! adc_input {
+    ($name:ident, $index:literal, $trigger_stream:ident, $data_stream:ident,
+     $spi:ident, $trigger_channel:ident, $dma_req:ident) => {
+        /// $spi is used as a ZST (zero-sized type) for indicating a DMA transfer into the SPI TX FIFO
```
ryan-summers, line 38: Nit: SPI is no longer a ZST, since it contains the (unused) timer channel for safety guarantees. (This was not introduced here, but we can fix it while updating this).

src/adc.rs
```diff
@@ -27,358 +27,195 @@ static mut SPI_START: [u16; 1] = [0x00];
...
+
+            /// Release a buffer of ADC samples to the pool.
+            ///
+            /// # Args
+            /// * `next_buffer` - Buffer of ADC samples to be re-used.
+            pub fn release_buffer(
+                &mut self,
+                next_buffer: &'static mut [u16; SAMPLE_BUFFER_SIZE],
+            ) {
+                self.next_buffer.replace(next_buffer); // .unwrap_none() https://github.com/rust-lang/rust/issues/62633
```
ryan-summers, line 214: Why restructure this into two functions? We were previously leveraging rust ownership semantics to guarantee this via borrowing, but now we're incurring a run-time overhead? Is this to explicitly make it clear when the buffer is released as opposed to implicitly allowing the borrow to terminate due to scope?

src/dac.rs
```diff
@@ -11,306 +11,147 @@ use super::{
...
-            None,
-            trigger_config,
-        );
-
-        Self {
-            transfer,
-            // Note(unsafe): This buffer is only used once and provided for the next DMA transfer.
-            next_buffer: unsafe { Some(&mut DAC0_BUF1) },
-            first_transfer: true,
+static mut DAC_BUF: [[[u16; SAMPLE_BUFFER_SIZE]; 2]; 2] =
```
ryan-summers, line 16: Same as ADC - this could theoretically be on one line.

src/dac.rs
```diff
@@ -11,306 +11,147 @@ use super::{
...
+                    while !self.transfer.get_transfer_complete_flag() {}
+                }
+
+                // Start the next transfer.
+                self.transfer.clear_interrupts();
+                let (prev_buffer, _) =
+                    self.transfer.next_transfer(next_buffer).unwrap();
+
+                // .unwrap_none() https://github.com/rust-lang/rust/issues/62633
+                self.next_buffer.replace(prev_buffer);
```
ryan-summers, line 150: I think it's safe to silently ignore the return value here - why add documentation around the unwrap?

src/main.rs
```diff
@@ -729,8 +728,7 @@ const APP: () = {
         sampling_timer.start();
 
         init::LateResources {
-            afe0: afe0,
-            afe1: afe1,
+            afes: (afe0, afe1),
```
ryan-summers, line 731: Nice change using tuples here. Makes a much nicer API

ryan-summers (changes requested):

src/dac.rs
jordens: There wasn't any documentation.
This makes it clear that we are expecting a `None` in return by design. And we can enforce it once that is stabilized.

src/adc.rs
jordens: Sure. This is just using the same style as before.

jordens: It doesn't work.

ryan-summers: Ah, that must not work in static mut scope then. That's rather annoying, but I guess we can live with how this looks. Sorry for leading you down the wrong path

src/adc.rs
jordens: It was. And also to keep up the symmetry with the DAC API to help me understand things and differential design.
Let's revert that and see whether there are further changes around that DMA API that may warrant this.

src/main.rs
```diff
@@ -747,34 +745,30 @@ const APP: () = {
...
+            c.resources.adcs.0.acquire_buffer(),
+            c.resources.adcs.1.acquire_buffer(),
+        ];
+        let dac_samples = [
+            c.resources.dacs.0.acquire_buffer(),
+            c.resources.dacs.1.acquire_buffer(),
+        ];
+
+        for channel in 0..adc_samples.len() {
+            for sample in 0..adc_samples[0].len() {
```
ryan-summers, line 759: should this be `0..adc_samples[channel].len()`? I don't think it should make a difference, but would make more sense logically.

ryan-summers (changes requested):
One minor fix - other than that, a smoke test with a 1KHz test signal on the input is showing proper operation of both ADC/DAC pairs

src/main.rs
jordens: It doesn't matter since arrays are to be homogeneous. Let's leave it.

Issue #147, opened 2020-10-07
[meta] use case analysis


jordens:
With Stabilizer we have a powerfull CPU sitting between ADCs, DACs and DDS channels. Since this is extremely generic it allows dozens of different use cases to be implemented. Implementing all or even many of them is tricky as there are lots of interdependencies, constraints and corner cases. The interactions can also lead to bottle-necks since they compete for the same resources (CPU time, latency). We need to come up with a clear set of high-value use cases and their requirement in terms of features to be implemented. Then we can decide whether features can be made available at compile-time or run-time and how that affects usability, testing and deployment. I'll try to start with the use cases here and then once the features become clearer, fork them into their own issues.

## Use cases

0. IIR control: D-part, I^2 part, notches, LP, HP
1. WMS: Lock-in, Modulation
2. MTS: Same as WMS from Stabilizer perspective
3. FMS/PDH: Demodulation with (Pounder) and then it's either DC (nothing new) or superhet-to-DC (then MTS/WMS)
4. Phase lock: Lock-in R/phi, phase unwrap
5. Complex MIMO lock: Some beat input, demodulate superhet with Pounder, feedback amplitude error on AOM amplitude and phase error on AOM FM/PM and slow frequency error on laser piezo/current/temperature DAC output.

## I/O Architecture

* The acquisition and transmission of ADC/DAC/DDS samples should be pushed into the peripherals as much as possible
* The ADC should run at max rate (allowed by peripherals and hardware), the processing interrupt can run at a submultiple.
* ADC samples should end up in a CPU-accessible buffer (timer+DMA+SPI) at least as long as the relative ADC sample/processing rate ratio.
* DAC samples should be take from a memory buffer (timer+DMA+SPI) at least as long as the relative DAC sample/processing rate ratio.
* DAC sample emission should be triggered by the processing routine or by fixed timer.
* DDS should transfer a memory block of register settings (one entire profile consisting of frequency, phase, and amplitude on all four channels) to the DDS chip via timer+DMA+QSPI at max throughput. Triggered by processing routine or timer.
* Two-level modulation of the DDS would also be useful but is somewhat restricted by our design (P0,P1 on non-timer pins)
* Digital I/O should also be handled by timer+DMA. This applies to the DDS SYNC_CLK PA0 (ETR and "reverse" timestamping of the free CPU clock), the digital front panel inputs (flags, timestamping).

## Processing blocks

[...]

PR #186, opened 2020-12-02
Adding support for Pounder DDS QSPI profile stream

### LINKED ISSUE #172: ad9959 little things
https://github.com/quartiq/stabilizer/blob/a7b6b5c796c09c15927e98c8553d0e80c2593d3a/ad9959/src/lib.rs#L130

Can be `or(Err(Error::Pin))` It hink (also other places).

https://github.com/quartiq/stabilizer/blob/a7b6b5c796c09c15927e98c8553d0e80c2593d3a/ad9959/src/lib.rs#L138

Is `delay_ms(0)` and can become `delay_us(1u8)` everywhere. `reference_clock` >= 1 MHz always.

ryan-summers:
This is an initial draft PR of the proposed updates to the high-speed QSPI data stream to Pounder.

Current measurements indicate that each profile serialization + staging process currently takes 300ns for a profile of 12 bytes. The actual QSPI data write is only 75ns for 3 32-bit words, so there is room for future improvement when the layout of each profile is known in advance.

TODO:
- [x] Document new modules

Reserved for future PR:
Define a "static" layout for the pounder profile to avoid the overhead of the serialization process. This would allow us to copy bytes into an array at known offsets and do it really quickly.

This fixes #172 and addresses part of the IO architecture in #147


src/main.rs
```diff
@@ -594,20 +598,61 @@ const APP: () = {
...
+                    let mut hrtimer = hrtimer::HighResTimerE::new(
+                        dp.HRTIM_TIME,
+                        dp.HRTIM_MASTER,
+                        dp.HRTIM_COMMON,
+                        ccdr.clocks,
+                        ccdr.peripheral.HRTIM,
+                    );
+
+                    // IO_Update should be latched for 50ns after the QSPI profile write. Profile writes
+                    // are always 16 bytes, with 2 cycles required per byte, coming out to a total of 32
```
ryan-summers, line 629: This is now configurable, but the _limit_ is 16 bytes. This is only limited by the QSPI FIFO size and the size used in the `ad9959` serializer, so it could theoretically change in the future and invalidate these numbers

ryan-summers:
DS4_QuickPrint14

Above is a capture of SCK going to the DDS and the IO_Update signal. As can be seen, IO_Update is properly toggling on for 50nS after the DDS profile has been written over SPI

ad9959/src/lib.rs
```diff
@@ -522,14 +479,143 @@ where
...
+        data[0] = register as u8;
+        data[1..][..value.len()].copy_from_slice(value);
+        self.index += value.len() + 1;
+    }
+
+    /// Get the serialized profile as a slice of 32-bit words.
+    ///
+    /// # Note
+    /// The serialized profile will be padded to the next 32-bit word boundary by adding dummy
+    /// writes to the CSR or FR2 registers.
```
jordens, line 595: I think `LSRR` is better for the 3-byte padding. I don't that ramping being used in streaming mode but `FR2 != 0` might be useful in the future in streaming mode.

ad9959/src/lib.rs
```diff
@@ -522,14 +479,143 @@ where
...
+    /// * `acr` - If provided, indicates the amplitude control register for the channels.
+    pub fn update_channels(
+        &mut self,
+        channels: &[Channel],
+        ftw: Option<u32>,
+        pow: Option<u16>,
+        acr: Option<u16>,
+    ) {
+        // The user should have provided something to update.
+        assert!(
```
jordens, line 556: That's not really required for functionality, right? I'd remove it.

jordens (changes requested):

ad9959/src/lib.rs
```diff
@@ -522,14 +479,143 @@ where
...
+            0 => {}
+            1 => {
+                // For a pad size of 1, we have to pad with 5 bytes to align things.
+                self.add_write(Register::CSR, &[(self.mode as u8) << 1]);
+                self.add_write(Register::FR2, &[0, 0, 0]);
+            }
+            2 => self.add_write(Register::CSR, &[(self.mode as u8) << 1]),
+            3 => self.add_write(Register::FR2, &[0, 0, 0]),
+
+            _ => panic!("Invalid"),
```
jordens, line 612: `unreachable!()`

src/pounder/dds_output.rs
```diff
@@ -0,0 +1,107 @@
...
+        for word in profile.iter() {
+            // Note(unsafe): We are writing to the SPI TX FIFO in a raw manner for performance. This
+            // is safe because we know the data register is a valid address to write to.
+            unsafe {
+                core::ptr::write_volatile(
+                    &regs.dr as *const _ as *mut u32,
+                    *word,
+                );
+            }
+        }
```
irwineffect, line 70: Before entering this writing loop, should we quickly check the FLEVEL or BUSY bits in the status register to let us know if writing would stall? I'm assuming we don't expect that to ever happen if everything goes as planned.

irwineffect (commented):
Should we use integer arithmetic for all the frequency math to avoid any issues of floating point inaccuracy? We start to see inaccuracies for f32 when dealing with any integer values greater than 2^24 (~16.7*10^6). Or do we not care?

jordens:
> Should we use integer arithmetic for all the frequency math to avoid any issues of floating point inaccuracy? We start to see inaccuracies for f32 when dealing with any integer values greater than 2^24 (~16.7*10^6). Or do we not care?

For precision and metrology applications we'll need to be able to set and get frequencies without rounding errors. Setting requires that the user can pass in (ultimately) the desired frequency as a fraction `p/q` of the reference frequency (`p,q` integer). `p` contains the PLL multiplier and the FTW, `q` is the max FTW. Getting requires giving the user that fraction.
For internal DSP it's much better to do the processing in machine units anyway. And those are integers (e.g. the FTW).
But also for many non-precision and non-metrology applications it's a bad idea to have the user supply and process fractions (and deal with the constraints on those!). For those applications `f32` in SI units will be sufficient and most convenient.
Then again even for some "harmless" applications (e.g. superhet or certain fiber length stabilizations) we need to deal with the exact frequency differences/ratios between two DDS channels. Also for those cases intermediate conversions from and to `f32` SI would be fatal.

ad9959/src/lib.rs
```diff
@@ -522,14 +479,143 @@ where
...
+        }
+        unsafe {
+            core::slice::from_raw_parts::<'a, u32>(
```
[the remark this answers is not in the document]
ryan-summers: The issue here is that we're returning a slice and not an array. The slice has to be owned by something to exist properly. If we consumed `self` to generate a `[u32; 4]` array here and then loaned a slice ref to that array, the array itself would actually go out of scope too soon (in this function return). Hence, we have to keep `self` alive for the entire lifetime of the u32 slice.

I've played around with it and haven't found a nicer way to move ownership around, but if you have an idea, feel free to suggest it

src/main.rs
```diff
@@ -594,20 +598,60 @@ const APP: () = {
...
+                    // QSPI clock cycles. The QSPI is configured for 40MHz, so this comes out to an
+                    // offset of 800nS. We use 900ns to be safe - note that the timer is triggered after
+                    // the QSPI write, which can take approximately 120nS, so there is additional
+                    // margin.
+                    hrtimer.configure_single_shot(
+                        hrtimer::Channel::Two,
+                        50_e-9,
```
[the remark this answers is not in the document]
ryan-summers: It's actually 4 SYNC_CLK cycles, where SYNC_CLK is 1/4 of the REF_CLK, so it's 100 MHz in the current config (DDS is running at 400MHz). I'm using 50ns here to give us some rounding error latency from the 40ns requirement. Updating the comments to indicate this.

ad9959/src/lib.rs
```diff
@@ -522,14 +479,143 @@ where
...
+
+        if let Some(acr) = acr {
+            self.add_write(Register::ACR, &acr.to_be_bytes());
+        }
+    }
+
+    /// Add a register write to the serialization data.
+    fn add_write(&mut self, register: Register, value: &[u8]) {
+        let data = &mut self.data[self.index..];
+        assert!(value.len() + 1 <= data.len());
```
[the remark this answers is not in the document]
ryan-summers: I am obviously traumatized from my C-programming time and hyper paranoid about array bounds. You are correct that this should be unnecessary.

src/hrtimer.rs
```diff
@@ -0,0 +1,121 @@
...
+        };
+
+        // The period register must be greater than or equal to 3 cycles.
+        let period = (source_cycles / divider as u32) as u16;
+        assert!(period > 2);
+
+        // We now have the prescaler and the period registers. Configure the timer.
+        self.timer
+            .timecr
+            .modify(|_, w| unsafe { w.ck_pscx().bits(divider + 4) });
```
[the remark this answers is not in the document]
ryan-summers: Added docs and fixed some bugs where the writes were unsound. Thanks for pointing this out.

src/pounder/dds_output.rs
ryan-summers: I added a warning if FLEVEL is not zero. I'm unconvinced we should fatally error, but we should inform the user at a minimum.

ad9959/src/lib.rs
ryan-summers: I agree FR2 isn't the best register - I've updated to LSRR, thanks for the suggestion

ad9959/src/lib.rs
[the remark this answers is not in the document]
jordens: Instead of the slice, maybe return (move) the array and a length.

src/main.rs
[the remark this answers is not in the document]
jordens: Yes. But the DDS could be configured with external clock and without PLL. Then it runs at 1 MHz and needs 4 µs to register the update. Ack the comment.

jordens (approved):

irwineffect (approved):

jordens:
Bors isn't behaving.

PR #195, opened 2020-12-08
Feature/digital input stamp

ryan-summers:
This PR implements support for asynchronous collection of timestamps on the DI0 input.

TODO:
 - [x] https://github.com/stm32-rs/stm32h7xx-hal/pull/173 should get merged into the `stm32-rs/dma` branch
- [x] Add manual configuration of sampling + timestamping timer prescalers + periods as per #170 
- [x] Verify the accurate collection of timestamp values


src/main.rs
```diff
@@ -287,9 +288,29 @@ const APP: () = {
...
+            let timer5 =
+                dp.TIM5.timer(1.khz(), ccdr.peripheral.TIM5, &ccdr.clocks);
+
+            // The time stamp timer must run at exactly a multiple of the sample timer based on the
+            // batch size. To accomodate this, we manually set the period identical to the sample
+            // timer, but use a prescaler that is `BATCH_SIZE` longer.
+            let mut timer = timers::TimestampTimer::new(timer5);
+            timer.set_period(sampling_timer.get_period());
+            timer.set_prescaler(
+                sampling_timer.get_prescaler() * SAMPLE_BUFFER_SIZE as u16,
```
jordens: The prescaler decreases resolution. It should be as low as possible. Instead increase the period.
Or is that not feasible?

src/digital_input_stamper.rs
```diff
@@ -0,0 +1,146 @@
...
+///!
+///! This module provides a means of timestamping the rising edges of an external reference clock on
+///! the DI0 with a timer value from TIM5.
+///!
+///! This module only supports input clocks on DI0 and may or may not utilize DMA to collect
+///! timestamps.
+///!
+///! # Design
+///! An input capture channel is configured on DI0 and fed into TIM5's capture channel 4. TIM5 is
+///! then run in a free-running mode with a configured frequency and period. Whenever an edge on DI0
```
jordens, line 11: Frequency is inverse period by definition. What do you want to say here?

src/digital_input_stamper.rs
```diff
@@ -0,0 +1,146 @@
...
+///! triggers, the current TIM5 capture value is recorded as a timestamp. This timestamp can be
```
jordens, line 12:
```suggestion
///! triggers, the current TIM5 counter value is captured and recorded as a timestamp. This timestamp can be
```

src/digital_input_stamper.rs
```diff
@@ -0,0 +1,146 @@
...
+///! either directly read from the timer channel or can be collected asynchronously via DMA
+///! collection.
+///!
+///! When DMA is used for timestamp collection, a DMA transfer is configured to collect as many
+///! timestamps as there are samples, but it is intended that this DMA transfer should never
+///! complete. Instead, when all samples are collected, the module pauses the DMA transfer and
+///! checks to see how many timestamps were collected. These collected timestamps are then returned
+///! for further processing.
+///!
+///! To prevent silently discarding timestamps, the TIm5 input capture over-capture interrupt is
```
jordens, line 22:
```suggestion
///! To prevent silently discarding timestamps, the TIM5 input capture over-capture interrupt is
```

src/digital_input_stamper.rs
```diff
@@ -0,0 +1,146 @@
...
+}
+
+impl InputStamper {
+    /// Construct the DI0 input timestamper.
+    ///
+    /// # Args
+    /// * `trigger` - The capture trigger input pin.
+    /// * `stream` - The DMA stream to use for collecting timestamps.
+    /// * `timer_channel - The timer channel used for capturing timestamps.
+    /// * `batch_size` - The number of samples collected per processing batch.
```
jordens, line 68: Probably better to call this 
```suggestion
    /// * `buffer_size` - The maximum number of timestamps collected per processing batch.
```

to distinguish between the (variable) number of timestamps in the buffer and (fixed) number of ADC/DAC samples in the batch.

It makes sense to be able to force the non-DMA mode even for larger batches.

src/timers.rs
```diff
@@ -0,0 +1,221 @@
...
+                /// Get the timer capture/compare channels.
+                pub fn channels(&mut self) -> [< $TY:lower >]::Channels {
+                    self.channels.take().unwrap()
+                }
+
+                /// Get the prescaler of a timer.
+                #[allow(dead_code)]
+                pub fn get_prescaler(&self) -> u16 {
+                    let regs = unsafe { &*hal::stm32::$TY::ptr() };
+                    regs.psc.read().psc().bits() + 1
```
jordens, line 39: What if `PSC = 0xffff` (which I think is valid)?

src/main.rs
```diff
@@ -281,15 +285,57 @@ const APP: () = {
...
+
+            // The time stamp timer must run at exactly a multiple of the sample timer based on the
+            // batch size. To accomodate this, we manually set the period identical to the sample
+            // timer, but use a prescaler that is `BATCH_SIZE` longer.
+            let mut timer = timers::TimestampTimer::new(timer5);
+
+            let period: u32 = {
+                let batch_duration: u64 =
+                    SAMPLE_BUFFER_SIZE as u64 * ADC_SAMPLE_TICKS as u64;
+                let batches_per_overflow: u64 =
```
ryan-summers, line 326: We actually need to calculate the maximum power-of-two that is less-than-or-equal to batches_per_overflow and use that here.

jordens (changes requested):

src/main.rs
ryan-summers: This was some legacy implementation before our conversation today - it's also no longer present.

src/timers.rs
ryan-summers: These functions are no longer necessary, so I'm going to remove them. This was a bit of a hack early on to gain raw access to the prescaler

src/digital_input_stamper.rs
ryan-summers: Is it still appropriate to use the samples per batch as an upper-bound on the number of timestamps? We still use `SAMPLE_BUFFER_SIZE` to declare the size of the DMA transfer (and thus the maximum buffer_size).

src/digital_input_stamper.rs
ryan-summers: This is referring to the period register (ARR) and the prescaler (tick frequency), I'll think of better wording

ryan-summers:
As a test case, I input a 10Hz square wave (3V amplitude) to DI0 and collected 32 timestamps. I then calculated the delta between the timestamps and printed it using GDB:
```
892                     info!("All stamps gathered: {:?}", deltas);
(gdb) p/x deltas
$1 = [0x9894d3, 0x9894d2, 0x9894d2, 0x9894d2, 0x989dc7, 0x9894d2, 0x9894d2, 0x9894d1, 0x9894d1, 
  0x9894d1, 0x9894d0, 0x9894cf, 0x9894d0, 0x9894d0, 0x989dc5, 0x9894d0, 0x9894d1, 0x9894d1, 0x9894d3, 
  0x9894d2, 0x9894d2, 0x9894d2, 0x9894d1, 0x9894d1, 0x989dc6, 0x9894d0, 0x9894d1, 0x9894d2, 0x9894d2, 
  0x9894d2, 0x9894d3]
(gdb) p deltas
$2 = [9999571, 9999570, 9999570, 9999570, 10001863, 9999570, 9999570, 9999569, 9999569, 9999569, 
  9999568, 9999567, 9999568, 9999568, 10001861, 9999568, 9999569, 9999569, 9999571, 9999570, 9999570, 
  9999570, 9999569, 9999569, 10001862, 9999568, 9999569, 9999570, 9999570, 9999570, 9999571]
(gdb) 
```

As can be seen, the deltas are all approximately near the count of 10,000,000, which corresponds with 10ms with the timestamp timer running at 100MHz. The variations here are likely due to inaccuracies in my clock source as opposed to fluctuations in the timer timestamp captures.

src/digital_input_stamper.rs
jordens: As discussed I would actually lean towards ripping out the DMA here and just say that there will be max 1 timestamp per batch.

jordens:
> As a test case, I input a 10Hz square wave (3V amplitude) to DI0 and collected 32 timestamps. I then calculated the delta between the timestamps and printed it using GDB:

Nice!
But do check the max input range of the DIO (from schematics). I don't know whether it deals with negative voltages.

src/digital_input_stamper.rs
```diff
@@ -0,0 +1,76 @@
...
+    ///
+    /// To prevent timestamp loss, the batch size and sampling rate must be adjusted such that at
+    /// most one timestamp will occur in each data processing cycle.
+    pub fn latest_timestamp(&mut self) -> Option<u32> {
+        if self.capture_channel.check_overcapture() {
+            panic!("DI0 timestamp overrun");
+        }
+
+        self.capture_channel.latest_capture()
+    }
```
irwineffect, line 75: The reference manual specifies:

> In order to handle the overcapture, it is recommended to read the data before the
overcapture flag. This is to avoid missing an overcapture which could happen after reading
the flag and before reading the data.

I believe `latest_capture()` should read the capture register, then read/clear the interrupt flag, then check the overcapture flag. I think it would be best to move `check_overcapture()` inside `latest_capture()` rather than do it in this function to ensure in the future that all reads of the capture register are properly checked.

`latest_capture()` then could return a `Result<Option<u32>, E>` where E would indicate that an overcapture occurred.

src/digital_input_stamper.rs
```diff
@@ -0,0 +1,76 @@
...
+///!
+///! This module provides a means of timestamping the rising edges of an external reference clock on
+///! the DI0 with a timer value from TIM5.
+///!
+///! # Design
+///! An input capture channel is configured on DI0 and fed into TIM5's capture channel 4. TIM5 is
+///! then run in a free-running mode with a configured tick rate (PSC) and maximum count value
+///! (ARR). Whenever an edge on DI0 triggers, the current TIM5 counter value is captured and
+///! recorded as a timestamp. This timestamp can be either directly read from the timer channel or
+///! can be collected asynchronously via DMA collection.
```
irwineffect, line 11: If we use DMA to collect the timestamp, wouldn't we also want another DMA transaction to clear the interrupt flag to prevent the overcapture flag from being set? Or would we just ignore the overcapture flag in this scenario?

irwineffect (changes requested):

src/design_parameters.rs
```diff
@@ -4,3 +4,6 @@ pub const ADC_SETUP_TIME: f32 = 220e-9;
 
 /// The maximum DAC/ADC serial clock line frequency. This is a hardware limit.
 pub const ADC_DAC_SCK_MHZ_MAX: u32 = 50;
+
+/// The optimal counting frequency of the hardware timers used for timestamping and sampling.
+pub const TIMER_FREQUENCY_MHZ: u32 = 100;
```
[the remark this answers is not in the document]
ryan-summers: I was avoiding this because the `500.mhz()` is not a `const-fn` so it can't be called in the `const` variables. However, I think we could instead wrap all these up with a manual `MegaHertz{100}` to construct the object instead. I like that idea

src/main.rs
```diff
@@ -54,8 +56,9 @@ use smoltcp::wire::Ipv4Address;
 
 use heapless::{consts::*, String};
 
-// The desired sampling frequency of the ADCs.
-const SAMPLE_FREQUENCY_KHZ: u32 = 500;
+// The number of ticks in the ADC sampling timer. The timer runs at 100MHz, so the step size is
+// equal to 10ns per tick.
+const ADC_SAMPLE_TICKS: u32 = 256;
```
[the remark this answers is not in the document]
ryan-summers: This is intended to be user configurable, but there is a desire for it to be a multiple of two. @jordens I haven't taken a close look at the math, but is there a mathematical simplification that we get from specifying sample rates this way? Otherwise James is right and it can be somewhat opaque to figure out the real sample rate.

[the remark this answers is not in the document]
jordens: Right. Let's add the resulting sample rate to the comment (10/256 MHz). The reason for ticks is to get an accurate ratio to the other timers (without rounding errors) and for a power of two is accurate and simpler math (#170).

src/digital_input_stamper.rs
ryan-summers: The solution there would likely be to listen for the overcapture flag and generate an interrupt if it ever occurs. Either that, or check the overcapture whenever servicing the completion of the DMA capture.

ryan-summers: In any case, since we aren't using DMA here, this isn't an issue.

jordens (approved):
Good from my end.

src/main.rs
```diff
@@ -629,25 +667,25 @@ const APP: () = {
...
-                        900_e-9,
+                        design_parameters::POUNDER_IO_UPDATE_DURATION,
+                        design_parameters::POUNDER_IO_UPDATE_DELAY,
                     );
 
                     // Ensure that we have enough time for an IO-update every sample.
+                    let sample_frequency = (design_parameters::TIMER_FREQUENCY.0
+                        as f32
+                        * 1_000_000.0)
+                        / ADC_SAMPLE_TICKS as f32;
```
irwineffect, line 683: Now that `design_parameters::TIMER_FREQUENCY` is a `MegaHertz` I was wanting to rewrite it as 
```Rust
(Hertz::from(design_parameters::TIMER_FREQUENCY).0 as f32)                  
```
But it looks like the HAL implemented the `Into` traits instead of the `From` traits...I wonder if that was intentional? Convention is to implement `From` because `Into` can automatically be derived from `From`. Because of that, I think we need to use this less pretty syntax:
```Rust                                                       
(Into::<Hertz>::into(design_parameters::TIMER_FREQUENCY).0 as f32)          
```
Or perhaps
```Rust
let sample_frequency: Hertz = design_parameters::TIMER_FREQUENCY.into();
let sample_frequency = (sample_frequency.0 as f32) / (ADC_SAMPLE_TICKS as f32);
```

src/timers.rs
```diff
@@ -0,0 +1,218 @@
...
+                        None
+                    };
+
+                    // If there is an overcapture, return an error.
+                    if sr.[< cc $index of >]().bit_is_clear() {
+                        Ok(result)
+                    } else {
+                        Err(())
+                    }
+                }
```
irwineffect, line 170: I think we might want to read the SR twice, like so:
```Rust                                        
let sr1 = regs.sr.read(); // Read SR to check if CCxIF is set                   
let ccx = regs.[< ccr $index >].read();                                         
let sr2 = regs.sr.read(); // Read SR again to make sure CCxOF was not set       
                                                                                
let result = if sr1.[< cc $index if >]().bit_is_set() {                                          
    Some(ccx.ccr().bits())                                                         
} else {                                                                        
    None                                                                        
};                                                                              
                                                                                
// If there is an overcapture, return an error.                                 
if sr2.[< cc $index of >]().bit_is_clear() {                                    
    Ok(result)                                                                  
} else {                                                                        
    Err(())                                                                     
}  
```

I think we want to check the CCxIF bit to make sure the capture register has been updated before we read the capture register. But then we then want to check the CCxOF bit after we have read the capture register to make sure an overcapture didn't happen, which could have invalidated our read of the capture register.

According to the reference manual, we don't need to explicitly clear the CCxIF flag, this happens automatically when we read the capture register.

However, what about the scenario where the CCxIF flag is set, therefore the capture register is valid, we read it, but then another capture comes in (basically we just barely read in time before an overcapture occurred)? Do we want to detect and handle that case? Should we allow for returning two timestamps, or just treat it as if an overcapture occurred, as that likely indicates a timing problem in the code?

Should we clear the CCxOF bit when an overcapture occurs to return things to a good state?

irwineffect (changes requested):

src/timers.rs
ryan-summers: > However, what about the scenario where the CCxIF flag is set, therefore the capture register is valid, we read it, but then another capture comes in (basically we just barely read in time before an overcapture occurred)? Do we want to detect and handle that case? Should we allow for returning two timestamps, or just treat it as if an overcapture occurred, as that likely indicates a timing problem in the code?

I don't think that's indicative of ill-operating code. In fact, I think that would indicate perfectly functional code. As long as the capture is never lost, there's no issue with the data. We will get around to the timestamp in the future.

> Should we clear the CCxOF bit when an overcapture occurs to return things to a good state?

It doesn't matter right now (since we panic if there is ever an error), but you're right. I've updated it to clear the flag before returning the error.

irwineffect (approved):

PR #196, opened 2020-12-09
Feature/pounder timestamping

ryan-summers:
This PR implements timestamping using the external pounder SYNC_CLK to drive an internal timer. This addresses the IO architecture in #147 and the timestamping logic outlined in #170 

This finalizes the asynchronous IO architecture and timestamp acquisition features.

These changes require a hardware revision of Pounder before they can be used:
- https://github.com/sinara-hw/Pounder/issues/76

TODO:
- [x] Update documentation
- [x] https://github.com/quartiq/stabilizer/pull/195 should be merged
- [x] https://github.com/stm32-rs/stm32h7xx-hal/pull/173
- [x] https://github.com/stm32-rs/stm32h7xx-hal/pull/171
- [x] Add feature for V1.1 pounder support (changes to DDS_Reset and SYNC_CLK pins)

### Testing

I manually updated the `process()` function to record timestamps and calculate delays between each sample. Visual inspection showed a nominal 40 tick delay between each sample, where each tick corresponds to 32ns (1/4 of a 125MHz clock). Given that the ADCs were configured to sample every 128 ticks of a 100MHz clock, the delays of both clocks align to ~1.28uS/sample, which indicates that the pounder clock is correctly timestamping ADC samples.


jordens:
Is there a way we can merge this without regression on existing v1.0 Pounders? Maybe this is the time for `#[cfg(feature = pounder_v1_1)]` or similar?

ryan-summers:
> Is there a way we can merge this without regression on existing v1.0 Pounders? Maybe this is the time for `#[cfg(feature = pounder_v1_1)]` or similar?

Yeah, that shouldn't be too hard. I'll add this to the checklist above.

dtcallcock:
Given https://github.com/sinara-hw/Pounder/issues/84#issuecomment-741893894, it's not clear that more than ~2 boards were ever made.  If you already patched one to v1.1 and it'll be straightforward to do the other(s), it might not be worth it.

jordens:
> Given [sinara-hw/Pounder#84 (comment)](https://github.com/sinara-hw/Pounder/issues/84#issuecomment-741893894), it's not clear that more than ~2 boards were ever made. If you already patched one to v1.1 and it'll be straightforward to do the other(s), it might not be worth it.

It's worth it for us.

src/timers.rs
```diff
@@ -29,6 +29,8 @@ pub enum TriggerSource {
 }
 
 pub enum Prescaler {
```
jordens, line 31: Isn't there already such an enum/variant in the PAC for that?
https://stm32-rs.github.io/stm32-rs/STM32H743x.html#TIM2:SMCR:ETPS

src/design_parameters.rs
```diff
@@ -7,3 +7,12 @@ pub const ADC_DAC_SCK_MHZ_MAX: u32 = 50;
...
 pub const TIMER_FREQUENCY_MHZ: u32 = 100;
+
+/// The DDS reference clock frequency in MHz.
+pub const DDS_REF_CLK_MHZ: u32 = 100;
+
+/// The multiplier used for the DDS reference clock PLL.
+pub const DDS_MULTIPLIER: u8 = 5;
+
+/// The rate of the DDS SYNC_CLK in MHz is always 1/4 that of the internal PLL clock.
+pub const DDS_SYNC_CLK_MHZ: u32 = DDS_REF_CLK_MHZ * DDS_MULTIPLIER as u32 / 4;
```
jordens, line 18: I'd refactor this as `DDS_SYNC_CLK_DIV: u32 = 4`. And we only need that once below. So it could be inlined.

src/timers.rs
ryan-summers: Yes, but the PAC-specific value is not generic over all possible timer types (e.g. there's a separate, equivalent value for TIM1, etc.), so I've made it generic  here.

src/design_parameters.rs
ryan-summers: I believe since these are all `const`, they should actually be evaluated at compile-time, and thus always inlined. I'll go ahead and add a definition though, since it will make it clear where the magic number is coming from.

jordens: I'm not sure that the result will always be the same (independent of whether this is compile-time or run-time). Operator precedence can lead to different rounding with f32. As explained elsewhere it's best work in integer multipliers and dividers as much as possible and only convert to f32 where really needed. E.g. for the timestamping the only intrinsic `f32` is `pub const DDS_REF_PER_TIMER: f32 = DDS_REF_MHZ/TIMER_CLK_MHz`. The rest are fractions/multiples of that.

src/timers.rs
irwineffect: Perhaps a PR should be made for the PAC?

irwineffect (approved):

src/design_parameters.rs
ryan-summers: I don't know if I'm following you here. All the numbers in this are unsigned integers - there's no floating point operation being performed. By substituting in the constants, you can see that the calculation is:
```rust
pub const DDS_SYNC_CLK_MHZ: u32 = 100_u32 * 5_u8 as u32 / 4
```

Writing that out, I'm noticing there's ambiguity in the integer math, so I'm adding a parenthesis (although Rust guarantees left-> right analysis of * and / operators)

jordens: I was comparing using the multipliers/dividers as integers (and avoiding the calculation and comparison of new SI frequencies/periods) with rounding. Yes it's not f32. In that regard i32 MHz is even worse than f32 Hz. It would be very wrong for e.g.  `DDS_REF_CLK_MHZ = 25`, `DDS_MULTIPLIER=28`.

src/timers.rs
jordens: Ah. Hmm. I don't know what's easiest/best then.

src/design_parameters.rs
ryan-summers: I wasn't really following still until I went and looked at the code fora  refactor and I now see what you are getting at. I've updated it to not calculate the SYNC_CLK using integer math - it's now offloaded into a float before any division occurs.

ryan-summers: Closing as this should now be resolved. Thanks for pointing it out

irwineffect (approved):

jordens (approved):

PR #199, opened 2020-12-10
Feature/cossin tuneup

jordens:
* shrink the LUT by another bit
* correctly use the octant bit to offset the dphi to LUT entry midpoint
* add more diagnistics to the unittest and rewrite it in relative units
* MSB-align phase and output to match the PLL data, dynamic range and
  remove the need for roudning bias.
* clean up the build.rs table generator a bit

Now the quadratic error (the little parabolas), the LUT data quantization error (the scatter) and the phase quantization error (not visible here) all three are of similar magnitude. The max and rms quadrature errors are now 1e-5 and 6e-8 relative to full scale over all possible 20 bit values.

Error in the first octant
image
FFT at some 120 dBc
image


matthuszagh (approved):
Looks good to me.

FYI: [godbolt](https://rust.godbolt.org/z/hb6ecn)

PR #200, opened 2020-12-11
dsp: add host benchmark

jordens:


jordens:
bors try

jordens:
bors merge

.gitignore
```diff
@@ -1,3 +1 @@
 /target
-/dsp/target
```
matthuszagh, line 2: Was this change intentional?

jordens: Yes. Not needed with workspaces iirc.

Issue #206, opened 2020-12-15
atan2


matthuszagh:
This post first provides an analysis of floating-point implementations. I will update it later with an analysis of a fixed-point implementation.

# Floating point

An atan2 implementation should meet the following requirements:

1. It should have high accuracy near the axes (at the output values -pi, -pi/2, 0, pi/2, pi) and coarse accuracy away from the axes. Specifically, it should have a maximum absolute error of 1e-5 on the axes and 1e-1 midway between the axes (at pi/4, 3pi/4, etc.). The acceptable error is linearly interpolated between the minima and maxima.
2. The error should be sufficiently smooth. Specifically, `abs(absolute_error(atan2(y+dy, x+dx)) - absolute_error(atan2(y,x))) < 1e-5 for dx,dy<1e-5`.

The optimal procedure for meeting these criteria seems to be:

1. Map the inputs to new inputs that yield an output angle between -pi/4 and pi/4. 
2. Compute an approximation of `atan` that works well for these angles.
3. Compensate for the mapping.

The mapping and compensation is shown in the following table:

```
| angle          | equivalent atan expression in [-pi/4,pi/4] |
|----------------+--------------------------------------------|
| [-pi/4,pi/4]   | atan(y/x)                                  |
| [-pi/2,-pi/4]  | -pi/2 - atan(x/y)                          |
| [pi/4,pi/2]    | pi/2 - atan(x/y)                           |
| [-3pi/4,-pi/2] | -pi/2-atan(x/y)                            |
| [pi/2,3pi/4]   | pi/2-atan(x/y)                             |
| [-pi,-3pi/4]   | -pi+atan(y/x)                              |
| [3pi/4,pi]     | pi+atan(y/x)                               |
```

There are a number of different options for the atan approximation. The first 3 are taken from [Rajan 2006: Efficient Approximations for the Arctangent Function](http://www-labs.iro.umontreal.ca/~mignotte/IFT2425/Documents/EfficientApproximationArctgFunction.pdf). (4) is taken from [this Mathematics StackExchange answer](http://www-labs.iro.umontreal.ca/~mignotte/IFT2425/Documents/EfficientApproximationArctgFunction.pdf).
1. `pi/4 * x`
2. `pi / 4 * x + 0.285 * x * (1 - abs(x))`
3. `x / (1.0 + 0.28086 * x ** 2)`
[...]

PR #207, opened 2020-12-17
atan

matthuszagh:
Adds 2-argument arctangent function. Parameters and result are `i32` integers for fast computation. Only the 16 MSBs of the inputs are used (16 LSBs are discarded). The x and y inputs can range from -1 to 1, which corresponds to `i32::MIN` and `i32::MAX`, respectively. The output ranges from -pi to pi, which corresponds to `i32::MIN` and `i32::MAX`, respectively.

- [godbolt](https://rust.godbolt.org/z/nahKrT)

# Related
- #206


dsp/src/iir.rs
```diff
@@ -1,24 +1,14 @@
...
 
+use super::abs;
 use core::f32;
 
 // These are implemented here because core::f32 doesn't have them (yet).
 // They are naive and don't handle inf/nan.
 // `compiler-intrinsics`/llvm should have better (robust, universal, and
 // faster) implementations.
 
-fn abs<T>(x: T) -> T
```
jordens, line 11: I think the `copysign` etc can also be elevated like this.

dsp/src/trig.rs
```diff
@@ -143,4 +228,40 @@ mod tests {
         assert!(max_err.0 < 1.1e-5);
         assert!(max_err.1 < 1.1e-5);
     }
+
+    #[test]
+    fn atan2_absolute_error() {
```
jordens, line 233: We need to make sure that we explicitly test the corner case inputs as well: min, max, -1, 0, 1.

dsp/src/trig.rs
```diff
@@ -1,8 +1,85 @@
...
+
+    if max == 0 {
+        return 0;
+    }
+
+    let ratio = (min << 15) / max;
+
+    let mut angle = {
+        // pi/4, referenced to i16::MAX
+        const PI_4_FACTOR: i32 = 25735;
```
jordens, line 54: Compute it, and let the compiler do that at compile time. See `cossin()`.

dsp/src/trig.rs
```diff
@@ -1,8 +1,85 @@
...
+        // 0.285, referenced to i16::MAX
+        const FACTOR_0285: i32 = 9339;
+        // 1/pi, referenced to u16::MAX
+        const PI_INVERTED_FACTOR: i32 = 20861;
```
jordens, line 58: These are all "factors". Just say what they are, e.g. `PI_OVER_4` (or `PI_4`), `ZERO_285` (or just `K`), `ONE_OVER_PI` (or `INV_PI`) and let the compiler compute them.

dsp/src/trig.rs
```diff
@@ -1,8 +1,85 @@
...
+
+        let r1 = shift_round(ratio * PI_4_FACTOR, 15);
+        let r2 = shift_round(
+            (shift_round(ratio * FACTOR_0285, 15)) * (i16::MAX as i32 - ratio),
+            15,
+        );
+        (r1 + r2) * PI_INVERTED_FACTOR
```
jordens, line 65: No need to round twice and then add. This should be possible to do with fewer `shift_round()`.

Also please just simplify the entire expression first. The 1/pi can be pulled in. That will simplify it significantly.

dsp/src/trig.rs
```diff
@@ -1,8 +1,85 @@
...
+///
+/// The angle between the x-axis and the ray to the point (x,y). The
+/// result range is from i32::MIN to i32::MAX, where i32::MIN
+/// corresponds to an angle of -pi and i32::MAX corresponds to an
+/// angle of +pi.
+pub fn atan2(y: i32, x: i32) -> i32 {
+    let y = y >> 16;
+    let x = x >> 16;
+
+    let ux = abs::<i32>(x);
```
jordens, line 28: Do you need to specify the type here.

dsp/src/trig.rs
```diff
@@ -143,4 +228,40 @@ mod tests {
         assert!(max_err.0 < 1.1e-5);
         assert!(max_err.1 < 1.1e-5);
     }
+
+    #[test]
```
jordens, line 232: I'd like a host benchmark as well. See `dsp/benches/cossin.rs`.

dsp/src/trig.rs
```diff
@@ -1,8 +1,85 @@
...
+/// The angle between the x-axis and the ray to the point (x,y). The
+/// result range is from i32::MIN to i32::MAX, where i32::MIN
+/// corresponds to an angle of -pi and i32::MAX corresponds to an
+/// angle of +pi.
+pub fn atan2(y: i32, x: i32) -> i32 {
+    let y = y >> 16;
+    let x = x >> 16;
+
+    let ux = abs::<i32>(x);
+    let uy = abs::<i32>(y);
```
jordens, line 29: I would try helping the compiler a bit. You need those three octant bits (`x < 0`, `y < 0`, `uy > ux`) later again.

dsp/src/trig.rs
```diff
@@ -1,8 +1,85 @@
...
+    //
+    // which is taken from Rajan 2006: Efficient Approximations for
+    // the Arctangent Function.
+    let (min, max) = if ux < uy { (ux, uy) } else { (uy, ux) };
+
+    if max == 0 {
+        return 0;
+    }
+
+    let ratio = (min << 15) / max;
```
jordens, line 50: A word on the 15 would be helpful. I know why but it takes time to derive it.

jordens (changes requested):

dsp/src/trig.rs
```diff
@@ -1,8 +1,85 @@
...
+///
+/// * `y` - Y-axis component.
+/// * `x` - X-axis component.
+///
+/// # Returns
+///
+/// The angle between the x-axis and the ray to the point (x,y). The
+/// result range is from i32::MIN to i32::MAX, where i32::MIN
+/// corresponds to an angle of -pi and i32::MAX corresponds to an
+/// angle of +pi.
```
jordens, line 23: Is that really correct? `i32::MAX != -i32::MIN`.

dsp/src/trig.rs
```diff
@@ -41,39 +48,40 @@ pub fn atan2(y: i32, x: i32) -> i32 {
     //
     // which is taken from Rajan 2006: Efficient Approximations for
     // the Arctangent Function.
-    let (min, max) = if ux < uy { (ux, uy) } else { (uy, ux) };
+    let (min, max) = if y_greater { (x, y) } else { (y, x) };
```
jordens, line 51: Now that you don't use `y` and `x` anymore below, just

```suggestion
    if y_greater {
        core::mem::swap(&mut x, &mut y);
     }
```

and keep using those.

dsp/src/trig.rs
jordens: I think it's actually `i32::MIN == -pi` and at the same time also `i32::MIN == pi`. Then `i32::MAX == pi*(1 - 1/(1 << 31))`

jordens (commented):
Minor nit.
Looks good!

dsp/src/trig.rs
matthuszagh: I'll add min and max. -1,0,1 are already included because `NUM_VALS` is odd. Would you like me to make this more clear with a comment, or something else?

jordens: All good like that

jordens:
Bors r+

dsp/src/trig.rs
matthuszagh: Oh, right that makes sense. It's also worth noting that `i32::MIN` is not actually reachable. I can get `i32::MAX` with `atan2(0, i32::MIN)`. The closest results are 2147461318 and -2147461318 with `atan2(1 << 16, i32::MIN))` and `atan2(-1 << 16, i32::MIN))`.

jordens:
Bors r+

Issue #166, opened 2020-11-12
Measure run-time task load


ryan-summers:
As part of the stabilizer firmware, it has been brought up that it would be useful to be able to measure and report "task-load" for the various interrupt tasks in RTIC.

This would then be used similar to a `top` process indicating how much time each processing task is using. This could then be reported as diagnostic telemetry.

There may be built-in functionality in RTIC or by using the ITM to accomplish this.

jordens:
Something simpler and easier than but not unlike the FreeRTOS [uxTaskGetSustemState()/vTaskGetInfo()](https://www.freertos.org/uxTaskGetSystemState.html).

ryan-summers:
I asked about this on the RTIC Matrix room and it was indicated that RTIC does not have this information inherently, but it would be something that they may be interested in adding to the framework.

It was suggested that `embedded-time` could be used and provided to RTIC to automatically calculate task durations. It's likely that this would require an RFC to RTIC if we want to get it integrated into the framework

jordens:
Could maybe be a trampoline that wraps the ISRs and records their cycle usage. Or alternatively some explicit hooks that the tasks need to call explicitly on entry and exit to be recorded.

jordens:
https://github.com/rtic-scope

jordens:
sufficiently good solutions exist with `itm trace` we pushed to `probe-rs`.
closing for now.
Might even get better once/if we build a stack sampler and flamegraphs.

PR #208, opened 2020-12-17
Adding documentation, updating DAC output timing

### LINKED ISSUE #171: Document IO architecture
As part of the IO redesign, we need to describe the following:
* How IO processing works
* What peripherals/resources are being used
* What triggers and timers are employed
* Design tradeoffs with other implementations
* Potential future work

### LINKED ISSUE #185: First DAC output code may not be synchronized with the sampling timer
The very first output code when scheduling a set of DAC output codes may not occur at the timer roll-over event. Instead, the first sample will be generated immediately when the samples are scheduled. It will then stay active until the next roll-over and the second code will be generated.

This is likely caused by the DMA request being asserted before the transfer is started. Thus, immediately after the stream is enabled, the request is handled (as opposed to waiting for the next compare event to generate a request).

To resolve this, we will need to ensure that there is no pending DMA request for the channel before enabling the output stream or we will need to ensure that we are scheduling DAC outputs at the sample timer rollover point.

DS4_QuickPrint13

In the above image, CH4 indicates when ADC and DAC DMA transfers are swapped (first pulse is ADC, second is DAC). CH1 is the DAC0 output and CH2 is DAC0. The sample timer rate can be seen to be the distinct, regular transitions in the outputs. Immediately at the DAC DMA update, it can be seen that a spuriously short DAC output code is generated as a result of this bug.

ryan-summers:
This PR adds extensive documentation to the asynchronous IO modules on stabilizer. This fixes #171 

This PR also updates the DAC codes so that DAC + ADC batches are always synchronous (DAC codes are exactly 2 batches after the ADC sample that generated them). In this implementation, DAC batches are always delayed by exactly 1 additional batch size from the original ADC sample batch that generated them.

There are thus two DAC batches with fixed output-zero codes at the very start of stabilizer processing.

This PR fixes #185 

TODO:
- [x] Verify timing on stabilizer hardware
- [x] Investigate updating DAC DMA to use DBM to prevent potential sample loss


jordens:
Pretty sure the delay is two batches, not one. You'd measure from first sample of an ADC batch to (corresponding) first sample of a DAC batch.

jordens (changes requested):

src/adc.rs
```diff
@@ -1,18 +1,75 @@
...
+///!
+///!
+///! # Note
+///!
+///! While there are two ADCs, only a single ADC is configured to generate transfer-complete
+///! interrupts. This is done because it is assumed that the ADCs will always be sampled
+///! simultaneously. If only a single ADC is used, it must always be ADC0, as ADC1 will not generate
+///! transfer-complete interrupts.
+///!
+///! There is a very small amount of latency between sampling of ADCs due to bus matrix priority. As
```
irwineffect, line 63: How might we be able to calculate this latency?

src/dac.rs
```diff
@@ -1,8 +1,54 @@
...
+///!
+///! In order to ensure alignment between the ADC sample batches and DAC output code batches, a DAC
+///! output batch is always exactly 1 batch after the ADC batch that generated it.
+///!
+///! The DMA transfer for the DAC output codes utilizes a double-buffer mode to avoid losing any
+///! transfer events generated by the timer (for example, when 2 update cycles occur before the DMA
+///! transfer completion is handled). In this mode, there should always be a valid buffer in the
+///! "next-transfer" double-buffer location for the DMA transfer. Once a transfer completes,
+///! software then has exactly one batch duration to properly configure the next buffer before it
+///! begins transfer.
```
irwineffect, line 24: What happens if this timing constraint is missed?

src/dac.rs
```diff
@@ -1,8 +1,54 @@
...
+///! For some applications, it may be desirable to generate a single DAC code from multiple ADC
+///! samples. In order to maintain timing characteristics between ADC samples and DAC code outputs,
+///! applications are required to generate one DAC code for each ADC sample. To accomodate mapping
+///! multiple inputs to a single output, the output code can be repeated a number of times in the
+///! output buffer corresponding with the number of input samples that were used to generate it.
+///!
+///!
+///! # Note
+///!
+///! There is a very small amount of latency between updating the two DACs due to bus matrix
```
irwineffect, line 37: How much latency? We describe it in terms of "timer ticks" below, but nano- or microseconds would also be nice.

irwineffect:
Something that is implied but not plainly explained is how the ADC/DAC constraints sum up to create the constraints of the DSP application code. Something like:

> Processing time for the DSP application code is bounded by the following constraints:
> 
> DSP application code starts after the ADC has generated a batch of samples, and must be completed by the time the next batch of ADC samples has been acquired (plus the FIFO buffer time). If this constraint is not met, <what will happen?>.
> 
> The DSP application code must also fill out the next DAC output buffer in time such that the DAC can switch to it when it has completed the current buffer. If this constraint is not met <what will happen?>
> 
> Because the ADC and DAC operate at the same rate, these two constraints actually implement the same time bounds, meeting one also means the other is also met.

How will the end user know that their DSP application code meets these constraints? Once this code is stabilized I imagine that we can provide equations for how to precisely calculate how much time the DSP application has based on parameters like sampling rate and batch size. 

Will we eventually have some sort of burden reporting mechanism so the end user can test code and see how much more they can add before hitting one of the limits?

irwineffect:
Concerning my burden question, it seems like https://github.com/quartiq/stabilizer/issues/166 might address it?

src/adc.rs
```diff
@@ -1,18 +1,75 @@
...
+///! The SPI peripheral internally has an 8- or 16-byte TX and RX FIFO, which corresponds to a 4- or
+///! 8-sample buffer for incoming ADC samples. During the handling of the DMA transfer completion,
+///! there is a small window where buffers are swapped over where it's possible that a sample could
+///! be lost. In order to avoid this, the SPI RX FIFO is effectively used as a "sample overflow"
+///! region and can buffer a number of samples until the next DMA transfer is configured. If a DMA
+///! transfer is still not set in time, the SPI peripheral will generate an input-overrun interrupt.
+///! This interrupt then serves as a means of detecting if samples have been lost, which will occur
+///! whenever data processing takes longer than the collection period.
+///!
+///!
```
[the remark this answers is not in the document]
ryan-summers: Correct - it does not increase us to 4 + batch_size because DBM means that there's always a DMA transfer configured with the SPI peripheral. Since a DMA transfer is always draining the SPI, the internal SPI FIFO can never fill beyond a single sample.

Thus, if we have a batch size of 1-3, we would actually have less time to service the DMA transfer completion in DBM

src/adc.rs
ryan-summers: The latency is 1 bus matrix clock. However, since the bus might be used by for example the CPU, it might get delayed further. Because this latency is in the range of a handful of AXI bus clocks, that's what I used the phrase "a very small amount"

ryan-summers: Essentially, the latency is not deterministic, hence why it's not clearly specified here

src/adc.rs
```diff
@@ -1,18 +1,75 @@
...
+///! While there are two ADCs, only a single ADC is configured to generate transfer-complete
+///! interrupts. This is done because it is assumed that the ADCs will always be sampled
+///! simultaneously. If only a single ADC is used, it must always be ADC0, as ADC1 will not generate
+///! transfer-complete interrupts.
+///!
+///! There is a very small amount of latency between sampling of ADCs due to bus matrix priority. As
+///! such, one of the ADCs will be sampled marginally earlier before the other because the DMA
+///! requests are generated simultaneously. This can be avoided by providing a known offset to the
+///! sample DMA requests, which can be completed by setting e.g. ADC0's comparison to a counter
+///! value of 0 and ADC1's comparison to a counter value of 1.
```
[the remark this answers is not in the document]
ryan-summers: The ADC0/ADC1 never race for the bus because of natural priority of the DMA transfers (since they are using the same DMA peripheral). This means the DMA peripheral arbitrates between ADC0 and ADC1 internally in a deterministic manner.

src/dac.rs
ryan-summers: This is the same as with the ADCs. It's actually an undeterministic number of AXI bus cycles, hence why it's unspecified.

ryan-summers:
@irwineffect I've added the blurb you provided to describe the DSP timing constraints - thanks for writing it up. In regard to knowing whether an application is meeting timing:

The ADC inputs listen for an overrun on the SPI peripheral, which generates an interrupt that panics. Thus, if timing is _not_ met, the firmware will panic and make it obvious to the end user that timing is not being met. Ultimately, #166 will also aid with this as you point out.

src/adc.rs
irwineffect: Could we specify some sort of worst case number? This allows someone to make the determination of "I don't care about this latency".

irwineffect (approved):

jordens (approved):

irwineffect (approved):

ryan-summers:
Timing for batch sizes > 1 was verified. It was found that batch_size = 1 results in an extra batch delay from input -> output, and I haven't been able to figure out the cause. I spawned an issue in https://github.com/quartiq/stabilizer/issues/229 so we can track it as necessary.

PR #213, opened 2021-01-04
pll update input is named "x" not "input"

matthuszagh:


jordens:
bors merge

Thanks.

jordens:
We should pin clippy

jordens:
Bors merge

PR #222, opened 2021-01-09
Lockin

matthuszagh:
Integrate PLL and integer math to lock-in functionality.


dsp/src/lockin.rs
```diff
@@ -302,11 +263,13 @@ pub fn decimate(
...
-/// * `signal` - Complex signal to decimate.
-pub fn magnitude_phase(signal: &mut [Complex<f32>]) {
+/// * `signal` - Complex signal for which the magnitude and phase
+/// should be computed. TODO currently, we compute the square of the
+/// magnitude. This should be changed to be the actual magnitude.
+pub fn magnitude_phase(signal: &mut [Complex<i32>]) {
     signal.iter_mut().for_each(|s| {
-        let new_i = libm::sqrtf([s.0, s.1].iter().map(|i| i * i).sum());
-        let new_q = libm::atan2f(s.1, s.0);
+        let new_i = [s.0, s.1].iter().map(|i| i * i).sum();
```
jordens, line 271: This is longer and harder to understand than `s.0 * s.0 + s.1 * s.1`.

jordens: Also again. We don't always need the magnitude output. Leave this to the user in `process()`.

dsp/src/lockin.rs
```diff
@@ -135,120 +129,89 @@ impl Lockin {
...
     /// an insufficient number of timestamps to perform processing,
     /// `Err` is returned.
-    ///
-    /// # Assumptions
-    ///
-    /// `demodulate` expects that the timestamp counter value is equal
-    /// to 0 when the ADC samples its first input in a batch. This can
-    /// be achieved by configuring the timestamp counter to overflow
-    /// at the end of the ADC batch sampling period.
     pub fn demodulate(
```
jordens, line 142: I don't understand why this is still called `demodulate`. It does so much more.

* There seems to be the need for a bit of code (either a struct or just a `fn`) that does the timestamp handling (the branching, the division) before the PLL. Is it better to wrap the PLL to encapsulate all that (`update(Option<timestamp>) -> `(demodulation_frequency, demodulation_phase)`)?
* Then there is some code that does the sample processing. It seems prudent to not allocate and generate the complete demodulation signal and then do the mixing, and filtering, each with a batched loop with intermediate buffers. But instead move the code up into a single loop, if necessary within `process()`. The API is tiny now and it seems it doesn't need more wrapping. Also in the end the different use cases will look different. There may be some that will decimate, others may not. Yet others might have two biquads in chain. And there might be the `Unwrapper` as well. Some might take the phase, others might be interested only in the amplitude.

Maybe a good suggestion to understand what I mean would be for you to try to actually use the code in the processing ISR in `main.rs`. That should show whether these wrappers are actually convenient or whether the code would be much shorter and easier to understand if they were inlined.

dsp/src/lockin.rs
```diff
@@ -52,47 +52,36 @@
...
-    fn push(&mut self, new_element: Option<T>) {
-        // For array sizes greater than 2 it would be preferable to
-        // use a rotating index to avoid unnecessary data
-        // copying. However, this would somewhat complicate the use of
-        // iterators and for 2 elements, shifting is inexpensive.
-        self[1] = self[0];
-        self[0] = new_element;
-    }
-}
+pub const ADC_BATCHES_LOG2: usize =
```
jordens, line 69: This deserves a comment explaining what it means.

dsp/src/lockin.rs
```diff
@@ -52,47 +52,36 @@
...
+use super::iir_int::{IIRState, IIR};
+use super::pll::PLL;
+use super::trig::{atan2, cossin};
+use super::{divide_round, Complex};
 
-/// The number of ADC samples in one batch.
-pub const ADC_SAMPLE_BUFFER_SIZE: usize = 16;
-/// The number of outputs sent to the DAC for each ADC batch.
-pub const DECIMATED_BUFFER_SIZE: usize = 1;
+/// TODO these constants are copied from main.rs and should be
```
jordens, line 60: I think if you remove the wrappers and move code into the `process()` ISR in `main.rs` then these will move as well and the TODO is resolved.

jordens (changes requested):
I would suggest you try to actually use the code in the processing ISR in main.rs. That should show whether these wrappers are actually convenient or whether the code would be much shorter and easier to understand if they were inlined.

The moving of the batch and timer constants here doesn't seem right. `dsp` should be standalone and should not dictate the application. Inlining the code may quickly make the TODO redundant.

dsp/src/lockin.rs
```diff
@@ -102,31 +91,36 @@ impl Lockin {
...
-            phase_offset: phase_offset,
-            sample_period: sample_period,
-            harmonic: harmonic,
-            timestamps: [None, None],
-            iir: iir,
-            iirstate: [[0.; 5]; 2],
+            harmonic,
+            phase_offset,
+            batch_index: 0,
+            last_phase: None,
```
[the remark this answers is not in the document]
matthuszagh: Can you clarify this point? `0` is a valid last phase value.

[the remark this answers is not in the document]
jordens: I mean the initial state. I'd like to get rid of the option and the branching. Instead just set some initial phase and frequency that's consistent with the PLL's initial state (which will be 0, 0).

dsp/src/lockin.rs
matthuszagh: I've removed the lockin module completely and moved this code into main. A new `TimestampHandler` receives timestamps, wraps the PLL and computes the demodulation frequency and initial phase. Everything else is done in `process()`. Let me know if this is more along the lines of what you had in mind.

src/main.rs
```diff
@@ -60,14 +60,26 @@ use heapless::{consts::*, String};
...
 
 // The desired ADC sample processing buffer size.
-const SAMPLE_BUFFER_SIZE: usize = 8;
+const SAMPLE_BUFFER_SIZE_LOG2: usize = 3;
+const SAMPLE_BUFFER_SIZE: usize = 1 << SAMPLE_BUFFER_SIZE_LOG2;
+
+// The number of ADC batches in one timer overflow period.
+pub const ADC_BATCHES_LOG2: usize =
+    32 - SAMPLE_BUFFER_SIZE_LOG2 - ADC_SAMPLE_TICKS_LOG2 as usize;
+pub const ADC_BATCHES: usize = 1 << ADC_BATCHES_LOG2;
```
jordens, line 73: I think it's less code if you inline these.

src/main.rs
```diff
@@ -145,6 +161,79 @@ type AFE1 = afe::ProgrammableGainAmplifier<
...
+    }
+
+    /// Compute the initial phase and frequency of the demodulation signal.
+    ///
+    /// # Args
+    /// * `timestamp` - Counter value corresponding to an external reference edge.
+    ///
+    /// # Returns
+    /// Tuple consisting of the initial phase value and frequency of the demodulation signal.
+    pub fn update(&mut self, timestamp: Option<u32>) -> (u32, u32) {
```
jordens, line 204: I'd set `ADC_SAMPLE_TICKS_LOG2` and `SAMPLE_BUFFER_SIZE_LOG2` as struct members (passed in `new()`) and move the struct and the impl into `dsp/src/reciprocal_pll.rs`.

src/main.rs
```diff
@@ -226,10 +318,10 @@ const APP: () = {
...
 
         // Format: iir_state[ch][cascade-no][coeff]
-        #[init([[[0.; 5]; IIR_CASCADE_LENGTH]; 2])]
-        iir_state: [[iir::IIRState; IIR_CASCADE_LENGTH]; 2],
-        #[init([[iir::IIR { ba: [1., 0., 0., 0., 0.], y_offset: 0., y_min: -SCALE - 1., y_max: SCALE }; IIR_CASCADE_LENGTH]; 2])]
-        iir_ch: [[iir::IIR; IIR_CASCADE_LENGTH]; 2],
+        #[init([[[0; 5]; IIR_CASCADE_LENGTH]; 2])]
+        iir_state: [[IIRState; IIR_CASCADE_LENGTH]; 2],
+        #[init([[IIR { ba: [1, 0, 0, 0, 0] }; IIR_CASCADE_LENGTH]; 2])]
+        iir_ch: [[IIR; IIR_CASCADE_LENGTH]; 2],
```
jordens, line 324: These IIRs should remain `f32` for now.

src/main.rs
```diff
@@ -205,6 +294,9 @@ const APP: () = {
         adcs: (Adc0Input, Adc1Input),
         dacs: (Dac0Output, Dac1Output),
         input_stamper: digital_input_stamper::InputStamper,
+        timestamp_handler: TimestampHandler,
+        iir_lockin: IIR,
+        iir_state_lockin: [IIRState; 2],
```
jordens, line 299: Correct. This IIR should be `i32`.

jordens (changes requested):
Nice!

Once `TimestampHandler` is factored out into `dsp` (see below), let's move the remaining changes to `main.rs` into a new PR (e.g. `lockin-bin`). Then this `dsp` PR can land right now and we can use the `lockin-bin` as a trial for the multiple-bin refactoring.

jordens (approved):
Nice.

Issue #218, opened 2021-01-08
Support selective feature usage and multiple binary support


ryan-summers:
Currently, stabilizer has many features enabled by default in a single binary, such as:
* Two ADC channels
* Two DAC channels
* DI0 timestamping
* [Future] Pounder DDS timestamping
* Two/four IIR filters

Not all applications require all of these features, and each functionality incurs run-time overhead that reduces available DSP time. As such, features that are not required should be easy to disable and/or omit from the binary.

Additionally, stabilizer needs to support multiple binaries in the future, and we should not be duplicating setup code across the various applications.

This issue encompasses providing a generic (library based) means of configuring stabilizer hardware for all applications, but also encompasses allowing for the configured features (mentioned above) to be selectively enabled/disabled based on the application needs.

PR #226, opened 2021-01-13
Lockin binary

matthuszagh:
Initial implementation of a lock-in binary. This is part of the move to make separate binaries for different signal processing flows.

# Related
- #218 
- This PR is motivated by the comments in #222.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -60,13 +60,20 @@ use heapless::{consts::*, String};
 // The number of ticks in the ADC sampling timer. The timer runs at 100MHz, so the step size is
 // equal to 10ns per tick.
 // Currently, the sample rate is equal to: Fsample = 100/256 MHz = 390.625 KHz
-const ADC_SAMPLE_TICKS: u16 = 256;
+const ADC_SAMPLE_TICKS_LOG2: u16 = 8;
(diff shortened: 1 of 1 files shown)


jordens (approved):
Nice. I'll merge this.

In a follow-up PR, could you add an integration test along the sames lines as this code to `dsp/tests/lockin.rs` and make sure it is hit with CI?
Before you had some good code that generates a signal (desired frequency plus some interferrer) and checks the correct demodulated amplitude and the rejection of the interferer. I think the harness for all that can be pretty small and compact. Feel free to use all of `std` or other array/numerics crates for this.

jordens:
Note that I've merged this into `quartiq:lockin-bin` to stage it for the multi-bin changes.

matthuszagh:
> In a follow-up PR, could you add an integration test along the sames lines as this code to `dsp/tests/lockin.rs` and make sure it is hit with CI?
> Before you had some good code that generates a signal (desired frequency plus some interferrer) and checks the correct demodulated amplitude and the rejection of the interferer. I think the harness for all that can be pretty small and compact. Feel free to use all of `std` or other array/numerics crates for this.

Definitely, will do.

PR #230, opened 2021-01-18
Refactoring to support multiple apps

### LINKED ISSUE #218: Support selective feature usage and multiple binary support

ryan-summers:
This PR closes #218 by configuring all hardware interfaces to a default-off state. They may be enabled by using the associated `start()` function, which allows for selective usage of adcs/dacs/timestampers. Additionally, all hardware configuration has been moved to `src/hardware/configuration.rs` to facilitate easily writing multiple applications without copying the initialization code.

@jordens Would you mind taking a look at the structure? I haven't touched the `idle` task as I believe that will be changing significantly as soon as we refactor to use MQTT and the new settings configurations.

## Testing
I ran a square wave input signal to ADC0 and measured the expected output signal on DAC0 with the expected delay.


jordens (approved):
Nice.
I think that we'll want to have the files with `src/main.rs -> bin/dual-iir.rs` and then `src/lib.rs` collecting all the exports. Like https://github.com/knurling-rs/app-template/tree/main/src

PR #240, opened 2021-01-26
Adding internal lock-in integration demo

ryan-summers:
This PR adds a demo application for #60 case (a).

This is currently a work-in progress and requires hardware testing to verify.


src/bin/lockin-internal-demo.rs
```diff
@@ -0,0 +1,294 @@
...
+            let y = value * i16::MAX as f32;
+            // Note(unsafe): The DAC_SEQUENCE values are guaranteed to be normalized.
+            let y = unsafe { y.to_int_unchecked::<i16>() };
+
+            // Convert to DAC code
+            dac_samples[0][i] = y as u16 ^ 0x8000;
+        }
+
+        // TODO: Verify that the DAC code is always generated at T=0
+        let (pll_phase, pll_frequency) = c.resources.pll.update(Some(0));
```
jordens, line 119:
```suggestion
        let pll_phase = 0i32;
        let pll_frequency = 1i32 << (32 - 3); // 1/8 of the sample rate
```

src/bin/lockin-internal-demo.rs
```diff
@@ -0,0 +1,294 @@
...
+        dacs: (Dac0Output, Dac1Output),
+        net_interface: hardware::Ethernet,
+
+        #[init([0.; 5])]
+        iir_state: iir::IIRState,
+
+        #[init(iir::IIR { ba: [1., 0., 0., 0., 0.], y_offset: 0., y_min: -SCALE - 1., y_max: SCALE })]
+        iir: iir::IIR,
+
+        pll: TimestampHandler,
```
jordens, line 44:
```suggestion
```

src/bin/lockin-internal-demo.rs
```diff
@@ -0,0 +1,294 @@
...
+        // Configure the microcontroller
+        let (mut stabilizer, _pounder) = hardware::setup(c.core, c.device);
+
+        let pll = TimestampHandler::new(
+            4, // relative PLL frequency bandwidth: 2**-4, TODO: expose
+            3, // relative PLL phase bandwidth: 2**-3, TODO: expose
+            ADC_SAMPLE_TICKS_LOG2 as usize,
+            SAMPLE_BUFFER_SIZE_LOG2,
+        );
+
```
jordens, line 59:
```suggestion
```

src/bin/lockin-internal-demo.rs
```diff
@@ -0,0 +1,294 @@
...
+        let lockin = Lockin::new(
+            &iir_int::IIRState::default(), // TODO: lowpass, expose
```
jordens, line 61:
```suggestion
            &iir_int::IIRState([21101, 42202, 21101, 2137942692, -1064243070]),
```

is a 1e-3 corner frequency butterworth lowpass

jordens (changes requested):
Without having tested anything, these are the most obvious changes needed.

src/bin/lockin-internal-demo.rs
ryan-summers: Would you mind providing context as to how the values were generated so I can add a reference in the code?

jordens: This is `lowpass_iir_coefficients(1e-3, 1/2f64.sqrt(), 2.)` from `dsp/src/lockin.rs`. We can reasonably move that into `dsp/src/iir_int.rs` and port it to `cossin()`. It doesn't neet do be fast.

jordens: https://github.com/quartiq/stabilizer/blob/c769bdbd4cb5759fff601b2a99e6822a1a31c421/src/bin/lockin.rs#L59

ryan-summers:
DS4_QuickPrint3

DAC0 = internally generated sinusoid, looped back and sampled by ADC1
DAC1 = phase output signal after going through IIR filter with gain = 1000, corner frequency = 10KHz, and then output scaled by 1e12.

ryan-summers:
(These captures are copied from the Quartiq matrix chat)

image

In the above capture, CH1 is the DAC0/ADC1 internally generated sinusoid. Channel 3 is the filtered phase output - full-scale DAC is equivalent to +/- PI phase. MATH is configured to be the phase measurement in radians (given gain = 2 and 10.24V = FS, 5.12V/rad is used). As shown over longer time periods, the phase is less stable.

image

Above is the same signal captured over a shorter time period

Note that at these very low amplitudes (order of ~7-8mV on phase output) there's significant measurement

src/bin/lockin-internal-demo.rs
```diff
@@ -0,0 +1,160 @@
...
+    fn process(c: process::Context) {
+        let adc_samples = c.resources.adc1.acquire_buffer();
+        let dac_samples = [
+            c.resources.dacs.0.acquire_buffer(),
+            c.resources.dacs.1.acquire_buffer(),
+        ];
+
+        // DAC0 always generates a fixed sinusoidal output.
+        for (i, value) in DAC_SEQUENCE.iter().enumerate() {
+            // Full-scale gives a +/- 12V amplitude waveform. Scale it down to give +/- 100mV.
```
jordens, line 77: 10.24 V should be full scale on both ADC and DAC. Let's say `0.9*i16::MAX` to not clip anything. But why `1/120`?

src/bin/lockin-internal-demo.rs
```diff
@@ -0,0 +1,160 @@
+#![deny(warnings)]
```
jordens, line 1: Let's call it `lockin-internal`.

jordens (changes requested):

Issue #149, opened 2020-10-13
MQTT settings/status/telemetry framework


jordens:
To make MQTT easy to deploy on Stabilizer (and Thermostat, Thermostat_EEM, Driver, Booster, Humpback etc.) we need to reduce it's footprint in the code by developing some kind of framework.

### Status quo

For Stabilizer we have currently to route the json-over-TCP requests:
https://github.com/quartiq/stabilizer/blob/66c917b576ece8499769847e7a3eacbdd91a955b/src/main.rs#L109-L158

For Booster there is a sizeable amount of request routing, deserialization, serialization code. Adding a setting or telemetry quantity to the stack requires multiple adjustment in several different places.

### Ideas

(ripped from #147): Stabilizer is somewhat similar to redpid (https://github.com/quartiq/redpid). Redpid has a pretty convenient solution to the settings/status problem. For migen/misoc we have a very generic way of configuration-status-registers. They are instantiated in code with a single line (e.g. https://github.com/quartiq/redpid/blob/master/gateware/limit.py#L55) and then end up on a bus with mapping generated automatically (https://github.com/quartiq/redpid/blob/master/test/csrmap.py#L109).
This is very different from Stabilizer in terms of code (HDL versus machine) and transport/protocol (HDL bus CSRs over other buses and serialization versus MQTT) but the user-facing concept is the same: lots of knobs and displays on a big switchboard. The knobs being connected to settings and the displays being connected to telemetry.
Maybe for Stabilizer we want something similar (a named map of settings and status values).
The settings aspect should be double-buffered/paged/ping-pong to allow atomic updates of multiple settings synchronously without having to transfer the entire settings space.
And we need to think a bit about the status telemetry: There might be a case for configurable rates and suppression of some telemetry updates.

### Context

* IoT: https://github.com/quartiq/stabilizer/issues/54#issuecomment-607478814

irwineffect:
Ryan and I discussed a few ideas yesterday, below are my notes:

# Introduction
We are thinking about having a primary setting struct which would be a struct of structs of
fields that operates as a staging area for new settings. The user will define
this structure. Settings change requests from MQTT will operate on this struct,
but the settings will not "go live" until the user initiates a "commit"
request. At this point, the settings framework will trigger an "commit
settings" event. This "update settings" routine would be written by the user, it
will be given the settings structure, the user would use this structure to
populate the individual configurable resources that each task needs.

# The Global Setting Structure
[...]

PR #242, opened 2021-01-27
MQTT settings and telemetry

ryan-summers:
This PR addresses #149 by exposing a simple MQTT-based settings interface

**TODO**:
- [x] Resolve mutable ownership issues in the `MqttInterface`
- [x] Expand settings interface to `lockin.rs`
- [ ] Add all settings (not yet supported by `miniconf`)
- [ ] Finalize and publish the `miniconf` package
- [x] Move `MqttInterface`-like behaviors over to `miniconf`
- [x] Move smoltcp-nal to a separate crate
- [ ] Test on hardware
- [ ] Implement stabilizer telemetry


jordens:
There will be items where committing them should not be unconditional. E.g. the state of an IIR filter: it makes sense to set them (to do a bumpless transfer you set the coefficients to "hold" the output, then manually change the output state, then set the coefficients to start steering the output) but it makes no sense to set it on each commit() even if it hasn't been dirtied. One option would be to encode that in the settings, i.e. have a setting that explicitly marks other settings as dirty and to be committed.

ryan-summers:
Closing in favor of https://github.com/quartiq/stabilizer/pull/261

PR #254, opened 2021-02-01
readme: fix source path and binary names

sergachev:


jordens:
Needs a general overhaul but this is certainly a good quick fix.

PR #255, opened 2021-02-02
Rj/timestamp tweaks

jordens:


src/hardware/configuration.rs
```diff
@@ -211,15 +211,15 @@ pub fn setup(
...
         timer5.set_tick_freq(design_parameters::TIMER_FREQUENCY);
 
         // The timestamp timer must run at exactly a multiple of the sample timer based on the
         // batch size. To accomodate this, we manually set the prescaler identical to the sample
         // timer, but use a period that is longer.
         let mut timer = timers::TimestampTimer::new(timer5);
 
-        let period = digital_input_stamper::calculate_timestamp_timer_period();
-        timer.set_period_ticks(period);
+        timer.set_period_ticks(u32::MAX);
```
ryan-summers, line 222: I was under the impression we wanted the timestamp timer to overflow at a power-of-two at-least greater than the sampling timer period: https://github.com/quartiq/stabilizer/issues/170


> Timestamping timer overflow/reload period is j*n*t. Locked through the CPU PLL to the CPU XO. Not yet/necessarily deterministic (across reboots, resets, timer reconfigurations) phase w.r.t. ADC/DAC sampling timer. j >= 1, a power of two, as large as possible within the timer counter size.


Given my understanding of the timestamp timer now, I actually don't think we really care, since we internally manage roll-overs for phase calculation and setting the period to u32::MAX should support all of our use cases without issue. That being said, would you mind updating the comment on 217-220 appropriately?

src/hardware/configuration.rs
```diff
@@ -211,15 +211,15 @@ pub fn setup(
         // Configure the timer to count at the designed tick rate. We will manually set the
         // period below.
         timer5.pause();
+        timer5.reset_counter();
```
ryan-summers, line 214: Resetting here is unnecessary - the timer is reset immediately before enabling the counter in `lockin-external.rs::init()` during `Timer5::start()`

src/bin/lockin-external.rs
```diff
@@ -117,8 +117,14 @@ const APP: () = {
         let iir_state = c.resources.iir_state;
         let lockin = c.resources.lockin;
 
+        let t = c
+            .resources
+            .timestamper
+            .latest_timestamp()
+            .unwrap_or_else(|t| t) // Ignore timer capture overflows.
```
ryan-summers, line 124: Are we currently hitting cases where the capture is overflowing? I feel a tad uncomfortable with silently ignoring these faults, since they _should_ be indicating that something isn't working how it should.

src/bin/lockin-external.rs
```diff
@@ -117,8 +117,14 @@ const APP: () = {
         let iir_state = c.resources.iir_state;
         let lockin = c.resources.lockin;
 
+        let t = c
```
ryan-summers, line 120: s/t/timestamp

(rename t -> timestamp)

ryan-summers (changes requested):

src/hardware/configuration.rs
jordens: #170 is make obsolete by the implementation and documentation of the PLLs. We want and strictly need a i32 overflowing timer.

src/bin/lockin-external.rs
jordens: They are hit easily in practice when playing with the device by either glitchy inputs or fast input frequency. Electrical inputs should never panic/crash the firmware. In all cases, especially this one, the DSP/code will recover once the source of the invalid input is removed.

src/hardware/configuration.rs
jordens: What about Timer2 and Timer3 (e.g. https://github.com/quartiq/stabilizer/blob/e1c87c149f4ed3b76eba343703d5d3879cafb3ae/src/hardware/configuration.rs#L157). Aren't they reset during `start()`ed as well?

There is an underlying issue here: we need to be aware that the timestamping and the sampling timer are not strictly synchronized in their initial counter values. This will lead to a non-deterministic phase shift.

ryan-summers: Synchronization is not at issue here from my understanding - if we want them to be in sync, that can be done (we do it for the shadow_sampling_timer), but we should do it in a separate PR - this was a design decision that we explicitly made early on.

`timer3`'s reset is necessary because it only receives a "TRIGGER" synchronization input from TIM2 (which starts the counter but does not reset it). However, you are correct that TIM2's reset here is superfluous. In any case, this is a superficial reset that doesn't have an effect on functional firmware, so it's acceptable to leave it in or remove it (user preference).

src/bin/lockin-external.rs
ryan-summers: Would you mind adding a TODO here to increment some counter value if we hit a fault? That would at least remind us to report these failures via telemetry so we can diagnose a faulty connection or input in the future.

jordens: There is no TODO needed here. A timer overflow has the similar consequences and the same severity as a voltage clipping or some signal hitting a limit.

src/hardware/configuration.rs
jordens: It's good to add that design decision to the code. I have done that now.
And I've removed the spurious TIM2 and TIM5 resets.

src/bin/lockin-external.rs
ryan-summers: I'm simply trying to say it would be a useful metric to at least inform the user that timestamps are overflowing when we don't expect them to. Otherwise, it may be difficult to pinpoint faults

jordens: Signalling that the reference clock is fast or slow or the input voltage is high or low is nice. Those are parts of the bigger topic of monitoring. There are a good dozen equally important other places. They can be readily identified. I wouldn't sprinkle TODOs unless we have a clear plan or desire to signal but trouble finding them.

PR #256, opened 2021-02-02
Adding support for cargo-embed

ryan-summers:
This PR adds support for flashing using cargo-embed to simplify the flashing process. This removes the need for `openocd` and external tooling when flashing a complete binary.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 /target
 openocd.gdb
+Embed.toml
--- a/Embed.toml
+++ b/Embed.toml
@@ -1,0 +1,3 @@
+[default.general]
+chip = "STM32H743ZITx"
+connect_under_reset = true
--- a/README.md
+++ b/README.md
@@ -40,6 +40,11 @@ See https://github.com/sinara-hw/Stabilizer
 * `cargo build --release`
 * Do not try the debug (default) mode. It is guaranteed to panic.
 
+### Using Cargo-embed
+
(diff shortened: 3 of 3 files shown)


.gitignore
```diff
@@ -1,2 +1,3 @@
 /target
 openocd.gdb
+Embed.toml
```
jordens, line 3:
```suggestion
```

Neither should be ignored. Both should be tracked. If people need custom scripts then we need to find another way.

jordens (changes requested):
The problem with documenting all these (4?) different ways of flashing the device is maintaining that documentation and its divergence. But yeah.

ryan-summers:
> The problem with documenting all these (4?) different ways of flashing the device is maintaining that documentation and its divergence. But yeah.

Agreed - I would ideally like to remove openocd entirely from our support, given that the current workflow often requires a new user to compile openocd from source.

If we don't want to support embed, I can set up a local Embed.toml file - I just figured it would be easier to point users to this approach since it avoids the openocd headache

jordens:
Embed is fine. I'd rather dump the objdump, DFU, and ST-link mass storage sections. But I'd like to keep openocd as long as it's the only really usable gdb interface.

jordens (approved):

ryan-summers:
Just an FYI - cargo embed _does_ have GDB support, but we would need to refactor our cargo-runner to use it.

jordens:
I know. Last I checked it was mostly unusable and generally an order of magnitude slower than openocd.

PR #257, opened 2021-02-03
bump HAL and smoltcp

jordens:
This bumps the HAL (still dma branch) and smoltcp (to 0.7).
Removes a few lifetime parameters due to the smoltcp changes. Adds type parameters to `dma::Transfer`.

* [x] Test on hardware

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -966,6 +966,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "smoltcp"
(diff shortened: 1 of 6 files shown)


ryan-summers:
I have verified this works as expected with a 1KHz sinusoid input on ADC0 using the `dual-iir` binary.

ryan-summers (approved):

PR #258, opened 2021-02-03
Updating input capture for timers

ryan-summers:
This PR fixes the register configuration for input capture configurations and updates the lockin-external application to enable the timestamp capture mechanism.

**Testing**

Debug code was inserted into the DSP routine such that a log was generated after each 1000 samples. A 1KHz signal was plugged in to the DI0 input and it was verified that after approximately 1 second, a break-point tripped at the 1000 sample log message.

A capture of DAC0/DAC1 with DI0 and ADC0 driven with a 1KHz signal is shown below:
DS4_QuickPrint7

### DIFF
--- a/src/bin/lockin-external.rs
+++ b/src/bin/lockin-external.rs
@@ -71,6 +71,9 @@ const APP: () = {
         // Start sampling ADCs.
         stabilizer.adc_dac_timer.start();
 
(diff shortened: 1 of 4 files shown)


jordens:
Nice!

jordens (approved):

Issue #70, opened 2019-12-03
integrator hold


jordens:
During `process()`, if a digital input is high, do not update the IIRs.

* Strictly speaking the correct implementation is to update the `y` (output) delay line with the unchanged previous output, but keep updating the `x` delay line with new input. This seems to be semantically best implemented by updating with a "unity-gain-feedback-iir" instead of the usual "active" one. As a first approximation to this, just not calling `iir.update(..)` is fine.
* Would be nice to have the pin functionality configurable or at least not conflicting. I.e. not have code conflict between the use of DIN0 as ref input for timestamping and the use as a GPIO (to be read explicitly). I.e. be able to use either functionality for DIN0 in different binaries. Conveniently, the DI0/1 signals each are available on two pins. This is easy then.
* This integrator hold flag should be and-ed with a flag `integrator_hold_chA_dinB_en` and or-ed with a flag `integrator_hold_chA` from the network-accessible settings.

Note: DINs should have pulldowns. https://github.com/sinara-hw/Stabilizer/issues/91

PR #261, opened 2021-02-03
MQTT settings and telemetry

ryan-summers:
This PR replaces https://github.com/quartiq/stabilizer/pull/242

This PR addresses #149 by exposing a simple MQTT-based settings interface

**TODO**:
- [x] Resolve mutable ownership issues in the `MqttInterface`
- [x] Expand settings interface to `lockin.rs`
- [x] Add all settings (not yet supported by `miniconf`)
- [x] ~~Finalize and publish the `miniconf` package~~ Deferred to the future.
- [x] Move `MqttInterface`-like behaviors over to `miniconf`
- [x] Move smoltcp-nal to a separate crate
- [x] Test on hardware
- [x] ~~Implement stabilizer telemetry~~: Deferred to future


jordens:
If there is an ergonomic advantage over using raw mqtt CLI utilities, let's also do a minimal CLI example client implementation for the IIR parameters in `dual-iir`.

Cargo.toml
```diff
@@ -46,13 +45,16 @@ paste = "1"
...
+
 [dependencies.mcp23017]
 git = "https://github.com/mrd0ll4r/mcp23017.git"
 
-[dependencies.smoltcp]
-version = "0.7"
-features = ["ethernet", "proto-ipv4", "socket-tcp", "proto-ipv6"]
-default-features = false
+[dependencies.smoltcp-nal]
+git = "https://github.com/vertigo-designs/smoltcp-nal.git"
```
ryan-summers, line 56: Update git dependencies to the quartiq repositories - also, publish `smoltcp-nal`

dsp/Cargo.toml
```diff
@@ -7,6 +7,7 @@ edition = "2018"
 [dependencies]
 libm = "0.2.1"
 serde = { version = "1.0", features = ["derive"], default-features = false }
+serde-json-core = "0.1"
```
ryan-summers, line 10: I don't think we should need this. Might be a bug in miniconf that's requiring this

src/bin/dual-iir.rs
```diff
@@ -106,159 +140,47 @@ const APP: () = {
...
         loop {
-            let tick = Instant::now() > next_ms;
-
-            if tick {
-                next_ms += 400_000.cycles();
-                time += 1;
-            }
-
+            let sleep = c.resources.mqtt_interface.lock(|interface| {
+                !interface.network_stack().poll(clock.current_ms())
```
ryan-summers, line 149: This sleep seems to be behaving oddly - the network interface doesn't seem like it's waking up at the correct points and this needs some investigation.

src/bin/dual-iir.rs
```diff
@@ -106,159 +140,47 @@ const APP: () = {
...
 
+    #[task(priority = 1, resources=[mqtt_interface, afes, iir_ch, iir_state])]
+    fn settings_update(mut c: settings_update::Context) {
+        let settings = &c.resources.mqtt_interface.settings;
+
+        // Update the IIR channels.
+        c.resources.iir_ch.lock(|iir| *iir = settings.iir_ch);
+
+        // Update the IIR states.
+        c.resources.iir_state.lock(|iir| *iir = settings.iir_state);
```
ryan-summers, line 177: @jordens You had mentioned that we don't want to update the IIR channels all the time - would we always want to update the `iir_state` and `iir_channel` simultaneously? Also, would we want to update CH0/1 independently?

src/bin/lockin-external.rs
```diff
@@ -4,20 +4,43 @@
...
 };
 
+#[derive(Deserialize, StringSet)]
+pub struct Settings {
+    data: u32,
+}
+
+impl Settings {
+    pub fn new() -> Self {
+        Self { data: 5 }
```
ryan-summers, line 28: Replace with settings similar to `dual-iir`

src/hardware/cycle_counter.rs
```diff
@@ -0,0 +1,40 @@
+use rtic::cyccnt::{Duration, Instant, U32Ext};
```
ryan-summers, line 1: Comments throughout this file

src/bin/dual-iir.rs
ryan-summers: Hmm, scratch that, appears to be working fine now

src/bin/dual-iir.rs
jordens: Yes. Only update the `iir_state` if that is explicitly flagged.

Also, this is somewhat related to https://github.com/quartiq/stabilizer/issues/70
Let me know if that description isn't clear.

dsp/src/iir.rs
```diff
@@ -11,8 +12,7 @@ use core::f32;
 /// To represent the IIR coefficients, this contains the feed-forward
 /// coefficients (b0, b1, b2) followd by the negated feed-back coefficients
 /// (-a1, -a2), all five normalized such that a0 = 1.
-#[derive(Copy, Clone, Default, Deserialize, Serialize)]
-pub struct Vec5(pub [f32; 5]);
+pub type Vec5 = [f32; 5];
```
jordens, line 15: The reason I wanted a newtype is to be able to implement things on it, like in `irr_int`.

ryan-summers: This type isn't supported by serde-json-core and it was why we were getting panics previously. We can't have anonymous fields in structs, or we will panic when deserializing them. If we want a custom type, we have to implement it with named fields.

That being said, perhaps we can redefine `Vec5` to
```rust
pub struct Vec5 {
    data: [f32; 5],
}
```

to get the best of both worlds?

jordens: Yes. That's fine!

jordens: In fact, maybe a trait for the type alias is better.

dsp/Cargo.toml
```diff
@@ -9,6 +9,10 @@ libm = "0.2.1"
 serde = { version = "1.0", features = ["derive"], default-features = false }
 generic-array = "0.14"
 
+[dependencies.miniconf]
```
jordens, line 12: I think this is better done with a `patch` at the root. `dsp` shouldn't need `miniconf`, right?

ryan-summers: dsp needs `miniconf::StringSet` so that it can use it in the `#[derive(StringSet)]` for the DSP types, like IIR. If we don't want those to be directly settable, we can remove the dependency

jordens: Ah ok.
I was wondering when it would make sense to bring miniconf down into dsp. And this is just that case. Otherwise, we'd have to duplicate the IIR struct (structurally) in the settings.

jordens: But I still think the this should be done with a `patch` at the root Cargo.toml, no?

dsp/src/iir.rs
jordens: #274

dsp/Cargo.toml
ryan-summers: `patch` is generally used to override an existing dependency with a different version. In this case, `miniconf` isn't available at all in our dependency tree, so I'm not sure what we could `patch`. Unless you're referring to something else?

I dislike having to doubly specify the `miniconf` dependency, but I think that's a result of how we've structed `dsp` as a separate crate instead of a module (like `hardware`) within the `stabilizer` crate.

ryan-summers: Unless you're specifying the specify `branch` specifier to only appear at the root and use patch to push it down our dependency graph?

I plan on waiting until that branch is merged to mainline and we release miniconf before merging this, so I've been leaving it as-is for now.

jordens:
```
[dependencies]
miniconf = "0.1"
```

in both Cargo.toml and then

```
[patch.crates-io]
miniconf = { git = "https://github.com/quartiq/miniconf.git", branch = "rs/issue-21/terminal-array-elements"}
```

only at the root works afaict.

src/bin/lockin-external.rs
```diff
@@ -4,31 +4,106 @@
...
+
+    // Harmonic index of the LO: -1 to _de_modulate the fundamental (complex conjugate)
+    harmonic: i32,
+
+    // Demodulation LO phase offset
+    phase_offset: i32,
+
+    // Log2 lowpass time constant
+    time_constant: u8,
+}
```
jordens, line 38: We don't necessarily need to bother with the other binaries in this PR yet. Having just `dual-iir` for now would be ok.

ryan-summers: I was thinking the same thing - otherwise, things start to get a big big. Let me yank all of the lockin-external changes to a different branch so we can keep things tidy.

src/bin/dual-iir.rs
jordens: The more I think about it, for now it's ok to never update `iir_state`. Let's keep it out of settings.

dsp/Cargo.toml
jordens: Yeah. It's a nit. Just a few fewer lines.

jordens (approved):

ryan-summers:
I have bench tested setting various values of `dual-iir`, such as the IIR y_offset parameter and the AFE gain. Everything appears to be in working order for merging.

ryan-summers:
bors merge

PR #263, opened 2021-02-04
deglitch timer input, miscellaneous changes

jordens:


src/hardware/digital_input_stamper.rs
```diff
@@ -47,6 +47,10 @@ impl InputStamper {
         let input_capture =
             timer_channel.into_input_capture(timers::tim5::CaptureSource4::TI4);
 
+        // FIXME: hack in de-glitching filter
+        let regs = unsafe { &*hal::stm32::TIM5::ptr() };
+        regs.ccmr2_input().modify(|_, w| w.ic4f().bits(0b0011));
+
```
jordens, line 53: @ryan-summers I didn't see an easy way to put this into `timers.rs` since both the PAC coverage is incomplete and the different timers somewhat incompatible as well.

ryan-summers: We can do something similar to what I did for the CaptureTriggerX input (re-export the register locally). I can implement the change later if you'd like - adding support for the input filtering shouldn't be too onerous.

jordens: That would be better. Still there are the undocumented fields in the PAC making all this less than optimal.

ryan-summers: PR for the fix (PR into this branch): https://github.com/quartiq/stabilizer/pull/265

src/bin/lockin-external.rs
```diff
@@ -142,187 +121,26 @@ const APP: () = {
             .last()
             .unwrap();
 
-        // convert i/q to power/phase,
-        let power_phase = true; // TODO: expose
-
-        let mut output = if power_phase {
+        let conf = "frequency_discriminator";
```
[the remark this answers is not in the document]
jordens: I agree. In this case it was a bit more code. And since the enum would be moved into the settings structure and exposed via miniconf, it seems ok to improve when doing that refactor.

PR #265, opened 2021-02-09
Adding support for input capture prefilter configuration

ryan-summers:
This adds support for configuring the input capture prefilter for an input capture channel.

I looked through timers TIM1-TIM5, and TIM8, and all of them have the same filter configurations from what I can tell. There's no unique enums for the prefilter configuration unfortunately.

TODO:
- [ ] Test on hardware for proper configuration


jordens:
Nice. Thanks. I had the impressions that the channel to register-bit routing was different. But maybe that was already covered with the macro name composition.

PR #277, opened 2021-02-19
Adding general MQTT utility for stabilizer

ryan-summers:
This PR adds a general utility script to easily configure stabilizer settings via MQTT.

The script is agnostic to the underlying exposed settings, but supports parsing arrays, floats, and integers. Any other data will be assumed to be a string.


stabilizer.py
```diff
@@ -0,0 +1,151 @@
...
+Description: Provides an API for controlling Stabilizer over Miniconf (MQTT).
+"""
+import argparse
+import asyncio
+import json
+
+from gmqtt  import Client as MqttClient
+
+
+def parse_value(value):
```
ryan-summers, line 14: I would like to ideally move `parse_value` and `MiniconfApi` over to the `miniconf` repository in the future and release it as a package via pip, but it's existing here as an initial prototype. Let me know what you think @jordens

stabilizer.py
```diff
@@ -0,0 +1,151 @@
...
+
+    # Establish a communication interface with stabilizer.
+    interface = await MiniconfApi.create(args.stabilizer, args.broker)
+
+    request = None
+
+    # In the exceptional case that this is a terminal value, there is no key available and only a
+    # single value.
+    if len(args.values) == 1 and '=' not in args.values[0]:
+        request = parse_value(args.values[0])
```
ryan-summers, line 117: Need to dump this into JSON as well

stabilizer.py
```diff
@@ -0,0 +1,150 @@
...
+        """
+        self.command_complete.clear()
+        self.client.publish(topic, payload=message, qos=0, retain=False,
+                            response_topic=self.response_topic)
+        await self.command_complete.wait()
+
+        response = self.response
+        self.response = None
+
+        return response
```
jordens, line 93: These should be a map from topic to `Future`s that you `set_result()` in `_handle_response`.
There can be multiple in flight with `asyncio`.

stabilizer.py
jordens: Just `json.loads()`?

stabilizer.py
```diff
@@ -0,0 +1,150 @@
...
+        request = parse_value(args.values[0])
+    else:
+        # Convert all of the values into a key-value list.
+        request = dict()
+        for pair in args.values:
+            key, value = pair.split('=')
+            request[str(key)] = parse_value(value)
+
+    response = await interface.set_setting(args.setting, json.dumps(request))
+    print(f'+ {response}')
```
jordens, line 126:
```python
logger = logging.getLogger(__name__)

...

logger.debug(...)
```

And `.add_argument("-v", ...)`.

stabilizer.py
jordens: But Yes. Move `Miniconf` to `miniconf` when convenient and ~stable.

ryan-summers: Ah, json-loads is a much simpler approach. I don't know why that slipped my mind. Good catch

stabilizer.py
ryan-summers: Current design does it all sequentially, since the CLI only supports a single setting configuration at once - the key=value pairs are parsed into a single JSON object that is sent to a single settings message, unless I'm misunderstanding what you're asking?

jordens: Yes. No problem if the only user is this CLI.
But other users of the API could fire multiple requests before the responses come in.

ryan-summers: In the interest of simplicity, I would say let's leave it for now, and potentially expand if we have that use case later on

jordens: Ok. But then:

* Use a `Future` instead of the `Event`+`self.response`.
* Add documentation stating that the user needs to ensure only one request in flight.

ryan-summers: There's actually no data the the API can use to differentiate one response from another, so it's not possible to have a mapping. If we wanted to support multiple in-flight messages, we would have to do one of:
* Utilize fire-and-forget configurations
* Modify `miniconf` to add a `Property` to the response object indicating the setting that the response is associated with.

jordens: You could have stabilizer reply on unique response topics or return a token. But restricting this to a single call in flight is OK for now. Just make sure it's documented.
That's one thing. Using `Future` instead of `Event` + `self.response` is the other comment, to make this use more typical `asyncio` patterns.

jordens (approved):

PR #291, opened 2021-03-02
Updating minconf + minimq dependencies

### LINKED ISSUE #288: MQTT setting commit on startup
When Stabilizer boots, the current settings should be read from the (retained) MQTT topics, committed to hardware (and arguably dumped back to MQTT for any topics that might have been newly introduced in that firmware version).

Currently, an explicit commit seems to be necessary. (?)

ryan-summers:
This PR updates stabilizer to use the latest versions of miniconf and minimq. This adds:
* atomic updates of register blocks
* Requirements to configure settings at the terminal (leaf) node
* Support for unspecified, auto-assigned client IDs

This PR fixes #288 by removing the concept of "committing" settings in favor of continuous update.

### DIFF
--- a/src/bin/dual-iir.rs
+++ b/src/bin/dual-iir.rs
@@ -8,7 +8,7 @@ use stabilizer::hardware;
 
 use miniconf::{
     embedded_nal::{IpAddr, Ipv4Addr},
-    minimq, MqttInterface, StringSet,
+    minimq, Miniconf, MqttInterface,
 };
 use serde::Deserialize;
 
@@ -23,7 +23,7 @@ const SCALE: f32 = i16::MAX as _;
 // The number of cascaded IIR biquads per channel. Select 1 or 2!
 const IIR_CASCADE_LENGTH: usize = 1;
 
-#[derive(Debug, Deserialize, StringSet)]
+#[derive(Debug, Deserialize, Miniconf)]
 pub struct Settings {
     afe: [AfeGain; 2],
     iir_ch: [[iir::IIR; IIR_CASCADE_LENGTH]; 2],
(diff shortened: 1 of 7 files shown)


src/bin/dual-iir.rs
```diff
@@ -143,23 +139,15 @@ const APP: () = {
...
-                miniconf::Action::Continue => {
-                    if sleep {
-                        cortex_m::asm::wfi();
-                    }
-                }
-                miniconf::Action::CommitSettings => {
-                    c.spawn.settings_update().unwrap()
-                }
+                c.spawn.settings_update().unwrap()
             }
```
jordens, line 151:
```suggestion
            } else {
                cortex_m::asm::wfi();
            }
```

No?

ryan-summers: Was just playing around with the semantics here. I think it should just be `else if sleep` - otherwise, we may sleep when other (non MQTT) network traffic is received, which I don't think would be desirable for the new streaming app?

src/bin/dual-iir.rs
```diff
@@ -140,7 +140,7 @@ const APP: () = {
 
         loop {
             let sleep = c.resources.mqtt_interface.lock(|interface| {
-                interface.network_stack().poll(clock.current_ms())
+                !interface.network_stack().poll(clock.current_ms())
```
jordens, line 143: `poll() -> true` if the stack has seen activity.

jordens (changes requested):

src/bin/dual-iir.rs
jordens: Yes. I hadn't seen the subsequent commits yet.

src/bin/dual-iir.rs
ryan-summers: Correct, so this is saying `let sleep = !stack_was_updated`, which is semantically equivalent to `let do_not_sleep = stack_was_updated`

jordens: Ah, right. I got confused by the previous commit removing the negation.

PR #293, opened 2021-03-02
miniconf cli

jordens:
* retain messages
* use plain json for argument format
* rename feedback -> response to match code
* reorganize args
* tweak variable names to be more in line with miniconf and mqtt
* use `mqtt` as default broker address
* remove traces of stabilizer from script, rename to miniconf.py
* add usage example

dual-iir/lockin-external:

* update broker address (for HITL),
* change mqtt path to convention


miniconf.py
```diff
@@ -0,0 +1,117 @@
...
+    parser.add_argument('path', type=str,
+                        help='The setting path to configure')
+    parser.add_argument('value', type=str,
+                        help='The value of setting in JSON format')
+
+    args = parser.parse_args()
+
+    logging.basicConfig(
+            format='%(asctime)s [%(levelname)s] %(name)s: %(message)s',
+            level=logging.WARN - 10*args.verbose)
```
ryan-summers, line 104: Don't we want to by-default log the firmware response to the setting update?

src/bin/dual-iir.rs
```diff
@@ -62,13 +62,17 @@ const APP: () = {
 
         let mqtt_interface = {
             let mqtt_client = {
-                let broker = IpAddr::V4(Ipv4Addr::new(10, 34, 16, 1));
+                let broker = IpAddr::V4(Ipv4Addr::new(10, 34, 16, 10));
```
ryan-summers, line 65: Lets pull this default broker address out to a single config file so it stays consistent for all of the binaries. Maybe `design_parameters.rs`?

ryan-summers (changes requested):

src/bin/dual-iir.rs
jordens: Ack

miniconf.py
jordens: I made it `print()` that result since `gmqtt` is pretty talkative already at `INFO` and since this is a CLI utility wrapper where it seems fine to just print any result of the single tasks it's supposed to do.
But it would be better if we formalize the miniconf response to: (a) be machine readable JSON: `{code: 200, message: "OK"}` and (b) once we have readback actually post the value that's being used (which might differ from the request e.g. due to rounding a DDS FTW word).

ryan-summers (approved):

PR #298, opened 2021-03-04
Update stm32h7xx-hal

dnadlinger:
I haven't done any runtime tests using Pounder, but it seems
like the API change is just a consequence of making the QSPI
address width runtime-configurable.


jordens (approved):
Not tested but that was my perception of the changes as well.

ryan-summers (approved):
LGTM. Basic test on stabilizer shows that it still operates nominally.

PR #300, opened 2021-03-05
hardware: Enable DEMCR TRCENA on boot for cycle counter

### LINKED ISSUE #299: Irregular Ethernet stalls when not running in debugger
As mentioned on #quartiq, I'm seeing weird network hangs where Stabilizer stops responding on the network, which sometimes clear themselves after a few seconds.

Trying to track this down, I essentially used the MQTT client as a packet generator by making sure miniconf update errors don't panic the device (#286)
```diff
--- a/src/bin/dual-iir.rs
+++ b/src/bin/dual-iir.rs
@@ -149,7 +149,7 @@ const APP: () = {

             if c.resources
                 .mqtt_interface
-                .lock(|interface| interface.update().unwrap())
+                .lock(|interface| interface.update().unwrap_or(false))
             {
                 c.spawn.settings_update().unwrap()
             } else if sleep {
```
and changing the broker IP to a Linux box on the same subnet/physical switch that doesn't actually listen on the MQTT port. Other than this (and changing the IP address), the firmware is unmodified dual-iir master (65e9131bfb985e35585facabb27adeeae7b1da69).

Dumping the TCP traffic on the "fake" broker box, I do see a steady stream of SYN/RST pairs as expected, but there are periodic pauses on the order of ~1s, as evidenced by the vertical steps in this plot of packet arrival times:

image

I've spot-checked a few steps, and it looks like it's indeed Stabilizer not sending anything, rather than the Linux host not responding, or the capture buffer overflowing/…

There are no panics, and no obvious glitches in the IIR loop behaviour (DAC0 continues to show sine wave input applied to ADC0).

This feels quite similar to https://github.com/quartiq/stabilizer/issues/33.

dnadlinger:
Previously, the cycle counter would only work correctly when
running in the debugger (which would also enable tracing).

GitHub: Fixes #299.

### DIFF
--- a/src/hardware/configuration.rs
+++ b/src/hardware/configuration.rs
@@ -825,6 +825,14 @@ pub fn setup(
         None
     };
 
+    let cycle_counter = {
+        // Set TRCENA bit, which is required for DWT counter to tick. (This is also
+        // set automatically when running in the debugger, and only cleared on power
+        // reset, not on soft reset.)
+        core.DCB.enable_trace();
+        CycleCounter::new(core.DWT, ccdr.clocks.c_ck())
+    };
+
     let stabilizer = StabilizerDevices {
         afes,
         adcs,
@@ -833,7 +841,7 @@ pub fn setup(
         net: network_devices,
         adc_dac_timer: sampling_timer,
         timestamp_timer,
-        cycle_counter: CycleCounter::new(core.DWT, ccdr.clocks.c_ck()),
+        cycle_counter,
     };
 
     // info!("Version {} {}", build_info::PKG_VERSION, build_info::GIT_VERSION.unwrap());


src/hardware/configuration.rs
```diff
@@ -825,6 +825,14 @@ pub fn setup(
         None
     };
 
+    let cycle_counter = {
+        // Set TRCENA bit, which is required for DWT counter to tick. (This is also
+        // set automatically when running in the debugger, and only cleared on power
+        // reset, not on soft reset.)
+        core.DCB.enable_trace();
```
ryan-summers, line 832: This isn't an issue with other platforms because RTIC, when using the CycleCounter as a monotonic, automatically generates the `core.DCB.enable_trace()` call in the APP macro.

ryan-summers (approved):
This isn't an issue in `Booster` for example because RTIC internally generates the code to call `core.DCB.enable_trace()`.

PR #303, opened 2021-03-07
hardware: Wait for I2C EEPROM to come online after boot

### LINKED ISSUE #294: I2C EUI-48 readout unreliable
From time to time, the EUI-48 readout (MAC address) from the I2C EEPROM chip appears to be unreliable.

I've now observed this two times, on two different boards. Each time, the v1.1 Stabilizer in question was sitting happily in a corner somewhere with JTAG and power connected constantly, being reprogrammed as I was working on the firmware. The first time, the I2C transactions started failing altogether at some point. The second time, the address read back was 80:00:10:4d:03:8b consistently instead of the usual (presumably correct) 80:1f:12:5d:47:df. Both times, the behaviour was consistent across JTAG resets (`monitor reset halt`) and full reloads. Completely power-cycling the board (removing both 12 V in and the JTAG programmer) fixed the issue.

Any idea what could be going on here? This is quite possibly a hw issue; I'll try putting a scope on the bus next time.

dnadlinger:
GitHub: Fixes #294.

### DIFF
--- a/src/hardware/configuration.rs
+++ b/src/hardware/configuration.rs
@@ -486,13 +486,15 @@ pub fn setup(
             .set_speed(hal::gpio::Speed::VeryHigh);
     }
 
-    let mac_addr = match eeprom::read_eui48(&mut eeprom_i2c) {
-        Err(_) => {
-            info!("Could not read EEPROM, using default MAC address");
-            smoltcp::wire::EthernetAddress([0x10, 0xE2, 0xD5, 0x00, 0x03, 0x00])
-        }
-        Ok(raw_mac) => smoltcp::wire::EthernetAddress(raw_mac),
-    };
+    // The EEPROM only gets connected to I2C after the analog supplies are up, which
+    // takes a while (particularly on pre-v1.2 hardware). Wait until a read succeeds,
+    // and then read again to get non-corrupted data.
+    info!("Waiting for I2C EEPROM to become available...");
+    while eeprom::read_eui48(&mut eeprom_i2c).is_err() {}
+    let mac_addr = smoltcp::wire::EthernetAddress(
+        eeprom::read_eui48(&mut eeprom_i2c).unwrap(),
+    );
+    info!("MAC address: {}", mac_addr);
 
     let network_devices = {
         // Configure the ethernet controller


src/hardware/configuration.rs
```diff
@@ -486,13 +486,15 @@ pub fn setup(
...
-            info!("Could not read EEPROM, using default MAC address");
-            smoltcp::wire::EthernetAddress([0x10, 0xE2, 0xD5, 0x00, 0x03, 0x00])
-        }
-        Ok(raw_mac) => smoltcp::wire::EthernetAddress(raw_mac),
-    };
+    // The EEPROM only gets connected to I2C after the analog supplies are up, which
+    // takes a while (particularly on pre-v1.2 hardware). Wait until a read succeeds,
+    // and then read again to get non-corrupted data.
+    info!("Waiting for I2C EEPROM to become available...");
+    while eeprom::read_eui48(&mut eeprom_i2c).is_err() {}
```
ryan-summers, line 493: Reading in a loop like this may result in a hung application. Maybe a maximum number of attempts would be appropriate before panicking?

dnadlinger: Yes, having the red LEDs on the front panel after a while might be nicer when running without any debugger/logging. (With the debugger connected, the "Waiting" message would be enough indication.)

Any idea how to best determine the (large) number, though? I'm thinking booting up a board a few times and taking, say, twice or three times the maximum observed value for some safety margin, but I'm really not sure how big the spread would be in practice. (I guess the regulator set pin current would be quite tightly controlled.)

ryan-summers: The difficulty is that there is no upper bound on when this may succeed. It has been observed that P12V0A may _never_ come up properly on V1.1 hardware. Perhaps we should actually feature gate this to v1.1 stabilizers and have V1.2+ do a single read + unwrap. Given that - what do we _want_ to do if V1.1 has a stuck P12VA0 line? I imagine we should show red LEDs + loop?

dnadlinger: There are basically three modes: a) Comes up cleanly and quickly (v1.2+). b) Comes up cleanly, but takes a few seconds (v1.1). c) Never comes up (v1.1 due to that bug).

I don't think there is a case where v1.1 comes up only after, say, a minute.

ryan-summers: Yeah, I would then recommend that we do the following:
1. For V1.2 hardware, read the EUI48 once and unwrap the result. It should never fail.
2. For V1.1, read once per 500ms for ~5 seconds. After that, panic abort and illuminate LEDs.

Since we don't have support for run-time detection of the hardware yet, feel free to omit (1)

jordens: Agreed.

ryan-summers:
These fixes were implemented in https://github.com/quartiq/stabilizer/pull/336, so I'm closing this PR. Thanks for getting the conversation stated on this one!

PR #321, opened 2021-03-31
update cargosha256

HarryMakes:
This PR is to update `cargoSha256` for nixbld.m-labs.hk as of e92c2bd76c6f1a4257599f1ef371f505cec442fa.

I would like to ask for your comments about whether or not I should make this type of PR whenever I am aware of changes to crate dependencies. Unless there is objection, I do plan to keep this Git branch for updating information required by Nix for building the binaries, even if I am to submit a PR every time. Thanks!

### DIFF
--- a/cargosha256-dual-iir.nix
+++ b/cargosha256-dual-iir.nix
@@ -1,1 +1,1 @@
-"138kpxzxs73zhmd4xi5kw3fddb05gac4mpngizm01831n1ycyhl0"
+"06qsl59bljr637xcrplbij7ma8l7waryi4lkbd4fxjac0gqpn55s"


jordens:
Please comment and correct me if my understanding here is wrong, I'm not a nix expert, I just see these checksum updates coming in. And I'm not dogmatically or personally against nix in any way.

In my perspective this is severely flawed.

* It merely duplicates the functionality in cargo that ensures that dependency downloads are intact. Note that cargo already is meant to do that.
* It doesn't protect in any way against cargo or rustc being non-deterministic during build. They are black boxes. Black boxes can easily be non-deterministic. It only protects against cargo's download implementation being broken.
* You don't review the dependencies, you just trust and copy the hash you see on match failure. This approach to assume checksum failures are precisely **not** what they are intended to detect (non-deterministic downloads) but generally something benign (changing dependencies) blows a hole into the entire concept.
* If someone really wanted this despite the flaws above, TOFU could be implemented elsewhere, and automatically (i.e. when `Cargo.lock` changes). Implementing it elsewhere would also remove the noise here.
* It ignores most (2/3) of the binaries produced.
* It seems to require quite some maintenance on your side.
* It doesn't work most of the time because the dependencies change frequently. Builds are broken.
* I don't see the rationale for duplicating the build infrastructure here in nix. It doesn't run any tests, doesn't run benchmarks, doesn't check style, formatting, doesn't do HITL, isn't integrated into the development workflow.
* The only practical purpose the nix build seems to serve is to sometimes produce a `dual-iir` binary with patched default ip address and broker for your network. The former doesn't exist anymore and the latter will be removed once DNS is there.
* The decision to place the checksum here seems to be merely one of convenience as it may extrapolate for several successive commits. The checksum could just as well be placed in `nix-scripts`, no? It would equally work for multiple successive commits and support reproducing old builds (old stabilizer code with old nix-scripts) in the same way. Also note that `nix-scripts` needs to be changed regularly (whenever the code to be patched changes or when then binaries produced change).

In short, if I didn't misunderstand this completely, there is zero tangible gain and lots of noise.
I would suggest to remove this checksum file and implement TOFU or move it to `nix-scripts`. And if there are no other significant practical purposes then the one above, it seems adequate to stop continuous nix builds of this firmware altogether.

PR #322, opened 2021-03-31
rj/itcm

### LINKED ISSUE #315: Run `process()` task from ITCM
* To get most out of the faster (than flash) ITCM, the 64 bit dual issue pipeline, and the cortex-m7 instruction schedule coming with rust 1.52
* needs a custom `pre_init` (see https://github.com/rust-embedded/cortex-m-rt/pull/100 https://github.com/imxrt-rs/imxrt-rt/blob/6e8ec474abf88d769222df920f47ab057eda92d1/src/reset.c#L43-L60 https://github.com/mciantyre/teensy4-rs/blob/master/bin/start.s)
* The inlining in `release` mode looks good enough to not require to much proliferation of `link_section` attributes beyond the process task itself.

jordens:
* close #315
* would profit from cortex-m-rtic 0.6 elevating the attributes to the actual ISR thus removing the veneer


jordens:
bors try

jordens:
bors try

jordens:
bors try

src/startup.S
```diff
@@ -0,0 +1,44 @@
...
+  # Set DTCMCR.EN
+  orr r2, r2, r0
+  str r2, [r1]
+  dsb
+  isb
+
+  # Analogous to cortex-m-rt Reset code for .data copying.
+  # Initialise .itcm code. `__sitcm`, `__siitcm`, and `__eitcm` come from the
+  # linker script. Copy from r2 into r0 until r0 reaches r1.
+  ldr r0,=__sitcm
```
ryan-summers, line 29: General comment, but if we can guarantee that no ITCM code executes before `hardware::setup()`, why don't we do this initialization in rust-land? Then we can avoid the whole startup code / assembly section entirely. It may simplify things?

ryan-summers: Better yet, we could use a [`cortex-m-rt::pre_init` hook](https://docs.rs/cortex-m-rt/0.6.13/cortex_m_rt/attr.pre_init.html)

ryan-summers: I think that simplifies the build, since we don't need cc or any of the toolchain available

jordens: This is the `pre_init` hook.
Doing it in rust isn't completely trivial. I'd rather get a working solution first and then later optimize it.
https://github.com/rust-embedded/cortex-m-rt/issues/300 and https://github.com/rust-embedded/cortex-m-rt/issues/301

ryan-summers: Very valid point - that pointer provenance issue is super annoying.

jordens: Did it anyway. Seems to work.

ryan-summers: The problem is that it's _technically_ undefined behavior, so it _might_ break in the future - rust doesn't guarantee anything. It worked for years on cortex-m-rt, but then they noticed it and removed it. I don't know if it actually had a problem in that case, but a user was trying to reproduce what they had done and couldn't get it to work.

jordens: I don't think it is in this case. c-m-rt used `ptr` because it has no stack at that point. The problem is in using a `ptr::offset`.
Here I'm creating and using slices the same way you'd do for FFI. That also works "out of thin air".

ryan-summers: I think we want to do what they did in https://github.com/rust-embedded/r0/pull/25 to get around this.

jordens: There are already two.

jordens: And `r0` isn't used anymore.

src/hardware/configuration.rs
```diff
@@ -105,6 +105,32 @@ pub struct PounderDevices {
...
+    extern "C" {
+        static mut __sitcm: u32;
+        static mut __eitcm: u32;
+        static mut __siitcm: u32;
+    }
+    use core::{ptr, slice, sync::atomic};
+
+    // ITCM is enabled on reset on our CPU but might not be on others.
+    // Keep for completeness.
+    const ITCMCR: *mut u32 = 0xE000_EF90usize as _;
```
ryan-summers, line 119: Ideally, these would be added to `cortex-m-rt::peripheral::SCB`. It looks like they don't exist for now though, so this is fine as-is. Maybe make it `0xE00_EF90_usize`?

src/hardware/configuration.rs
```diff
@@ -105,6 +105,32 @@ pub struct PounderDevices {
 /// Static storage for the ethernet DMA descriptor ring.
 static mut DES_RING: ethernet::DesRing = ethernet::DesRing::new();
 
+/// Setup ITCM and load its code from flash
+unsafe fn setup_itcm() {
```
ryan-summers, line 109: Can we just tag this as `#[cortex_m_rt::pre_init]` and avoid having to manually call it during setup? Is there a reason we're doing it later on after hardware initialization (e.g. does this require RCC config, etc)? If so, we should document.

src/hardware/configuration.rs
```diff
@@ -105,6 +105,32 @@ pub struct PounderDevices {
...
+    ptr::write_volatile(ITCMCR, ptr::read_volatile(ITCMCR) | 1);
+    atomic::fence(atomic::Ordering::SeqCst);
+
+    let len =
+        (&__eitcm as *const u32).offset_from(&__sitcm as *const _) as usize;
+    let dst = slice::from_raw_parts_mut(&mut __sitcm as *mut _, len);
+    let src = slice::from_raw_parts(&__siitcm as *const _, len);
+    dst.copy_from_slice(src);
+
+    atomic::fence(atomic::Ordering::SeqCst);
```
ryan-summers, line 129: We have CPU fences, but not compiler fences. Should we add an `atomic::compiler_fence(SeqCst)` here as well to prevent the compiler from re-ordering accesses here?

ryan-summers: After reading more, I think an `atomic::fence(SeqCst)` is essentially a more powerful `compiler_fence()`, so I think this is okay as-is.

ryan-summers (changes requested):

src/hardware/configuration.rs
jordens: It is.

src/hardware/configuration.rs
jordens: As mentioned, this can't be pre_init because for example the stack isn't set up. It can be earlier in setup() but there is no real point in doing that. Manually calling it seems fine. We are calling pretty much everything manually.

src/hardware/configuration.rs
jordens: That would be wrong.

jordens: I'm not a fan of the additional underscore before the type. Looks like it's commonly done without. E.g. https://doc.rust-lang.org/rust-by-example/primitives/literals.html
There is no underscore after the 0x either.

src/hardware/configuration.rs
ryan-summers: The stack is set up before `Reset` is even called - this is done by the ARM architecture - the SP is configured as the first entry in the vector table and is part of the CPU boot process:
https://github.com/rust-embedded/cortex-m-rt/blob/master/link.x.in#L63

If the stack wasn't set up, one could _never_ use the pre-init hook.

At this stage, the only thing not initialized is the FPU and BSS/DATA. Given that we only use a single constant (which is RODATA, stored in flash), we should be initialized enough to execute this function in the pre-init hook (this is what the pre-init hook is designed for).

The reason I'm pushing this is that if we don't do this as early as possible during startup, we are leaving a foot-gun around for ourselves later. If we (for some reason) put some init code in the `ITCM` in the future and forget exactly where the ITCM is initialized, we might kill hours debugging.

Did you try running this as a `pre_init()` hook already and observe that it didn't work?

src/hardware/configuration.rs
ryan-summers: While I acknowledge that my original comment had a typo (which was fixed), my comment was around the underscore betweeo EF90 and usize. I still haven't decided which form I favor more, but in all honesty, no one is going to be passively glancing at this, and it's easy enough to read when you actually do care.

src/hardware/configuration.rs
ryan-summers: These are good points - I would encourage you to add the discussion here (relevant points thereof) as  documentation to this initialization function

jordens:
bors merge

jordens:
bors cancel

jordens:
bors merge

PR #327, opened 2021-04-07
HITL Support

ryan-summers:
This PR integrates initial HITL support for stabilizer.

The HITL:
* Pings stabilizer after programming
* Updates an MQTT setting
* Requests a run automatically when a PR is made in to `master`.
* Runs automatically every night

In the future, this will be expanded to do the following:
* Verify loopback gain and AFE configuration
* Append a check result to pull requests


hitl/run.sh
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/bash -e
```
jordens, line 1:
```suggestion
#!/bin/sh
```

For compatibility and fewer assumptions.

hitl/run.sh
```diff
@@ -0,0 +1,22 @@
...
+# Set up python for testing
+python3 -m venv --system-site-packages py
+source py/bin/activate
+python3 -m pip install -r requirements.txt
+
+# Test pinging Stabilizer. This exercises that:
+# * DHCP is functional and an IP has been acquired
+# * Stabilizer's network is functioning as intended
+# * The stabilizer application is opeerational
+ping -c 5 -w 20 gonnigan.ber.quartiq.de
```
jordens, line 19:
```suggestion
ping -c 5 -w 20 stabilizer-hitl
```

hitl/run.sh
```diff
@@ -0,0 +1,22 @@
...
+
+# Set up python for testing
+python3 -m venv --system-site-packages py
+source py/bin/activate
+python3 -m pip install -r requirements.txt
+
+# Test pinging Stabilizer. This exercises that:
+# * DHCP is functional and an IP has been acquired
+# * Stabilizer's network is functioning as intended
+# * The stabilizer application is opeerational
```
jordens, line 18:
```suggestion
# * The stabilizer application is operational
```

hitl/run.sh
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/bash -e
+
```
jordens, line 2: `set -eux` is usually a good choice. Otherwise commands will silently fail and not be printed.

.github/workflows/hitl.yml
```diff
@@ -1,9 +1,9 @@
 name: HITL
 
 on:
-  push:
-    branches: [ hitl ]
   workflow_dispatch:
+  pull_request:
+    branches: [ master ]
```
jordens, line 6: Then we can also add a cron schedule here. The only downside is appears to be the deploy review requests piling up. Let's see if we can handle that.

Also I think the best approach would be to have a `repository-dispatch` action that waits for the dispatch to complete for the token to be valid long enough. I don't know whether the ones available are stable and high quality enough to get us there. Alternatively let's create a PAT on `quartiq/hitl` that has write access. Might be less hassle.

ryan-summers: 1. Let's do the cron job in the `hitl` repository so that we don't need to approve the jobs. It should be relatively easy there.
2. There's a lot of complexity involved in making a `repository-dispatch` wait for the scheduled workflow to complete. How do you even know the `repository-dispatch` generates a job? What if it generates multiple? I think the simplest solution here is to just create a persistent PAT. They're easy to revoke if you don't want to continue using it in the future.

hitl/run.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
```
ryan-summers, line 1: @jordens I had to revert back to using `bash` here instead of `sh`. We use the `source` command` below, which is explicit to `bash` and is unavailable in `sh`.

.github/workflows/hitl.yml
jordens: Agreed on both.

But the problem with the cron job on `hitl` is that it doesn't have the dispatch object data and might be tricky to write/duplicate.

I think you'd just wait for the check that `hitl` would cross-post here (not wait for the job). Like `https://github.com/fountainhead/action-wait-for-check` or similar. But yes. Let's do PAT first and get the check link cross-posted first.

hitl/run.sh
```diff
@@ -0,0 +1,25 @@
...
+# Description:
+#   This shell file is executed by the hardware runner in Quartiq's office to exercise the various
+#   hardware aspects of Stabilizer.
+
+# Enable shell operating mode flags.
+set -eux
+
+# Set up python for testing
+python3 -m venv --system-site-packages py
+source py/bin/activate
```
jordens, line 15: `source` is a bashism. Use `.` or better yet, in this case directly call the desired command:

`py/bin/pip install ...`
and
`py/bin/python miniconf.py ...`

hitl/run.sh
jordens: `source` is a bashism. https://wiki.ubuntu.com/DashAsBinSh#source

hitl/run.sh
ryan-summers: Is there a real reason to go through these steps instead of just using `bash`? Github actions always uses bash and this is already specific to our GHA setup given the hostnames in use.

jordens: I'm more worried about the reproducibility if you source. The `py` env isn't deleted and recreated and at some point there might be a `ping` in `py/bin` because of some mess-up with `requirements.txt`. That will completely confuse everyone. Admittedly a bit unlikely but I regularly see python environments breaking for users exactly like this because they are `sourced` on top of each other or the user forgot to activate/deactivate or had that in `bashrc`/`profile` etc: Then you call `pip` thinking you call it in that environment but in fact you are installing in some other environment.

jordens: tl;dr I'm more worried about `source` than about the type of shell.

jordens: I'm also not insisting. Just explaining my past experience and the trouble some users find themselves in with python environments and changing PATHs like this.

jordens:
Sorry. I accidentally rejected that deployment there.

hitl/run.sh
ryan-summers: The environment is automatically cleaned up when GHA completes execution - each `run` step is executed in a temporary shell script, so any environment changes exist purely in the GHA run step and are automatically cleaned up on exit.

hitl/README.md
```diff
@@ -0,0 +1,25 @@
...
+repository in order to provide secure hardware testing of the stabilizer application in a public
+repository.
+
+**Note**: In order to ensure application security, all HITL runs must first be approved by a Quartiq
+representative before execution.
+
+# Hardware Configuration
+Stabilizer is configured with an ethernet connection to a router. The router runs a DHCP server for
+the local network, and Stabilizers MAC address has been assigned to the unique host name of
+`gonnigan.ber.quartiq.de`.
```
jordens, line 15:
```suggestion
# Configuration
* Stabilizer is configured with an ethernet connection to a router. The router runs a DHCP server for
the local network, and ensures that the Stabilizer used for these tests is available under the hostname `stabilizer-hitl`.
* An MQTT broker is running at the hostname `mqtt`.
```

jordens (approved):

PR #330, opened 2021-04-08
Fixing ACR write padding

### LINKED ISSUE #264: Pounder profile builder performs short ACR write
@jordens noted that it looks like the Pounder profile builder does a short write of the ACR register. The register is 24-bits, not 16. 

https://github.com/quartiq/stabilizer/blob/5951a0d41d313f354a4813252ff7856abd6f4fbe/ad9959/src/lib.rs#L585

ryan-summers:
This PR fixes #264 by properly assigning the ACR write to a 3 byte slice, which is equivalent to the size of the register in the AD9959.

A small number of other issues were uncovered and fixed in regard to the Pounder profile writer.

TODO:
- [x] Test ACR writes on pounder hardware

**Testing**
Verified that the FLS binary (not public) functioned with ACR updated to 99.9% amplied (1 less than FS).

### DIFF
--- a/ad9959/src/lib.rs
+++ b/ad9959/src/lib.rs
@@ -610,10 +610,10 @@ impl ProfileSerializer {
             1 => {
                 // For a pad size of 1, we have to pad with 5 bytes to align things.
                 self.add_write(Register::CSR, &[(self.mode as u8) << 1]);
-                self.add_write(Register::LSRR, &[0, 0, 0]);
+                self.add_write(Register::LSRR, &[0, 0]);
             }
             2 => self.add_write(Register::CSR, &[(self.mode as u8) << 1]),
-            3 => self.add_write(Register::LSRR, &[0, 0, 0]),
+            3 => self.add_write(Register::LSRR, &[0, 0]),
             4 => {}
 
             _ => unreachable!(),
--- a/ad9959/src/lib.rs
+++ b/ad9959/src/lib.rs
@@ -558,13 +558,14 @@ impl ProfileSerializer {
     /// * `channels` - A list of channels to apply the configuration to.
     /// * `ftw` - If provided, indicates a frequency tuning word for the channels.
     /// * `pow` - If provided, indicates a phase offset word for the channels.
-    /// * `acr` - If provided, indicates the amplitude control register for the channels.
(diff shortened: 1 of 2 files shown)


ad9959/src/lib.rs
```diff
@@ -582,7 +582,9 @@ impl ProfileSerializer {
         }
 
         if let Some(acr) = acr {
-            self.add_write(Register::ACR, &acr.to_be_bytes());
+            let mut data = [0; 3];
+            data[1..=2].copy_from_slice(&acr.to_be_bytes());
```
jordens, line 586: Better have `acr: Option<u32>` and copy the 3 low bytes.
We should cover the entire register (among other things, `Amplitude multiplier enable` to be set as well).

ryan-summers: The upper 8 bytes is actually just the ramp control register, which we don't really support at all right now, so that's why I was opting to default them to zero. The enable bit is bit 4 of ACR[1], which fits in the lower 16 bits

jordens: If we take the full ACR the full ramp functionality is supported and there is no mismatch between the meaning of `Acr` and the actual register content. Why restrict it to the 16 LSBs? That still exposes bits (15, 14, 11, 10) which then can/must not be used because of that restriction.

ad9959/src/lib.rs
```diff
@@ -558,13 +558,14 @@ impl ProfileSerializer {
     /// * `channels` - A list of channels to apply the configuration to.
     /// * `ftw` - If provided, indicates a frequency tuning word for the channels.
     /// * `pow` - If provided, indicates a phase offset word for the channels.
-    /// * `acr` - If provided, indicates the amplitude control register for the channels.
+    /// * `acr` - If provided, indicates the amplitude control register for the channels. The ACR
+    ///   should be stored in the 3 LSB of the word.
```
jordens, line 562:
```suggestion
    /// * `acr` - If provided, indicates the amplitude control register for the channels. The ACR
    ///   should be stored in the 3 LSB of the word. Note that if amplitude scaling is to be used,
    ///   the "Amplitude multiplier enable" bit must be set.
```

jordens (approved):
Minor documentation addition. Otherwise LGTM!

PR #332, opened 2021-04-09
Adding check wait for HITL

ryan-summers:
This PR integrates the HITL into the PR check process - specifically, the HITL now posts a check onto the PR that we can require to pass.

This required the following:
1. Wait for the check to complete
2. Wait for a startup period (in case of re-starting check) so that a previously-passed check isn't detected in (1)

### DIFF
--- a/.github/workflows/hitl.yml
+++ b/.github/workflows/hitl.yml
@@ -17,8 +17,8 @@ jobs:
           repository: quartiq/hitl
           client-payload: '{"github": ${{ toJson(github) }}}'
 
-    - uses: fountainhead/action-wait-for-check@v1.0.0
-      with:
-        token: ${{ secrets.GITHUB_TOKEN }}
-        checkName: HITL
-        ref: ${{ github.event.pull_request.head.sha }}
+      - uses: fountainhead/action-wait-for-check@v1.0.0
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          checkName: HITL
+          ref: ${{ github.event.pull_request.head.sha }}
--- a/.github/workflows/hitl.yml
+++ b/.github/workflows/hitl.yml
@@ -18,7 +18,7 @@ jobs:
           client-payload: '{"github": ${{ toJson(github) }}}'
 
       - name: Wait for startup
-        runs: sleep 30
+        run: sleep 30
 
       - uses: fountainhead/action-wait-for-check@v1.0.0
         with:
--- a/.github/workflows/hitl.yml
+++ b/.github/workflows/hitl.yml
@@ -1,4 +1,4 @@
-name: HITL
+name: HITL Run Request
 
 on:
   workflow_dispatch:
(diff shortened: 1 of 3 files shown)


.github/workflows/hitl.yml
```diff
@@ -16,3 +16,9 @@ jobs:
           event-type: stabilizer
           repository: quartiq/hitl
           client-payload: '{"github": ${{ toJson(github) }}}'
+
+      - uses: fountainhead/action-wait-for-check@v1.0.0
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          checkName: HITL
```
jordens, line 23: That's itself. That should block forever.

I think you want to rename [this one](https://github.com/quartiq/hitl/blob/main/.github/workflows/stabilizer.yml#L20), get it working and wait for it.

ryan-summers: https://github.com/quartiq/stabilizer/pull/332/checks?check_run_id=2304646861 seems to indicate it doesn't actually detect itself. In any case, I'm aware of this and planning on addressing it at a later step. This PR is still a draft as a means to force GHA and test the checks mechanism, so isn't ready for review at this point

ryan-summers: Also worth mentioning here after investigation: `HITL` isn't the name of this check - it's the name of the "check suite", so it doesn't actually register when querying for check results. The name of this check is actually the name of each job in this workflow, so it's likely some randomly assigned ID (we could specify it with `id`, but we don't care here).

This explains why it wasn't picking up itself earlier.

.github/workflows/hitl.yml
```diff
@@ -1,4 +1,4 @@
-name: HITL
+name: HITL Run Request
```
jordens, line 1: Maybe

```suggestion
name: HITL Trigger
```

And assuming the filename defines the bare "hitl" in https://github.com/quartiq/stabilizer/runs/2305293775, can re rename it to `.github/workflows/hitl_trigger.yml`?

.github/workflows/hitl.yml
```diff
@@ -16,3 +16,12 @@ jobs:
...
           repository: quartiq/hitl
           client-payload: '{"github": ${{ toJson(github) }}}'
+
+      - name: Wait for startup
+        run: sleep 30
+
+      - uses: fountainhead/action-wait-for-check@v1.0.0
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          checkName: HITL
```
jordens, line 26:
```suggestion
          checkName: HITL Run Status
```

* Can we move the `cargo flash` step into `hitl/run.sh`?
* Where is the `details_url` ending up? It would be nice to have that link from the commit checks on `stabilizer` to the hitl run on `hitl` (for those who can access it).

jordens (changes requested):
Cool!

.github/workflows/hitl.yml
ryan-summers: Re: `details_url` - this appears to be a broken portion of the GHA API: https://github.com/LouisBrunner/checks-action/issues/18

jordens (approved):
Nice

Issue #276, opened 2021-02-19
Discover MQTT topic for multiple Stabilizers with same firmware


dnadlinger:
Let's say Stabilizer was used for every relevant control task in my lab. I might end up with 20 different boards, maybe running some 4 different firmware variants (dual PI controller, Pounder PDH lock, specialty cascaded/lock-in/… configurations).

As far as I can tell, the design intent behind the just-merged MQTT interface is to have topics of the form `<device_id>/settings/path`, where `<device_id>` is some user-defined identifier. Ideally, it would not be necessary to compile a separate firmware image for each device, so the device id needs to be read from some sort of persistent storage.

What about including provisions for a default topic root of `stabilizer_<mac_address>`, where `mac_address` is the EUI-48 for the network interface (as read over I2C on boot) as a lower-case hex string?

Additionally, it is probably a good idea to make this configurable, for user-friendliness as well as to support swapping out hardware without having to modify the client code. (Or is it? There could be a separate name resolution layer instead…) In that case, this would most naturally be via a miniconf setting then written to EEPROM/flash, but then should miniconf start using the new device id at runtime, or should we wait until the next power cycle?

jordens:
https://github.com/quartiq/stabilizer/issues/54#issuecomment-607478814

dnadlinger:
Agreed re. topic design guide, but that still leaves the question of what the default interface should be/how to get the topic root configured after unboxing a new Stabilizer.

jordens:
Let it get the root from `dt/sinara/stabilizer/{eui48}/name` and default to just that if unavailable.
Since we have MQTT and since it's persistent, we can just use it and WILLs to store all settings.
MAC address becomes known through a sticker on the device with a QR code, through DHCP/ARP, and through a telemetry to the default topic.

dnadlinger:
What happens if `/name` is set while the Stabilizer in question is running? The device switches to the new root, and publishes the current configuration there?

jordens:
It doesn't republish configuration. But it would unsubscribe from settings messages on the previous name and subscribe to the new one.

jordens:
There is a general issue with the default settings being unknown on MQTT. That also applies here.

dnadlinger:
Usually, the current configuration would be available from retained messages, though, wouldn't it?

jordens:
Exactly.
[...]

PR #337, opened 2021-04-13
Utilize unique MQTT prefix identifiers

ryan-summers:
This PR addresses #276 by updating the API to utilize a default MQTT prefix calculated via the binary name and the MAC address.

In the future, this address will be persisted to flash bank 2 for persistence between flash operations and will be loaded on boot. It will also be configurable via MQTT.

This must be merged after https://github.com/quartiq/stabilizer/pull/314

### DIFF
--- a/src/bin/dual-iir.rs
+++ b/src/bin/dual-iir.rs
@@ -59,7 +59,10 @@ const APP: () = {
         let mqtt_settings = MqttSettings::new(
             stabilizer.net.stack,
             "",
-            "dt/sinara/stabilizer",
+            &net::get_device_prefix(
+                env!("CARGO_BIN_NAME"),
+                stabilizer.net.mac_address,
+            ),
             stabilizer.net.phy,
             stabilizer.cycle_counter,
         );
--- a/src/hardware/configuration.rs
+++ b/src/hardware/configuration.rs
@@ -57,6 +57,7 @@ impl NetStorage {
 pub struct NetworkDevices {
     pub stack: NetworkStack,
     pub phy: EthernetPhy,
+    pub mac_address: smoltcp::wire::EthernetAddress,
 }
 
 /// The available hardware interfaces on Stabilizer.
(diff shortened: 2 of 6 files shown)


src/net/mod.rs
```diff
@@ -107,3 +110,39 @@ where
...
+/// * `mac` - The ethernet MAC address of the device.
+///
+/// # Returns
+/// The MQTT prefix used for this device.
+pub fn get_device_prefix(
+    app: &str,
+    mac: smoltcp_nal::smoltcp::wire::EthernetAddress,
+) -> String<consts::U128> {
+    let mac_string = {
+        let mut mac_string: String<consts::U32> = String::new();
```
jordens, line 127: Why not `U18` then? Or do it together with the app name in one single `write!()`?

jordens (changes requested):

src/hardware/configuration.rs
```diff
@@ -551,7 +552,7 @@ pub fn setup(
             smoltcp::iface::NeighborCache::new(&mut store.neighbor_cache[..]);
 
         let interface = smoltcp::iface::EthernetInterfaceBuilder::new(eth_dma)
-            .ethernet_addr(mac_addr)
+            .ethernet_addr(mac_addr.clone())
```
jordens, line 555: As, clippy says, I think this should just be a copy.

src/bin/dual-iir.rs
```diff
@@ -62,7 +62,10 @@ const APP: () = {
         let mqtt_config = MiniconfInterface::new(
             stabilizer.net.stack,
             "",
-            "dt/sinara/stabilizer",
```
jordens, line 65: `hitl/run.sh` will need adapting as well

src/net/mod.rs
ryan-summers: This uses U32 because it's the largest power of two that supports the 17 byte string (5 hyphens + 2 * 6 characters). I'm inclined to leave it as-is for now since I don't see an apparent benefit to dropping down the buffer size.

ryan-summers: Also, this is done as a two-stage write to make the format clear and easy to process. Writing it all in one go would result in some weird looking format strings that would be hard to interpret. Feel free to propose alternatives though.

jordens: All good!

jordens (approved):

PR #338, opened 2021-04-14
hardware: add digital input support

jordens:
* The inputs of the buffer are not pulled up/down. That might make them
  unusable if left floating. But since the hold feature is disabled via MQTT `allow_hold` by default, there is no regression.

close: #70

### DIFF
--- a/src/bin/dual-iir.rs
+++ b/src/bin/dual-iir.rs
@@ -4,6 +4,8 @@
 
 use stm32h7xx_hal as hal;
 
+pub use embedded_hal::digital::v2::InputPin;
+
 use stabilizer::hardware;
 
 use miniconf::{minimq, Miniconf, MqttInterface};
@@ -12,7 +14,7 @@ use serde::Deserialize;
 use dsp::iir;
 use hardware::{
     Adc0Input, Adc1Input, AfeGain, CycleCounter, Dac0Output, Dac1Output,
-    NetworkStack, AFE0, AFE1,
+    DigitalInput1, NetworkStack, AFE0, AFE1,
 };
 
 const SCALE: f32 = i16::MAX as _;
@@ -39,6 +41,7 @@ impl Default for Settings {
 const APP: () = {
     struct Resources {
         afes: (AFE0, AFE1),
+        di1: DigitalInput1,
         adcs: (Adc0Input, Adc1Input),
         dacs: (Dac0Output, Dac1Output),
         mqtt_interface:
@@ -90,6 +93,7 @@ const APP: () = {
             adcs: stabilizer.adcs,
             dacs: stabilizer.dacs,
             clock: stabilizer.cycle_counter,
+            di1: stabilizer.digital_inputs.1,
         }
     }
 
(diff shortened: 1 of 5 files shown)


src/bin/dual-iir.rs
```diff
@@ -90,6 +99,8 @@ const APP: () = {
             adcs: stabilizer.adcs,
             dacs: stabilizer.dacs,
             clock: stabilizer.cycle_counter,
+            di1: stabilizer.digital_inputs.1,
```
ryan-summers, line 102: I would recommend something more descriptive for this variable name. `di1` is somewhat opaque. What about `digital_in1` or something?

src/hardware/configuration.rs
```diff
@@ -9,12 +9,12 @@ use stm32h7xx_hal::{
 
 use smoltcp_nal::smoltcp;
 
-use embedded_hal::digital::v2::{InputPin, OutputPin};
+pub use embedded_hal::digital::v2::{InputPin, OutputPin};
```
ryan-summers, line 12: Why make this a public use?

src/hardware/configuration.rs
```diff
@@ -439,6 +440,12 @@ pub fn setup(
         )
     };
 
+    let digital_inputs = {
+        let di0 = gpiog.pg9.into_floating_input();
```
ryan-summers, line 444: That's somewhat unfortunate that we don't have any pull on the digital inputs - leaving them floating doesn't feel right, but we can't do much after the input buffer stage.

ryan-summers (commented):
Some few minor comments

src/hardware/configuration.rs
jordens: Actually I wanted to import the trait from here and not from the hal. I'll fix that instead.

Cross-importing (in dual-iir) from the hal to bring a trait into scope for a type/struct that's defined somewhere else always annoys me. The usual `prelude` approach to get those traits also doesn't convince me. Maybe I'm doing it wrong.

src/hardware/configuration.rs
jordens: https://github.com/sinara-hw/Stabilizer/issues/91

src/hardware/configuration.rs
ryan-summers: I tend to approach it as "import where necessary" - so I just import `embedded_hal::digital::v2::InputPin` wherever I need to use it.

jordens: But that has two big disadvantages:

* Assume that the point of use is a different crate. Then you are needlessly creating a new dependency from the using crate to the trait crate. Re-exporting it would avoid that.
* It seems like a completely unrelated import. It's unclear from reading alone where (on what struct) that trait is actually used.

That seems to match https://github.com/rust-lang/api-guidelines/discussions/176 and could be addressed by the guidance described there.

jordens: Since this is a bit bigger than just this single instance, I'll leave it like it is currently (re-export the trait) and leave the hole topic to be addressed en-bloc later.

PR #340, opened 2021-04-19
TCP source port ranodmization

ryan-summers:
This PR adds support for randomizing the TCP port usage by the NAL. This helps to keep sockets flowing properly between resets.

## Testing

This branch was run numerous times on Stabilizer and Wireshark was used to observe traffic. It was observed that Stabilizer used different TCP ports in between each boot.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -716,7 +716,7 @@ dependencies = [
 [[package]]
 name = "smoltcp-nal"
 version = "0.1.0"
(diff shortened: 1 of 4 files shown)


jordens (approved):

jordens:
https://github.com/quartiq/stabilizer/runs/2381426594?check_suite_focus=true I think the ping timeout should be a bit longer, like 1 minute. It came up a bit later. It might just take a bit due to all the waiting, the EEPROM, switch STP port gating, DHCP.

PR #341, opened 2021-04-20
Feature/telemetry

### LINKED ISSUE #149: MQTT settings/status/telemetry framework

ryan-summers:
This PR fixes #149 

This PR adds the following:
* A `SystemTimer` used for RTIC scheduling (required for schedule periods > ~2 seconds)
* A refactor of the `src/net` directory
* Addition of network stack sharing via proxy (identical to implementation from `shared-bus`)
* Addition of a telemetry task

This must be merged after #352 

Future Work:
* move `src/net/shared.rs` into `embedded-nal` or some other separate crate.
* Investigate removal of shared network stack by making smoltcp sockets owned and interrupt-safe


src/hardware/system_timer.rs
```diff
@@ -0,0 +1,117 @@
...
+                }
+
+                let current_value = regs.cnt.read().bits();
+
+                // Check that an overflow didn't occur since we just cleared the overflow bit. If
+                // it did, loop around and retry.
+                if regs.sr.read().uif().bit_is_clear() {
+                    // Note(unsafe): We are in a critical section, so it is safe to read the
+                    // global variable.
+                    unsafe { Some((OVERFLOWS * 65535 + current_value) as i32) }
```
jordens, line 94: Sure it's `* 65535` and not `OVERFLOWS << 16`?

src/net/telemetry.rs
```diff
@@ -0,0 +1,168 @@
...
+        // integers, and u16::MAX / 2 is considered zero voltage output. Thus, the dynamic range of
+        // the output stage is +/- 10.24 V. At a DAC code of zero, there is an output of -10.24 V,
+        // and at a max DAC code, there is an output of 10.24 V.
+        let out0_volts = (10.24 * 2.0)
+            * (self.latest_outputs[0] as f32 / (u16::MAX as f32))
+            - 10.24;
+        let out1_volts = (10.24 * 2.0)
+            * (self.latest_outputs[1] as f32 / (u16::MAX as f32))
+            - 10.24;
+
```
jordens, line 102: Let's collect the scaling factors.

```suggestion
        let volt_per_lsb = (5.0 * 4.096 / 2.0 / i16::MAX as f32);
        let in0_volts =
            self.latest_samples[0] as f32 * (volt_per_lsb / afe0.as_multiplier());
        let in1_volts =
            self.latest_samples[1] as f32 * (volt_per_lsb / afe1.as_multiplier());

        // The output voltage is generated by the DAC with an output range of +/- 4.096 V. This
        // signal then passes through a 2.5x gain stage. Note that the DAC operates using unsigned
        // integers, and u16::MAX / 2 is considered zero voltage output. Thus, the dynamic range of
        // the output stage is +/- 10.24 V. At a DAC code of zero, there is an output of -10.24 V,
        // and at a max DAC code, there is an output of 10.24 V.
        let volt_per_lsb = (10.24 * 2.0 / u16::MAX as f32);
        let out0_volts = (self.latest_outputs[0] ^ 0x8000) as i16 as f32 * volt_per_lsb;
        let out1_volts = (self.latest_outputs[1] ^ 0x8000) as i16 as f32 * volt_per_lsb;
```

src/net/telemetry.rs
```diff
@@ -0,0 +1,168 @@
...
+/// run-time overhead during conversion to SI units.
+#[derive(Copy, Clone)]
+pub struct TelemetryBuffer {
+    /// The latest input sample on ADC0/ADC1.
+    pub latest_samples: [i16; 2],
+    /// The latest output code on DAC0/DAC1.
+    pub latest_outputs: [u16; 2],
+    /// The latest digital input states during processing.
+    pub digital_inputs: [bool; 2],
+}
```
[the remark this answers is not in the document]
ryan-summers: The intent of this code is to provide a generic telemetry implementation for some basic Stabilizer functionality, _not_ that this is the only type of telemetry that can be transmitted. This code is generally defined to provide a single place for converting basic things like DAC/ADC codes into analog voltage levels.

It is not meant to imply that this is the only type of telemetry supported - the telemetry client below accepts anything that is `Serialize`-able. This is just basic telemetry that I figured would be somewhat universal for all applications (where some may want to report more telemetry, e.g. Pounder).

Ditto on u16s, after analyzing how the codes work further, I agree that u16 all around is simpler.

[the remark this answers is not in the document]
ryan-summers: Also worth noting - this data is not serializeable and is never reported as actual telemetry (this is just a buffer to generate telemetry from), so the names are far less important as they are never exposed to the user.

src/bin/lockin.rs
```diff
@@ -229,31 +235,67 @@ const APP: () = {
                 *sample = value as u16 ^ 0x8000;
             }
         }
+
+        // Update telemetry measurements.
+        c.resources.telemetry.latest_samples =
+            [adc_samples[0][0] as i16, adc_samples[1][0] as i16];
```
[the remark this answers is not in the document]
ryan-summers: See comments above - the ADC chip actually interprets the data as i16 (whereas the DAC interprets data as u16), so I feel it makes the math far easier to follow if we keep them as i16, but I'm open to your thoughts here.

src/net/telemetry.rs
ryan-summers: Just a heads up that the DAC calculation is actually a linear transform with an offset of -10.24 volts, so there's both a slope + offset component, whereas the ADCs have just a slope component.

src/hardware/system_timer.rs
```diff
@@ -0,0 +1,117 @@
...
+                if regs.sr.read().uif().bit_is_set() {
+                    regs.sr.modify(|_, w| w.uif().clear_bit());
+                    unsafe {
+                        OVERFLOWS += 1;
+                    }
+                }
+
+                let current_value = regs.cnt.read().bits();
+
+                // Check that an overflow didn't occur since we just cleared the overflow bit. If
```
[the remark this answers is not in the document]
ryan-summers: This has the downside of potentially increasing interrupt latency in the application, although I would never expect this loop to occur more than two times in a row if it blocks pre-emption.

src/net/telemetry.rs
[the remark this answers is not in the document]
jordens: Ok. Then let's keep this basic telemetry message shared.

The ADC data is indeed two's complement and encoded as u16. The DAC data is "offset binary" and encoded as u16.
The encoding is always somewhat opaque.
You don't need to do any XORing to convert between i16 and u16 for ADC data. That's incorrect. Only the DAC data needs XORing. And XORing is the exact same operation as `-10.24` in f32.
Certainly a nice idiomatic solution would be to make newtypes for the ADC and DAC data with `from_raw()`, `as_raw()`, `From<i16>` and `From<f32>`. Then it's clear that we have encoded data and it's clear the the conversion can and should only happen at one point, not split like in the case of the ADC data now.
Yes. It's not exposed to the user, but still the names should be clear and consistent. The rename change is meant to help developers understand what's going on. I was utterly confused by `latest_samples` and wondered why this only exists for the ADC data and not for other "samples".

src/net/telemetry.rs
jordens: But the XOR of the MSB is right there in the table. That's not a lot of math.
If the XOR is frightening, and to be closer to your understanding you can also do `code.wrapping_add(0x8000)` or `code.wrapping_sub(0x8000)`. All three are exactly the same.

jordens: Calculation based on decimal constants like `32767` are really tricky and so frequently riddled with problems/bugs. Remember all the code-to-SI off-by-one conversions in Booster, Pounder or https://github.com/quartiq/stabilizer/pull/341#discussion_r628054750. These are binary systems. Why talk to them in decimal numbers?

src/net/telemetry.rs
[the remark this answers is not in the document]
ryan-summers: I've added some utilities in the `DacCode` and `AdcSample` types, as I like your approach of encoding the raw codes into types to make it clear what we're dealing with where. Let me know what you think.

src/net/telemetry.rs
jordens: And to make my point: the same error here again. The conversion isn't correct. The DAC scale isn't `10.24 * 2.0 / u16::MAX as f32` but `4.096 * 2.5 / (1 << 15) as f32`. That's different.

src/hardware/dac.rs
```diff
@@ -69,6 +69,39 @@ use hal::dma::{
...
+        // signal then passes through a 2.5x gain stage. Note that the DAC operates using unsigned
+        // integers, and u16::MAX / 2 is considered zero voltage output. Thus, the dynamic range of
+        // the output stage is +/- 10.24 V. At a DAC code of zero, there is an output of -10.24 V,
+        // and at a max DAC code, there is an output of 10.24 V.
+        //
+        // Note: The MAX code corresponding to +VREF is not programmable, as it is 1 bit larger
+        // than full-scale.
+        let dac_volts_per_lsb = 10.24 * 2.0 / (u16::MAX as u32 + 1) as f32;
+
+        (self.0 as f32) * dac_volts_per_lsb - 10.24
```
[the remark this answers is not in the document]
jordens: What distortion?

src/net/telemetry.rs
jordens: And finally also the ADC scale isn't `5.0 / 2.0 * 4.096 / i16::MAX as f32` but `5.0 * 4.096 / 2.0 / (1 << 15) as f32`. Again off-by-one.

ryan-summers: Just for clarities sake, you're neglecting to incorporate the offset voltage, which corresponds to the MSB. In any case, I actually think the XOR is more appropriate as I failed to note originally that the MAX value is actually a step below full-scale, so I'm updating to use that as well (especially since this will all be encapsulated now)

src/hardware/adc.rs
```diff
@@ -84,6 +84,28 @@ use hal::dma::{
     MemoryToPeripheral, PeripheralToMemory, Transfer,
 };
 
+/// A type representing an ADC sample.
+#[derive(Copy, Clone)]
+pub struct AdcSample(pub u16);
```
jordens, line 89: Why `Sample` for one but `Code` for the other? Please use `Code` for both.
And `impl From<AdcCode> for f32` instead of `Into` as suggested.

jordens: Or doesn't that work because of the orphaning rules?

jordens (changes requested):

src/hardware/adc.rs
ryan-summers: Given that the IIR coefficients handle the mapping of f32 -> code, I assumed we could forgo explicitly providing that type conversion.

jordens: DAC output data are samples with the exactly the same semantics as ADC input data samples.
Ack the `Into` preference!
To have SI semantics for the IIR filter we need to convert `y_offset`, `y_min` and `y_max`. The coefficients are agnostic to scaling because the scaling is the same here (apart from the AFE gain which I don't want to bring into the IIR equation).

src/hardware/dac.rs
```diff
@@ -69,6 +69,40 @@ use hal::dma::{
...
+
+impl Into<f32> for DacCode {
+    fn into(self) -> f32 {
+        // The output voltage is generated by the DAC with an output range of +/- 4.096 V. This
+        // signal then passes through a 2.5x gain stage. Note that the DAC operates using unsigned
+        // integers, and u16::MAX / 2 is considered zero voltage output. Thus, the dynamic range of
+        // the output stage is +/- 10.24 V. At a DAC code of zero, there is an output of -10.24 V,
+        // and at a max DAC code, there is an output of (slightly less than) 10.24 V.
+
+        let dac_volts_per_lsb = 10.24 * 2.0 / u16::MAX as f32;
```
jordens, line 85: The scaling is wrong again ;) Here and the same for `DacCode`.
And `u16::MAX/2` is not zero voltage output. Code `(1 << 15)` is.

It's much better to let the compiler do the math and just write down the electronics.


```suggestion
        let dac_volts_per_lsb = 4.096 * 2.5 / (1u16 << 15) as f32;
```

src/hardware/adc.rs
jordens: Let's defer the IIR parameter refactoring to another time.

ryan-summers: Yeah, I'm still not convinced the IIR needs to change anyways, since the coefficients are pretty much meaningless numbers to a human anyways and are generally calculated by some other utility.

src/hardware/system_timer.rs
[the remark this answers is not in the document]
jordens: In fact, it's guaranteed by design that you will have to check exactly twice. You could unroll the loop for more clarity: `check, (clear, incr), read, check, (clear, incr, read), calculate, return`

src/net/mod.rs
```diff
@@ -6,21 +6,145 @@
...
+        // a fixed length.
+        write!(
+            &mut mac_string,
+            "{:02x}-{:02x}-{:02x}-{:02x}-{:02x}-{:02x}",
+            mac[0], mac[1], mac[2], mac[3], mac[4], mac[5]
+        )
+        .unwrap();
+
+        mac_string
+    };
```
irwineffect, line 143: `EthernetAddress` already implements `Display`, does that not format it well?

irwineffect: Actually, the code is basically the same: https://docs.rs/smoltcp/0.4.0/src/smoltcp/wire/ethernet.rs.html#70-76

jordens (changes requested):
A couple fixes for off-by-one errors and documentation consolidation.

src/net/mod.rs
ryan-summers: I didn't know it implemented display - definitely going to use that. Thanks for pointing this out

src/hardware/configuration.rs
```diff
@@ -7,32 +7,50 @@ use stm32h7xx_hal::{
...
     pub dhcp_tx_metadata: [smoltcp::socket::RawPacketMetadata; 1],
     pub dhcp_tx_storage: [u8; 600],
     pub dhcp_rx_storage: [u8; 600],
 }
 
+#[derive(Copy, Clone)]
+pub struct SocketStorage {
+    rx_storage: [u8; 4096],
+    tx_storage: [u8; 4096],
+}
```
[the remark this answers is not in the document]
ryan-summers: No, it really doesn't. In fact, I believe we have our MQTT message sizes capped at 256 bytes, so a 4K buffer allows us to buffer 16 messages, when we're typically only ever buffering 1. I'll lower the buffer size down to 1K.

src/hardware/configuration.rs
```diff
@@ -42,9 +60,8 @@ impl NetStorage {
             )],
             neighbor_cache: [None; 8],
             routes_cache: [None; 8],
-            sockets: [None, None],
-            tx_storage: [0; 4096],
-            rx_storage: [0; 4096],
+            sockets: [None, None, None, None, None],
```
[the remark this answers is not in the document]
ryan-summers: In this case, `None` does not impl `Copy`, because the underlying `SocketSetItem` that the option wraps does not implement `Copy`. As such, we can't use `None; 5`. I think if we upped to a minimum supported rust version (MSRV) of 1.51, we could do this, but with older rust releases, it's not possible.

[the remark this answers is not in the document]
jordens: Ack!

src/hardware/adc.rs
```diff
@@ -84,6 +84,28 @@ use hal::dma::{
...
+    ///
+    /// # Note
+    ///
+    /// This does not account for the programmable gain amplifier at the signal input.
+    fn into(self) -> f32 {
+        // The input voltage is measured by the ADC across a dynamic scale of +/- 4.096 V with a
+        // dynamic range across signed integers. Additionally, the signal is fully differential, so
+        // the differential voltage is measured at the ADC. Thus, the single-ended signal is
+        // measured at the input is half of the ADC-reported measurement. As a pre-filter, the
+        // input signal has a fixed gain of 1/5 through a static input active filter.
```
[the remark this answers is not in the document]
ryan-summers: The main goal here is to describe to the reader the purpose of the 1 / 2.0 scaling factor, which is justified with the single-to-differential signal conversion. I'm fine with the more terse description.

src/hardware/dac.rs
```diff
@@ -69,6 +69,40 @@ use hal::dma::{
...
+        // integers, and u16::MAX / 2 is considered zero voltage output. Thus, the dynamic range of
+        // the output stage is +/- 10.24 V. At a DAC code of zero, there is an output of -10.24 V,
+        // and at a max DAC code, there is an output of (slightly less than) 10.24 V.
+
+        let dac_volts_per_lsb = 4.096 * 2.5 / (1u16 << 15) as f32;
+
+        // Note that the bipolar table is an offset-binary code, but it is much more logical and
+        // correct to treat it as a twos-complement value. TO do that, we XOR the most significant
+        // bit to convert it.
+        (self.0 ^ 0x8000) as i16 as f32 * dac_volts_per_lsb
```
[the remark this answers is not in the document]
ryan-summers: I think keeping it as self.o ^ 0x8000 is more comprehensive than `i16::wrapping_add(-0x8000)` - it's still not really clear to me how `-0x8000 as i16` should actually be interpreted by the compiler.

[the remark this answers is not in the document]
jordens: `-0x8000` **is** a `i16` literal. No need for casting and no interpretation necessary.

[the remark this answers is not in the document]
ryan-summers: My concern is not whether or not `-0x8000` is a literal, but rather what does it _actually_represent. In that literal, there are _two_ specifies for the sign bit - the MSB, and the negative sign. Which should matter? Which takes precedent? Does two negative signs result in a positive?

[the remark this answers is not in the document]
ryan-summers: I replaced it with `i16::MIN` for undersandability.

src/hardware/dac.rs
```diff
@@ -69,6 +69,40 @@ use hal::dma::{
...
+    }
```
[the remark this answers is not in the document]
ryan-summers: There's some issues with placing tests in the `stabilizer` crate - specifically, Rust can't find the crate `test`. As such, I'm avoiding tests, but I think it would be value-added in the future to add some small sanity check tests (e.g. verify code of 0 equates to near our negative scale, code of 0xFFFF equates to near our positive scale).

jordens (approved):

ryan-summers:
bors merge

PR #347, opened 2021-04-28
MQTT/Miniconf refactor

ryan-summers:
This PR refactors the MQTT/Miniconf API to make use of the refactored miniconf API. Routing has been moved to the `net` module.

**TODO**:
- [x] Test on hardware
- [x] Document modules
- [x] Merge dependency changes, update dependencies to use `rev`
- [x] Update `miniconf.py` to be `stabilizer.py`, update API


ryan-summers:
bors try

ryan-summers:
bors try

ryan-summers:
bors cancel

ryan-summers:
bors try-

ryan-summers:
bors retry

ryan-summers:
bors try-

ryan-summers:
bors try-

src/net/mqtt_interface.rs
```diff
@@ -0,0 +1,235 @@
...
+
+/// MQTT settings interface.
+pub struct MqttInterface<S>
+where
+    S: miniconf::Miniconf + Default + Clone,
+{
+    telemetry_topic: String<consts::U128>,
+    default_response_topic: String<consts::U128>,
+    mqtt: RefCell<minimq::MqttClient<minimq::consts::U256, NetworkStack>>,
+    settings: RefCell<S>,
```
jordens, line 20: Why do we need `RefCell` for these? All methods that need the mutable ref to the client have a mutable ref to `self` already.

src/net/mqtt_interface.rs
```diff
@@ -0,0 +1,235 @@
...
+
+use core::{cell::RefCell, fmt::Write};
+
+use heapless::{consts, String};
+use serde::Serialize;
+
+use super::{Action, MqttMessage, SettingsResponse, SettingsResponseCode};
+
+/// MQTT settings interface.
+pub struct MqttInterface<S>
```
jordens, line 13: I think we should work towards making this stabilizer-agnostic as well. We'd want the same in Thermostat and Booster.

src/net/mqtt_interface.rs
```diff
@@ -0,0 +1,235 @@
...
+            Ok(updated) => !updated,
+            Err(err) => {
+                log::info!("Network error: {:?}", err);
+                false
+            }
+        };
+
+        // If the PHY indicates there's no more ethernet link, reset the DHCP server in the network
+        // stack.
+        if self.phy.poll_link() == false {
```
jordens, line 91: Better than the clippy lint below, this should return a meaningful [enum](https://rust-lang.github.io/api-guidelines/type-safety.html#arguments-convey-meaning-through-types-not-bool-or-option-c-custom-type).

src/net/mqtt_interface.rs
```diff
@@ -0,0 +1,235 @@
...
+        if !self.subscribed && mqtt_connected {
+            let mut settings_topic: String<consts::U128> = String::new();
+            write!(&mut settings_topic, "{}/settings/#", self.id.as_str())
+                .unwrap();
+
+            self.mqtt
+                .borrow_mut()
+                .subscribe(&settings_topic, &[])
+                .unwrap();
+            self.subscribed = true;
```
jordens, line 128: Let's at least state that we don't see and don't process potential subscription failures and why.

src/net/mqtt_interface.rs
```diff
@@ -0,0 +1,235 @@
...
+        let mut update = false;
+        match self.mqtt.borrow_mut().poll(
+            |client, topic, message, properties| {
+                let (response, settings_update) =
+                    self.route_message(topic, message, properties);
+                client
+                    .publish(
+                        response.topic,
+                        &response.message,
+                        minimq::QoS::AtMostOnce,
```
jordens, line 141: Let's add a `TODO` that this should be `AtLeastOnce` or even `ExactlyOnce` (typo in minimq).

src/net/mqtt_interface.rs
```diff
@@ -0,0 +1,235 @@
...
+        let mut telemetry_topic: String<consts::U128> = String::new();
+        write!(&mut telemetry_topic, "{}/telemetry", prefix).unwrap();
+
+        let mut response_topic: String<consts::U128> = String::new();
+        write!(&mut response_topic, "{}/log", prefix).unwrap();
+
+        Self {
+            mqtt: RefCell::new(mqtt_client),
+            settings: RefCell::new(S::default()),
+            id: String::from(prefix),
```
jordens, line 60: Why the conversion and not just `to_owned`? You need `str` afterwards anyway.

jordens (changes requested):

src/net/messages.rs
```diff
@@ -0,0 +1,124 @@
...
+use core::fmt::Write;
+
+#[derive(Debug, Copy, Clone)]
+pub enum SettingsResponseCode {
+    NoError = 0,
+    NoTopic = 1,
+    InvalidPrefix = 2,
+    UnknownTopic = 3,
+    UpdateFailure = 4,
+}
```
irwineffect, line 13: Perhaps we should define these codes externally so that we can have one shared definition between this module and the receiver that interprets these codes? Also it looks like miniconf.py (right now) doesn't care about the value of the code except that it is non-zero, is there a need for these other values?

src/net/messages.rs
```diff
@@ -0,0 +1,124 @@
...
+            msg,
+            code: SettingsResponseCode::UpdateFailure as u8,
+        }
+    }
+
+    /// Construct a response from a custom response code.
+    ///
+    /// # Args
+    /// * `code` - The response code to provide.
+    pub fn code(code: SettingsResponseCode) -> Self {
```
irwineffect, line 113: Should we implement `From<SettingsResponseCode>` instead of this function?

src/net/messages.rs
```diff
@@ -0,0 +1,124 @@
...
+        let mut msg: String<consts::U64> = String::new();
+
+        // Note(unwrap): All code debug names shall fit in the 64 byte string.
+        write!(&mut msg, "{:?}", code).unwrap();
```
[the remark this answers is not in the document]
ryan-summers: Correct, it won't be machine parseable, but that's not the point of this one. For this specifically, we need to provide feedback to the user as to what kind of miniconf error occurred so they can figure out how they may have malformed their command. This is the only case that I believe should stay, otherwise `miniconf.py` will just say "Your command didn't work. Fix it", instead of something like "Command Error: UnknownSettingPath"

src/net/messages.rs
```diff
@@ -0,0 +1,124 @@
...
+pub enum SettingsResponseCode {
+    NoError = 0,
+    NoTopic = 1,
+    InvalidPrefix = 2,
+    UnknownTopic = 3,
+    UpdateFailure = 4,
+}
+
+/// Represents a generic MQTT message.
+pub struct MqttMessage<'a> {
```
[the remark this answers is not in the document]
ryan-summers: As per discussion, for now, this will live in stabilizer, but we will likely copy this all over to `miniconf` at a later date behind an MQTT-feature flag.

src/net/mqtt_interface.rs
ryan-summers: As per discussion, this will move to `miniconf` in the future as a generic API behind an mqtt feature-flag.

src/net/mqtt_interface.rs
ryan-summers: heapless::String does not implement `ToOwned`

src/net/mqtt_interface.rs
```diff
@@ -0,0 +1,235 @@
...
+                            Err(error) => {
+                                SettingsResponse::update_failure(path, error)
+                            }
+                        }
+                    }
+                    Some(_) => SettingsResponse::code(
+                        SettingsResponseCode::UnknownTopic,
+                    ),
+                    _ => SettingsResponse::code(SettingsResponseCode::NoTopic),
+                }
```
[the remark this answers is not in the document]
ryan-summers: As per discussion with @jordens yesterday, routing is being eliminated entirely - we're going to instead just instantiate MQTT clients for each task we have. This has the advantage of offloading all routing-related operations to the MQTT broker, so we can keep it entirely out of the application layer.

src/net/mqtt_interface.rs
ryan-summers: The phy API is part of the stm32h7xx-hal, so we don't control what it returns

miniconf.py
```diff
@@ -51,7 +52,8 @@ def _handle_response(self, _client, topic, payload, *_args, **_kwargs):
             logger.warning('Unexpected response on topic: %s', topic)
             return
 
-        self.inflight[topic].set_result(payload.decode('ascii'))
+        response = json.loads(payload)
+        self.inflight[topic].set_result((response['code'], response['msg']))
```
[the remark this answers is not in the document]
ryan-summers: Because tuples are generally much more accessible as return values than dictionaries. If you use a dictionary as a return argument, the end user then needs to know the specific keys to use to access the underlying data. However, if we use a tuple, it allows them to directly unpack like so:
```python
return_code, message = await future
```

Without unpacking into a tuple, it would have to be (note that the user now needs to know about the specific keys):
````python
response = await future
code = response['code']
message = response['msg']
````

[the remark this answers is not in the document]
Comment (ryan-summers) [reply miniconf.py:56]: In fact, not unpacking it at the API layer actually _forces_ us to document the response structure and make it backward compatible. With as-implemented, we can change the entire reporting structure of miniconf without worrying about breaking the external API.

File: src/net/mqtt_interface.rs (hunk 4 above)
Comment (ryan-summers) [reply src/net/mqtt_interface.rs:121]: Added a note here that we log subscription failures but do not act on them currently. I'm of the opinion there's not much more we can do, as there's either a firmware bug or we aren't compatible with the broker. Feel free to comment otherwise.

**Comment by ryan-summers** [comment]
bors try

File: miniconf.py [hunk 13]
```diff
@@ -32,27 +32,53 @@ def __init__(self, client, prefix):
...
+        try:
+            if correlation_data['id'] != self.uuid.hex:
+                self.logger.info('Ignoring correlation data for different ID')
+                return
+            pid = correlation_data['pid']
+        except KeyError:
+            self.logger.warning('Ignoring unknown correlation data: %s', correlation_data)
+            return
+
+        if pid not in self.inflight:
```
jordens, line 71: I'm actually in favor of not checking this and instead taking the `KeyError` that'll be thrown automatically.
Just let the error bubble up.

miniconf.py
```diff
@@ -32,27 +32,53 @@ def __init__(self, client, prefix):
...
+            self.logger.warning('Ignoring message with invalid correlation data')
+            return
+
+        # Validate the correlation data.
+        try:
+            if correlation_data['id'] != self.uuid.hex:
+                self.logger.info('Ignoring correlation data for different ID')
+                return
+            pid = correlation_data['pid']
+        except KeyError:
```
jordens, line 67: I'd just let the error bubble up.

miniconf.py
```diff
@@ -32,27 +32,53 @@ def __init__(self, client, prefix):
...
+            _qos: The quality-of-service level of the received packet
+            properties: A dictionary of properties associated with the message.
         """
-        if topic not in self.inflight:
-            # TODO use correlation_data to distinguish clients and requests
-            logger.warning('Unexpected response on topic: %s', topic)
+        # Extract corrleation data from the properties
+        try:
+            correlation_data = json.loads(properties['correlation_data'][0].decode('ascii'))
+        except (json.decoder.JSONDecodeError, KeyError):
```
jordens, line 57: I'd just let the error bubble up.

miniconf.py
```diff
@@ -32,27 +32,53 @@ def __init__(self, client, prefix):
...
+            self.logger.warning('Unexpected pid: %s', pid)
             return
 
-        self.inflight[topic].set_result(payload.decode('ascii'))
-        del self.inflight[topic]
+        try:
+            response = json.loads(payload)
+            self.inflight[pid].set_result((response['code'], response['msg']))
+            del self.inflight[pid]
+        except json.decoder.JSONDecodeError:
```
jordens, line 79: I'd just let the error bubble up.

miniconf.py
```diff
@@ -62,29 +88,39 @@ async def command(self, path, value):
             value: The value to write to the path.
 
         Returns:
-            The received response to the command.
+            (code, msg) tuple as a response to the command. `code` is zero for success and `msg` is
```
jordens, line 91: Unfortunately I can't continue commenting on the discussion because it's "outdated".
But I disagree. You say that returning a dict would leak implementation details. I disagree with that perspective. Instead:
* The implementation detail is not a freedom anymore since you now commit to returning certain data.
* Returning a tuple makes this not self-documenting anymore (while the response itself is self-documenting).
* It also prevents any extension of the response data (like for example returning the actual quantized or clipped values used).

The argument that unpacking is sometimes less code is valid. But it's not so important IMO: In our case here with f-strings, it's actually more code to unpack.

miniconf.py
```diff
@@ -62,29 +88,39 @@ async def command(self, path, value):
...
         """
         setting_topic = f'{self.prefix}/settings/{path}'
         response_topic = f'{self.prefix}/response/{path}'
-        if response_topic in self.inflight:
-            # TODO use correlation_data to distinguish clients and requests
-            raise NotImplementedError(
-                    'Only one in-flight message per topic is supported')
+
+        # Assign a unique identifier to this update request.
+        pid = self.request_id
```
jordens, line 98: `PID` is confusing. Why not stick with `request_id`?

miniconf.py
```diff
@@ -62,29 +88,39 @@ async def command(self, path, value):
...
         value = json.dumps(value)
-        logger.info('Sending %s to "%s"', value, setting_topic)
+        self.logger.info('Sending %s to "%s"', value, setting_topic)
         fut = asyncio.get_running_loop().create_future()
-        self.inflight[response_topic] = fut
+
+        self.inflight[pid] = fut
         self.client.publish(setting_topic, payload=value, qos=0, retain=True,
-                            response_topic=response_topic)
+                            response_topic=response_topic,
```
jordens, line 113: I wonder if it might be better to dispense with the `uuid` in the correlation data and instead have one response topic per client (`{self.prefix}/response/{uuid}` without `path`). I.e. let the broker do the routing again.
Not critical here but maybe worth a simplification/improvement issue.

miniconf.py
```diff
@@ -32,27 +32,53 @@ def __init__(self, client, prefix):
...
             prefix: The MQTT toptic prefix of the device to control.
         """
+        self.uuid = uuid.uuid1()
+        self.request_id = 0
         self.client = client
         self.prefix = prefix
         self.inflight = {}
         self.client.on_message = self._handle_response
         self.client.subscribe(f'{prefix}/response/#')
+        self.logger = logging.getLogger(__name__)
```
jordens, line 42: Could you elaborate why you don't like the module-level logger? It is [standard practice](https://github.com/python/cpython/blob/main/Lib/asyncio/log.py).
This doesn't help with distinguishing `Miniconf` instances.

jordens (changes requested):

miniconf.py
ryan-summers: I moved the logger into `Miniconf` because variables outside of objects are globals. Since we really didn't need the logger to be global (as it's only used within `Miniconf`), I moved it into `Miniconf` to satisfy PyLint. Other than that, there's no particular reason and I don't believe it has a practical implication on end-users.

miniconf.py
dnadlinger: Yes, this is imho the design intent. I've added correlation data precisely to implement it like this (choose topic based on UUID, and then use correlation data to disambiguate between multiple in-flight requests).

src/net/mqtt_interface.rs
ryan-summers: This sounds like we're trying to address a problem that we expect to have in the future. I would say that let's cross that bridge when we come to it - as of right now, we don't publish the active settings, so this is not an issue.

jordens: Agreed!

miniconf.py
jordens: But there are so many global variables. Everything from and inside imported modules to classes and their attributes and functions. A logger to me is **the** paradigmatic global resource in the same sense that `print()` accesses a global resource.
I would buy it if there is a custom formatter that adds e.g. `id(self)` to the logging messages. But that doesn't seem needed.

I have to say I'm disappointed by PyLint here ;)

miniconf.py
```diff
@@ -52,32 +52,13 @@ def _handle_response(self, _client, _topic, payload, _qos, properties):
...
-            self.inflight[pid].set_result((response['code'], response['msg']))
-            del self.inflight[pid]
-        except json.decoder.JSONDecodeError:
-            self.logger.warning('Invalid response format: %s', payload)
+        correlation_data = json.loads(properties['correlation_data'][0].decode('ascii'))
+
+        # Get the request ID from the correlation data
+        request_id = correlation_data['request_id']
+
+        self.inflight[request_id].set_result(json.loads(payload))
```
jordens, line 60: Thanks ;)
This then needs updating of the `command()` docstring.

miniconf.py
jordens: Maybe `flake8` is a bit better.
A slightly more technical reason for a single global logger is resource usage: the case where I have 100 Stabilizer in the lab and I control them with one `Miniconf(prefix, broker)` each. I don't need 100 `Logger` for that.

jordens (approved):
Looks good!

ryan-summers:
bors merge

Issue #86, opened 2020-04-01
Scan mode


dtcallcock:
It would be good to be able to repeatedly scan an output and then observe the response on the corresponding input (eg. a cavity resonance or an atomic feature) in real time on a PC.  This sounds fairly trivial to implement on the Stabilizer side, however a good interface to the PC may need some thought and I mention it now as it's relevant to #54.

jordens:
I don't see this as relevant for the choice of command interface infrastructure (#54). It's more part of some gui functionality.

dtcallcock:
> It's more part of some gui functionality.

Are you saying you would implement this entirely in the GUI (ie. have the GUI set the DAC, then request the ADC value for each point in the scan)?  Wouldn't network latency be an issue?  It seems like it would be more efficient to have some simple firmware code that generates the scan points on the fly based on a set of parameters (eg. start, end, step size) and then returns the whole scan.

dtcallcock:
We'd like to use stabilizer in anger but we really need this to do it. Would you be willing to prepare a contract to expedite this feature?

jordens:
Sure.

jordens:
* Implement two independent fully parametric triangular signal
generators (one for each DAC channel).
* They have runtime-adjustable frequency, amplitude through the
networking interface.
* Offset either implemented through the existing IIR filter offset or
as a separate offset in the triangular signal generators.
* These triangular signal generators may be unified with the lock-in
signal generator (thereby inheriting it's PLL locking features and
configuration options).
* There is a reference implementation to configure them (like
`stabilizer.py` currently).
* To use them (in typical spectroscopy alignment mode) you'd disable
the IIR filters (setting their gains to zero or just disabling the
integrator) and then configure/enable the triangular generators.
* If the offset is handled by the IIR filter, you'd get bump-less
transfer (when disabling the sweep generator and enabling the IIR
filter) for free.
* If the signal generation ends up being costly (in terms of CPU
time), we'll have to reduce the ADC/DAC sample rate or find other
tweaks to compensate for that.
* Other user interfaces (graphical or not), digital trigger
output/input for the signal generator, streaming of ADC/DAC/sweep data
for GUIs/network analysis etc. would be separate features.

dtcallcock:
> They have runtime-adjustable frequency, amplitude

[...]

PR #352, opened 2021-05-06
Refactoring lockin binaries

ryan-summers:
This PR merges `lockin-external` and `lockin-internal` into a single binary.

This has also refactored the output configuration to make it more configurable. This was needed to work well with the reference signal generation, as that output should necessarily just override a different setting.


src/bin/lockin.rs
```diff
@@ -18,16 +18,34 @@ use stabilizer::hardware::{
...
+const DAC_SEQUENCE: [i16; design_parameters::SAMPLE_BUFFER_SIZE] =
+    [ONE, SQRT2, 0, -SQRT2, -ONE, -SQRT2, 0, SQRT2];
+
 #[derive(Copy, Clone, Debug, Deserialize, Miniconf)]
 enum Conf {
-    PowerPhase,
+    Power,
+    Phase,
+    PllFrequency,
     FrequencyDiscriminator,
```
ryan-summers, line 33: @jordens Do you have better names than `PllFrequency` and `FrequencyDescriminator`? I don't know if the names I selected make sense for these measurements (as I'm not quite sure what these values are).

jordens: `InPhase` (`.re`), `Quadrature` (`.im`), `Magnitude` (`abs_sqr()`), `LogPower` (`log2()`), `Phase` (`arg()`), `PllFrequency` (`pll_frequency`), `Modulation` (`DAC_SEQUENCE`) should be the ones. Each channel should support all of them, independently.

ryan-summers: Alright, should now be cleaned up. Thanks for the input!

ryan-summers:
Alright, testing on Stabilizer with an oscilloscope shows that both channels are independently configurable and `miniconf.py` shows them swapping over measurements as requested.

@jordens This is ready for review whenever you get a chance.

src/bin/lockin.rs
```diff
@@ -137,29 +157,57 @@ const APP: () = {
...
-            .lockin_phase
-            .wrapping_add(pll_phase.wrapping_mul(settings.lockin_harmonic));
+                // Demodulation LO phase offset
+                let phase_offset: i32 = 1 << 30;
+
+                let sample_frequency = (pll_frequency as i32)
+                    .wrapping_mul(settings.lockin_harmonic);
+                let sample_phase = phase_offset.wrapping_add(
+                    pll_phase.wrapping_mul(settings.lockin_harmonic),
+                );
```
jordens, line 206: This should be unified with L182-L188, use `settings.phase_offset` as well, and pulled out of the `match`. But I can do that later.

src/bin/lockin.rs
```diff
@@ -137,29 +157,57 @@ const APP: () = {
...
-            settings.pll_tc[0],
-            settings.pll_tc[1],
-        );
+        let mut pll_frequency = 0;
+
+        let (sample_phase, sample_frequency) = match settings.lockin_mode {
+            LockinMode::External => {
+                let timestamp =
+                    c.resources.timestamper.latest_timestamp().unwrap_or(None); // Ignore data from timer capture overflows.
+                let (pll_phase, frequency) = c.resources.pll.update(
```
jordens, line 174: Probably better to call these `reference_phase` and `reference_frequency` and let both arms return `reference_frequency` (constant in the `Internal` case) that's then used as `conf::ReferenceFrequency` output.

src/bin/lockin.rs
```diff
@@ -18,16 +18,34 @@ use stabilizer::hardware::{
 use miniconf::Miniconf;
 use stabilizer::net::{Action, MqttInterface};
 
+// A constant sinusoid to send on the DAC output.
+// Full-scale gives a +/- 10.24V amplitude waveform. Scale it down to give +/- 1V.
+const ONE: i16 = ((1.0 / 10.24) * i16::MAX as f32) as _;
+const SQRT2: i16 = (ONE as f32 * 0.707) as _;
+const DAC_SEQUENCE: [i16; design_parameters::SAMPLE_BUFFER_SIZE] =
+    [ONE, SQRT2, 0, -SQRT2, -ONE, -SQRT2, 0, SQRT2];
```
jordens, line 26: Once #86 lands, this should be refactored to use that same signal generator.

jordens (approved):

jordens:
bors merge

ryan-summers:
bors cancel

ryan-summers:
bors merge

ryan-summers:
bors cancel

ryan-summers:
bors merge

PR #355, opened 2021-05-10
dsp/lowpass,lockin: const generics

jordens:


ryan-summers (approved):
Man, this is such a refreshing change. I'll work towards removing `generic-array` from our dependencies as well.

PR #358, opened 2021-05-13
RTT logging, probe-run

jordens:
- use rtt logging
- remove rtt feature
- update hitl script
- README: streamline docs a bit, add probe-run
- CHANGELOG: update
- README: document lockin, cleanup instructions, use probe-run
- set target-cpu=cortex-m7, MSRV 1.52.0, use probe-run

### DIFF
--- a/.cargo/config
+++ b/.cargo/config
@@ -1,16 +1,10 @@
 [target.'cfg(all(target_arch = "arm", target_os = "none"))']
-# runner = "probe-run --chip STM32H743ZITx --speed 30000"
-runner = "gdb-multiarch -q -x openocd.gdb"
+runner = "probe-run --chip STM32H743ZITx --speed 30000"
+# runner = "gdb-multiarch -q -x openocd.gdb"
 rustflags = [
     "-C", "link-arg=-Tlink.x",
-# The target (below) defaults to cortex-m4
-# There currently are two different options to go beyond that:
-# 1. cortex-m7 has the right flags and instructions (FPU) but no instruction schedule yet
-#    "-C", "target-cpu=cortex-m7",
-# 2. cortex-m4 with the additional fpv5 instructions and a potentially
-# better-than-nothing instruction schedule
-    "-C", "target-feature=+fp-armv8d16",
-# When combined they are equivalent to (1) alone
+    "-C", "link-arg=--nmagic",
+    "-C", "target-cpu=cortex-m7",
 ]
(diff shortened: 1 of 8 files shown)


jordens:
bors try

jordens:
bors try

jordens:
bors try

Bors doesn't look very cooperative today. HITL results are at https://github.com/quartiq/stabilizer/runs/2574636512?check_suite_focus=true

.cargo/config
```diff
@@ -1,15 +1,10 @@
 [target.'cfg(all(target_arch = "arm", target_os = "none"))']
-runner = "gdb-multiarch -q -x openocd.gdb"
+runner = "probe-run --chip STM32H743ZITx --speed 30000"
```
ryan-summers, line 2: Does probe-run spin up a GDB server as well? Or is this just a command to program the chip?

ryan-summers: In general, I'd prefer if we kept around an easy way to spin up GDB, which was previously probe-run. The command-line way to flash was previously just `cargo embed --bin dual-iir`, and `cargo run --bin dual-iir` was meant for interactive debug

jordens: There are still two ways to flash and while still getting gdb access as before. Just the command to start gdb changes.

* Start `cargo embed` with options to provide the gdb stub (https://github.com/probe-rs/cargo-embed/blob/master/src/config/default.toml#L71) or start `openocd` as before.
* `cargo build --release --bin app && gdb-multiarch -q -x openocd.gdb path/to/app` to build and start gdb.

You can still use `cargo embed` to flash in the same way. That doesn't change. The command to start GDB becomes a bit longer.
The tooling for GDB+openocd is (a) significantly more complex to install (b) we don't test it, (c) doesn't provide usable logging, and (d) it isn't used as frequently compared to "just" flashing. Therefore I'd rather have a low entry barrier, streamlined runner that gives me backtraces and convenient logging for debugging.

ryan-summers: Ah, probe-run gives a logging interface. That's no issue - I can get gdb set up locally or see if I can make a custom cargo profile or something.

ryan-summers (approved):

.cargo/config
jordens: A cargo profile might be nice for gdb.

jordens:
bors merge

PR #362, opened 2021-05-17
rj/visibility cleanup

jordens:
- dsp: reduce num dependency
- clean up item visibility


jordens:
bors try

src/bin/dual-iir.rs
```diff
@@ -2,19 +2,18 @@
...
-use hardware::{
-    Adc0Input, Adc1Input, AdcCode, AfeGain, Dac0Output, Dac1Output, DacCode,
-    DigitalInput0, DigitalInput1, InputPin, SystemTimer, AFE0, AFE1,
+use stabilizer::{
+    hardware::{
+        hal, setup, Adc0Input, Adc1Input, AdcCode, AfeGain, Dac0Output,
+        Dac1Output, DacCode, DigitalInput0, DigitalInput1, InputPin,
+        SystemTimer, AFE0, AFE1,
+    },
+    net::{Miniconf, NetworkUsers, Telemetry, TelemetryBuffer, UpdateState},
```
ryan-summers, line 14: In general, I'm more of a fan of directly importing things where used from the direct source. It makes it easier to trace where things are coming from.

If you export `Miniconf` in `stabilizer::net`, then someone has to look at `stabilizer::net` first to see that `Miniconf` comes from `miniconf::Miniconf` and isn't some alias of it.

However, when we import `miniconf::Miniconf` directly where used, it's quick and easy to reference exactly where Miniconf is being derived from.

In general, my preference is to import things from the source wherever possible to alleviate the mental burden of going through the manual lookup

src/bin/dual-iir.rs
```diff
@@ -69,7 +68,7 @@ const APP: () = {
     #[init(spawn=[telemetry, settings_update])]
     fn init(c: init::Context) -> init::LateResources {
         // Configure the microcontroller
-        let (mut stabilizer, _pounder) = hardware::setup(c.core, c.device);
+        let (mut stabilizer, _pounder) = setup(c.core, c.device);
```
ryan-summers, line 71: Honestly, I think `hardware::setup` is more descriptive than just `setup()`, but that's a personal preference. I just think the `hardware` portion makes it clear what type of setup is happening.

src/hardware/mod.rs
```diff
@@ -16,13 +16,13 @@ pub mod pounder;
...
-pub use digital_input_stamper::InputStamper;
-pub use pounder::DdsOutput;
-pub use system_timer::SystemTimer;
+pub use adc::*;
+pub use afe::{Gain as AfeGain, *};
+pub use cycle_counter::*;
+pub use dac::*;
+pub use digital_input_stamper::*;
+pub use pounder::*;
+pub use system_timer::*;
```
ryan-summers, line 25: [I am very opposed to this type of vague importing](https://stackoverflow.com/questions/2386714/why-is-import-bad). We should only be importing explicitly here. 

These modules all have internal types that we may not necessarily want or need to export publicly. It also makes it _extremely_ opaque to know what actually exists in `stabilizer::hardware`, since now you have to wade through all of these files to figure out which one contains the reference for `DdsOutput`.

src/hardware/pounder/mod.rs
```diff
@@ -1,21 +1,19 @@
...
-use super::hal;
-
-use attenuators::AttenuatorInterface;
-use rf_power::PowerMeasurementInterface;
-
-use embedded_hal::{adc::OneShot, blocking::spi::Transfer};
+pub use attenuators::*;
+pub use dds_output::*;
+pub use hrtimer::{Channel as HRTimerChannel, *};
+pub use rf_power::*;
```
ryan-summers, line 16: Same as above. Only import what we need to (fine with keeping the pub mods as private and re-exporting as implemented though).

src/net/mod.rs
```diff
@@ -20,10 +19,10 @@ mod telemetry;
...
 use messages::{MqttMessage, SettingsResponse};
 
-pub use miniconf_client::MiniconfClient;
-pub use network_processor::NetworkProcessor;
-pub use shared::NetworkManager;
-pub use telemetry::{Telemetry, TelemetryBuffer, TelemetryClient};
+pub use miniconf_client::*;
+pub use network_processor::*;
+pub use shared::*;
+pub use telemetry::*;
```
ryan-summers, line 25: Again, see above

ryan-summers (changes requested):

src/hardware/mod.rs
jordens: The link you give is about Python. This isn't Python.
With Rust there is no problem or stylistic downside to wildcard imports. They are not vague or fragile at all. They are used frequently all over the place. They are not in any way deprecated anywhere, not by clippy, not by any best practices, not by reality. Your editor/IDE will easily be able to determine where an import comes from.
Conversely the explicit importing doesn't work in practice as you can see here. Things that are `pub` at their implementation site and meant to be exported, are forgotten either by forgetting to `pub mod` the mod or by forgetting to `pub use` the item. The fact that explicit re-exporting is way to fragile and hard to maintain is the reasong for this PR.

src/bin/dual-iir.rs
jordens: I'm ok with keeping the namespace hierarchy and not `pub use` re-exporting.
But that isn't the style you have been following so far at all. Currently it's a happy mixture of `pub use` and `pub mod`.
It would remove the wildcard `pub use` re-exports but makes the actual use in the application quite a bit harder to read.

src/bin/dual-iir.rs
jordens: Maybe but why is function called `setup()` while the module (made up to 90% by `setup()`) is called `configuration`? Shouldn't that be `configure()` then?

src/bin/dual-iir.rs
ryan-summers: I don't think I'm completely following your logic here. This is purely in regard to keeping crate-dependency-level `use` statements where-used.

src/bin/dual-iir.rs
ryan-summers: I'm fine with naming either both `setup.rs` or `configuration.rs`, but I still think it would be useful to keep the `hardware` prefix here personally - could be convinced otherwise if we renamed to something like `hardware_setup(c.core, c.device)`. Anything else just feels far too generic to me.

src/hardware/mod.rs
ryan-summers: I know this isn't python, but the concepts are similar.

My main point here is that traversing exactly where something is coming from is hard while using wild-carding. While yes, an IDE/editor can solve this, it's worth noting that I still personally don't use an editor, so would have to grep to find where things are defined. With renaming (e.g. via use as), it gets even harder.

src/bin/dual-iir.rs
jordens: There are two topics of namespace management here. One is the re-exporting of to remove namespace hierarchy levels. E.g. `TelemetryBuffer`. This is `pub mod` to export the module vs `pub use` to re-export the item.
The other is re-exporting interfaces/traits from other crates. If we don't do that, it becomes extremely annoying to copy over and match exact dependencies in e.g. `fls`. Having to do so is pointless because the correct dependency is right there in `stabilizer`. In the case of macros (Miniconf, Serialize, Deserialize) it unfortunately doesn't work very well but for traits it's very convenient to be able to get them re-exported from `stabilizer` and very inconvenient to have to cross-depend on a crate and manage+track that dependency. Re-exporting also seems to be somewhat common, especially if the downstream crate will generally only need the trait because and through the intermediate crate.

src/hardware/mod.rs
jordens: > Mind elaborating on this point? I'm not sure I agree with you.

For example, it's impossible to fully use pounder in `fls` currently because multiple items (e.g. `PowerMeasurementInterface`) have been forgotten from the explicit re-export list even though those items are nicely `pub` at their origin (and obviously are required for public consumption). And that's bound to happen over and over again basically by design since we have external apps (like `fls`) covering different parts of the API than the crate-local apps.

src/bin/dual-iir.rs
ryan-summers: Given that we're treating stabilizer as a library, I can get on board with this choice. Thanks for taking the time to explain your reasoning here.

src/hardware/mod.rs
ryan-summers: When thinking about using `stabilizer` as a library, I'm getting more on-board with this. I think an important question to ask is whether or not we should be flattening the names. We could always just `pub use` all the modules and let the end user flatten things. This would help to address my previous issue of traceability, but keep the auto-public export functionality too. Thoughts?

src/bin/dual-iir.rs
jordens: Context:
[serde-json-core](https://github.com/rust-embedded-community/serde-json-core/blob/986dcab749210bd1a4b25f6e75b1e5e549428fa4/src/lib.rs#L73) re-exports `heapless`. Maybe `miniconf` should re-export `serde-json-core` as well.

src/hardware/mod.rs
jordens: Ok. Changed to all `pub use` and no flattening.

src/bin/dual-iir.rs
jordens: changed to `stabilizer::hardware::setup::setup()`.

ryan-summers (approved):

jordens:
bors merge

PR #365, opened 2021-05-21
Adding basic task timing telemetry

ryan-summers:
This PR addresses #166 by adding some basic timers to the various tasks.

This is a minimalistic implementation so that I could get a guage as to how long tasks were taking, so it's not really considered complete.

Ideally:
* This could be integrated into RTIC
* Lower the serialization overhead required here
* We utilize the `SystemTimer` to calculate average load of each task

### DIFF
--- a/src/bin/dual-iir.rs
+++ b/src/bin/dual-iir.rs
@@ -5,15 +5,23 @@
 use stabilizer::{hardware, net};
 
 use miniconf::Miniconf;
-use serde::Deserialize;
+use serde::{Serialize, Deserialize};
 
 use dsp::iir;
 use hardware::{
     Adc0Input, Adc1Input, AdcCode, AfeGain, Dac0Output, Dac1Output, DacCode,
-    DigitalInput0, DigitalInput1, InputPin, SystemTimer, AFE0, AFE1,
+    DigitalInput0, DigitalInput1, InputPin, SystemTimer, AFE0, AFE1, TaskTimer, TaskInfo,
 };
 
 use net::{NetworkUsers, Telemetry, TelemetryBuffer, UpdateState};
+
+#[derive(Serialize)]
+pub struct TimedTelemetry {
+    basic: Telemetry,
+    tlm: TaskInfo,
+    dsp: TaskInfo,
+    settings: TaskInfo,
+}
 
 const SCALE: f32 = i16::MAX as _;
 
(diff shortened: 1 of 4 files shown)


jordens:
* It would be very useful to not bind this tooling to tasks but also to be able to time certain parts of tasks or to measure the idle time around `wfi`. E.g. I hacked that up to distinguish DMA overhead from actual DSP processing in `process()`.
* Publishing these with telemetry doesn't seem too useful for me for several reasons: The data is meaningful to the application developer and less to the application user. The measurement results are somewhat unlikely to change during use. They are basically representing software design results.
* Useful data for decision making is not so much the mean but the min and max.
* I'd stick to cycle-based data and not to SI seconds. Then it's independent of the CPU speed (removes code and is meaningful).
* I'd try to do a sampling monitor: have a max priority task that can interrupt everything and then merely determines (samples) (either from NVIC or maybe from stack) which other running task just interrupted. From that it can easily and with almost no overhead build statistics of all tasks running and their relative load as a histogram. Additionally you can dial down the overhead by changing the firing rate of that sampling tasks.

ryan-summers:
Just to provide context - this is some development tooling I created that I figured may address points of #166, so I put it up here as a draft to start up some conversation as to how #166 should be addressed.

I agree with many of your points here, so I'll close this, as I don't think it makes sense to merge these changes in their current form.

jordens:
For reference and in case I misplace this, here's the hack that I've bee carrying around.

```patch

 src/bin/dual-iir.rs | 30 ++++++++++++++++++++++++++++++
 1 file changed, 30 insertions(+)

diff --git a/src/bin/dual-iir.rs b/src/bin/dual-iir.rs
index dec0697..7ff4563 100644
--- a/src/bin/dual-iir.rs
+++ b/src/bin/dual-iir.rs
@@ -52,6 +52,32 @@ impl Default for Settings {
     }
 }

+struct TickTock {
+    last: u32,
+    n: i32,
+}
+
+impl TickTock {
+    const fn new(n: i32) -> Self {
+        Self { last: 0, n: n }
+    }
+
+    fn tick(&mut self) -> u32 {
+        let dwt = unsafe { &*stm32h7xx_hal::stm32::DWT::ptr() };
+        self.last = dwt.cyccnt.read();
+        self.last
+    }
+
+    fn tock(&mut self) {
+        if self.n > 0 {
+            let last = self.last;
+            let diff = self.tick().wrapping_sub(last);
+            log::info!("tock: {}", diff);
+            self.n -= 1;
+        }
+    }
+}
+
 #[rtic::app(device = stm32h7xx_hal::stm32, peripherals = true, monotonic = stabilizer::hardware::SystemTimer)]
 const APP: () = {
     struct Resources {
@@ -123,6 +149,9 @@ const APP: () = {
     /// the same time bounds, meeting one also means the other is also met.
     #[task(binds=DMA1_STR4, resources=[adcs, digital_inputs, dacs, iir_state, settings, telemetry], priority=2)]
     fn process(mut c: process::Context) {
+        static mut T: TickTock = TickTock::new(10);
+        T.tick();
+
         let adc0 = &mut c.resources.adcs.0;
         let adc1 = &mut c.resources.adcs.1;
         let dac0 = &mut c.resources.dacs.0;
@@ -189,6 +218,7 @@ const APP: () = {
                 })
             })
         });
+        T.tock();
     }

     #[idle(resources=[network], spawn=[settings_update])]
```

PR #367, opened 2021-05-24
dma: don't swap buffers

### LINKED ISSUE #260: Remove Pounder timestamping DMA
Since there will be exactly one timestamp per batch and not more (also for processing constraints), we can do without DMA. DMA buffer juggling is at likely as costly as accessing the peripheral directly.
Low priority, to be tackled when this is tested again.

jordens:
* This uses a new closure-based method to the DMA HAL implementation which
  gives access to the inactive buffer directly.
* It removes changing addresses, the third buffer for DBM, the inactive
  address poisoning, and allows the cancellation of the redundant repeat
  memory barriers and compiler fences.
* This is now around 20 instructions per buffer down from about 100 cycles
  before.
* Also introduces a new `SampleBuffer` type alias.
* The required unpacking of the resources structure is a bit annoying
  but could probably abstraced away.
* Reduced pounder capture rate to the batch rate using the prescaler.
* Removes the Pounder Timestamper DMA (close #260)

TODO:

* [x] Tested that dual-iir still works
* [x] Tested that DMA overflows are signaled as panics (batch size 1 at full rate)
* [x] Adapt `lockin`
* [x] Tested on FLS without pounder timestamp DMA.

### DIFF
--- a/src/hardware/pounder/dds_output.rs
+++ b/src/hardware/pounder/dds_output.rs
@@ -52,9 +52,11 @@
 ///! compile-time-known register update sequence needed for the application, the serialization
 ///! process can be done once and then register values can be written into a pre-computed serialized
 ///! buffer to avoid the software overhead of much of the serialization process.
+use log::warn;
+use stm32h7xx_hal as hal;
+
 use super::{hrtimer::HighResTimerE, QspiInterface};
 use ad9959::{Channel, DdsConfig, ProfileSerializer};
-use stm32h7xx_hal as hal;
 
 /// The DDS profile update stream.
 pub struct DdsOutput {
--- a/src/hardware/pounder/timestamp.rs
+++ b/src/hardware/pounder/timestamp.rs
@@ -22,9 +22,9 @@
 ///! mode. As soon as the DMA transfer completes, the hardware automatically swaps over to a second
 ///! buffer to continue capturing. This alleviates timing sensitivities of the DMA transfer
 ///! schedule.
+use crate::hardware::{design_parameters, timers};
+use core::convert::TryFrom;
 use stm32h7xx_hal as hal;
-
-use crate::hardware::timers;
 
 /// Software unit to timestamp stabilizer ADC samples using an external pounder reference clock.
 pub struct Timestamper {
@@ -64,8 +64,16 @@ impl Timestamper {
         timestamp_timer.set_trigger_source(timers::TriggerSource::Trigger1);
 
         // The capture channel should capture whenever the trigger input occurs.
-        let input_capture = capture_channel
+        let mut input_capture = capture_channel
             .into_input_capture(timers::tim8::CaptureSource1::TRC);
+
+        // Capture at the batch period.
+        input_capture.configure_prescaler(
+            timers::Prescaler::try_from(
+                design_parameters::SAMPLE_BUFFER_SIZE_LOG2,
+            )
+            .unwrap(),
+        );
 
         Self {
             timer: timestamp_timer,
--- a/src/hardware/pounder/timestamp.rs
+++ b/src/hardware/pounder/timestamp.rs
@@ -24,12 +24,27 @@
 ///! schedule.
 use stm32h7xx_hal as hal;
 
-use crate::hardware::timers;
+use hal::dma::{dma::DmaConfig, PeripheralToMemory, Transfer};
+
+use crate::hardware::{design_parameters::SAMPLE_BUFFER_SIZE, timers};
+
+// Three buffers are required for double buffered mode - 2 are owned by the DMA stream and 1 is the
+// working data provided to the application. These buffers must exist in a DMA-accessible memory
+// region. Note that AXISRAM is not initialized on boot, so their initial contents are undefined.
+#[link_section = ".axisram.buffers"]
+static mut BUF: [[u16; SAMPLE_BUFFER_SIZE]; 3] = [[0; SAMPLE_BUFFER_SIZE]; 3];
 
 /// Software unit to timestamp stabilizer ADC samples using an external pounder reference clock.
 pub struct Timestamper {
+    next_buffer: Option<&'static mut [u16; SAMPLE_BUFFER_SIZE]>,
     timer: timers::PounderTimestampTimer,
-    capture_channel: timers::tim8::Channel1InputCapture,
+    transfer: Transfer<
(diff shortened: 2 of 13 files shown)


src/lib.rs
```diff
@@ -1,8 +1,18 @@
...
+/// Macro to reduce rightward drift when calling the same closure-based API
+/// on multiple structs simultaneously, e.g. when accessing DMA buffers.
+/// This could be improved a bit using the tuple-based style from `mutex-trait`.
+#[macro_export]
+macro_rules! flatten_closures {
+    ($fn:ident, $e:ident, $fun:block) => {
+        $e.$fn(|$e| $fun ).unwrap()
+    };
+    ($fn:ident, $e:ident, $($es:ident),+, $fun:block) => {
+        $e.$fn(|$e| flatten_closures!($fn, $($es),*, $fun)).unwrap()
```
ryan-summers, line 16: Have you done a forced failure test to ensure that the flag check unwrap actually fails if the DSP process is taking too long? It would be good to verify that this is still caught.

It also occurs to me that we can now ignore the ADC0/ADC1 FIFO ISRs because we're offloading that logic to this unwrap here.

src/hardware/pounder/dds_output.rs
```diff
@@ -52,9 +52,11 @@
 ///! compile-time-known register update sequence needed for the application, the serialization
 ///! process can be done once and then register values can be written into a pre-computed serialized
 ///! buffer to avoid the software overhead of much of the serialization process.
+use log::warn;
+use stm32h7xx_hal as hal;
```
ryan-summers, line 56: Any reason why this file was changed? Was it clippy-related?

src/hardware/timers.rs
```diff
@@ -353,6 +356,21 @@ macro_rules! timer_channels {
...
+                ///
+                /// # Args
+                /// * `psc` - Prescaler exponent.
+                #[allow(dead_code)]
+                pub fn configure_prescaler(&mut self, prescaler: super::Prescaler) {
+                    // Note(unsafe): This channel owns all access to the specific timer channel.
+                    // Only atomic operations on completed on the timer registers.
+                    let regs = unsafe { &*<$TY>::ptr() };
+                    // Note(unsafe): Enum values are all valid.
+                    #[allow(unused_unsafe)]
```
ryan-summers, line 370: Is the unsafe actually necessary here? If it is, then we shouldn't need to allow an unused unsafe? I'm confused

src/hardware/pounder/timestamp.rs
```diff
@@ -22,29 +22,14 @@
 ///! mode. As soon as the DMA transfer completes, the hardware automatically swaps over to a second
 ///! buffer to continue capturing. This alleviates timing sensitivities of the DMA transfer
```
ryan-summers, line 23: Docs need to be updated here

ryan-summers (changes requested):
Other comments:

We need to update the SPI ISRs in the main apps to either listen for SPI errors, or just remove them entirely. It's my understanding that these are now unused as they are covered by the DMA DBM mode and unwrapped DMA overflow error

src/lib.rs
ryan-summers: Follow-up: It looks like you indicate this in your original post. Closing as resolved.

src/hardware/timers.rs
jordens: Rust is both erroring if the unsafe is missing and warning because of an unnecessary unsafe if it's there. May be a rust bug. I couldn't figure out a way that would make it happy.

src/hardware/pounder/dds_output.rs
jordens: Yes. Because of the removed old style `extern crate log` and knock-on clippy complaints.

jordens:
> Other comments:
> 
> We need to update the SPI ISRs in the main apps to either listen for SPI errors, or just remove them entirely. It's my understanding that these are now unused as they are covered by the DMA DBM mode and unwrapped DMA overflow error

Yes. `process()` being slow and loosing the race to the DMA buffer should not cause those ISRs to fire anymore (instead panic as tested).
But we were and are still listening for errors from the SPI peripheral. I expect them to be triggered if the DMA peripheral is too slow serving them or badly configured. But I don't know how to trigger them. I've updated the panics to explicitly state that the error is from the SPI peripheral and not DBM. But let's keep them. Apart from the code duplication I don't think they hurt us.

src/hardware/timers.rs
jordens:
```
error[E0133]: call to unsafe function is unsafe and requires unsafe function or block
   --> src/hardware/timers.rs:372:25
    |
372 |                         w.[< ic $index psc >]().bits(prescaler as u8)});
    |                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ call to unsafe function
...
397 | timer_channels!(PounderTimestampTimer, TIM8, u16);
    | -------------------------------------------------- in this macro invocation
    |
    = note: consult the function's documentation for information on how to avoid undefined behavior
    = note: this error originates in a macro (in Nightly builds, run with -Z macro-backtrace for more info)
```

versus

```
warning: unnecessary `unsafe` block
   --> src/hardware/timers.rs:370:62
    |
370 |                     regs.[< $ccmrx _input >]().modify(|_, w| unsafe {
    |                                                              ^^^^^^ unnecessary `unsafe` block
...
396 | timer_channels!(PounderTimestampTimer, TIM8, u16);
    | -------------------------------------------------- in this macro invocation
    |
    = note: this warning originates in a macro (in Nightly builds, run with -Z macro-backtrace for more info)
```

jordens:
bors merge

ryan-summers:
> But we were and are still listening for errors from the SPI peripheral. I expect them to be triggered if the DMA peripheral is too slow serving them or badly configured. But I don't know how to trigger them. I've updated the panics to explicitly state that the error is from the SPI peripheral and not DBM. But let's keep them. Apart from the code duplication I don't think they hurt us.

While we can still listen for these errors, we don't have to bind an ISR handler in the app. If we don't bind an ISR, they will use the default hardfault handler, which would clean up the apps a bit. In any case, it's a cosmetic difference imo.

jordens:
Ack. I'll leave them in dual-iir as the prime debugging app and remove them from lockin and fls.

PR #371, opened 2021-05-28
rj/pounder fixes

jordens:
- pounder: simplify attenuator spi interface
- deps: use mcp23017 release
- attenuators: use robust latching sequence
- rf_power: fix measurement
- pounder io extender: hack around some bug
- pounder: fix attenuator indices (latch and shiftreg)
- pounder: enum for gpio ext pins
- ad9959: refactor pad()

One big issue was the apparently broken mcp23017 `digital_write`. I couldn't get it to work.
For now, and since (a) we aren't handling the other pins in any configurable way, and (b) the readback-style modification is slow let's just to whole-port writes of constants.

### DIFF
--- a/src/hardware/pounder/mod.rs
+++ b/src/hardware/pounder/mod.rs
@@ -44,6 +44,8 @@ pub enum Error {
     Adc,
 }
 
+/// The numerical value (discriminant) of the Channel enum is the index in the attenuator shift
+/// register as well as the attenuator latch enable signal index on the GPIO extender.
 #[derive(Debug, Copy, Clone)]
 #[allow(dead_code)]
 pub enum Channel {
--- a/src/hardware/pounder/mod.rs
+++ b/src/hardware/pounder/mod.rs
@@ -313,12 +313,7 @@ impl PounderDevices {
             .map_err(|_| Error::I2c)?;
         devices
             .mcp23017
-            .write_gpio(mcp23017::Port::GPIOB, 1 << 5)
-            .map_err(|_| Error::I2c)?;
-
-        devices
-            .mcp23017
-            .digital_write(EXT_CLK_SEL_PIN, false)
+            .write_gpio(mcp23017::Port::GPIOB, 0x2F)
             .map_err(|_| Error::I2c)?;
 
         Ok(devices)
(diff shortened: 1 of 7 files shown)


src/hardware/pounder/mod.rs
```diff
@@ -40,10 +47,10 @@ pub enum Error {
 #[derive(Debug, Copy, Clone)]
 #[allow(dead_code)]
 pub enum Channel {
-    In0,
-    In1,
-    Out0,
-    Out1,
+    In0 = 0,
```
ryan-summers, line 50: What do these values correspond to? Would be worth documenting. It's not clear to me from reading the changes

src/hardware/pounder/mod.rs
```diff
@@ -329,12 +331,11 @@ impl AttenuatorInterface for PounderDevices {
...
         self.mcp23017
-            .digital_write(ATT_RST_N_PIN, false)
+            .write_gpio(mcp23017::Port::GPIOB, 0x0f)
             .map_err(|_| Error::I2c)?;
-        // TODO: Measure the I2C transaction speed to the RST pin to ensure that the delay is
-        // sufficient. Document the delay here.
+        // Duration of one I2C transaction is sufficiently long.
         self.mcp23017
-            .digital_write(ATT_RST_N_PIN, true)
+            .write_gpio(mcp23017::Port::GPIOB, 0x2f)
```
ryan-summers, line 338: I'd much rather we may some define for these constants. It's not ergonomic to look back at the schematic to figure out what 0x2F is supposed to correlate to (same for 0x0F above).

ryan-summers (commented):
Some minor clean-up requests, otherwise LGTM

src/hardware/pounder/mod.rs
jordens: Let's leave that for later. Magic constants matches the existing previous style (see e.g. L312 and L316).
This is about getting it to work now, not making it perfect.

src/hardware/pounder/mod.rs
jordens: It was in the commit message to 3616f1fa. Will ad a comment.

jordens:
bors merge

Issue #150, opened 2020-10-13
streaming


jordens:
* Minimum overhead, timestamped raw binary sample burst packets
* Containing a configurable subset of DAC/ADC/DDS high bandwidth data or anything from the signal crossbar
* Maybe UDP
* Maybe target controlled and configured via MQTT
* Best effort task, should not harm the main processing task performance.

forked from #54 
details in #99 and #147

dnadlinger:
Some notes on a MVP streaming implementation (which I may or may not get around to actually doing):

 - Addressed use cases:
   - Interactive tuning of feedback parameters/inspection of lock quality (stream error signal/IIR output, display in UI, or acquire several seconds and compute FFT, …)
   - Bespoke one-off scripts running on a PC for prototyping/handling more complex use cases, e.g. automatic lock acquisition for a few-Hz laser system where Stabilizer is used for auxiliary loops
 - (Soft) requirements/desiderata:
   - General focus: Best-effort, but with a design focus on lowest realistic overhead
   - Desirable throughput: Two 16 bit channels at full rate (~25 Mbit/s).
   - Single client support is enough; can always direct to a pubsub server of some description in the future
   - Seamless streaming of samples over many seconds should be supported (e.g. for tracking loop performance down to low frequencies)
 - Design proposal:
   - UDP, single target host/port configured via MQTT interface.
   - Regular samples every N processing cycles from a selected subset of channels (ADC/DAC streams, internal lockin/… signals, …), configured via MQTT interface.
   - Each UDP message is a "streaming frame", consisting of a fixed-size header and a configurable-length payload.
     - Payload: Row-major i16 array of samples (make generic in the future?), where rows are time and columns are different channels
     - Header: Channels selected, sample divider used, samples per frame, global timestamp (e.g. ticks since Stabilizer boot) of first sample.
   - Firmware keeps a ring buffer of frames
     - DSP task: If not currently filling a frame, pull one from the ring buffer (try next iteration if none available), fill in current timestamp and used settings. If space in frame, copy in requested data samples. If frame full (or streaming stopped (?)), push back into ring buffer.
     - Idle task: Pull out any completed frames, push onto wire (ideally zero-copy, though I don't know whether that's feasible in current smoltcp), return to queue.
     - No atomics necessary on either side, but barriers/volatile for interrupt safety.
   - Changing streaming settings requires stop and restart (possibly missing some samples) for simplicity.
- Comments: 
[...]

PR #380, opened 2021-06-09
Adding data livestream capability

ryan-summers:
This PR implements livestream capability for ADC and DAC samples. The current implementation provides a static configuration.

**TODO**
- [x] Clean up documentation
- [x] Implement streaming for `lockin` app
- [x] Test new block-packetization scheme
- [x] Add reference python stream receiver
- [x] Merge #381 first


dnadlinger:
Glad to see movement on this front. I've only glanced over the implementation – how is the block id behave in situations limited by either CPU time or network throughput? I'd be quite keen on having some identifier that allows one to unambiguously reconstruct longer sequences even when there are missed samples/packet loss from use close to the line rate (see #150).

jordens:
We don't have redundancy. Ergo it's impossible to reconstruct data in the presence of loss.
You can only reconstruct the timestamps of the data you have received using the block id.

ryan-summers:
We don't have redundancy, so reconstruction is not possible. That being said, on a closed LAN, I am able to stream data at full data rates (e.g. 780KHz DAC and ADC codes).

This implementation supports detection of dropped samples, but doesn't provide any mechanism to remedy that situation.

dnadlinger:
> This implementation supports detection of dropped samples, but doesn't provide any mechanism to remedy that situation.

Alright – I meant "reconstruct the sequence" as in figuring out the absolute sample times for packets you do receive, rather than reconstructing missing data.

As long as your low-frequency (many packets FFT analyser can detect missing data, it can just try again/downgrade the sample rate/…

ryan-summers:
bors try

Cargo.toml
```diff
@@ -60,7 +60,7 @@ rev = "70b0eb5"
 features = ["stm32h743v", "rt", "unproven", "ethernet", "quadspi"]
 # version = "0.9.0"
 git = "https://github.com/quartiq/stm32h7xx-hal.git"
-rev = "acd47be"
+rev = "33aa67d"
```
ryan-summers, line 63: This is actually the commit that utilizes `master` smoltcp. I erroneously added that comment in the last PR.

ryan-summers:
bors try

hitl/run.sh
```diff
@@ -31,5 +31,7 @@ python3 miniconf.py dt/sinara/dual-iir/04-91-62-d9-7e-5f afe/0='"G2"'
 python3 miniconf.py dt/sinara/dual-iir/04-91-62-d9-7e-5f afe/0='"G1"' iir_ch/0/0=\
 '{"y_min": -32767, "y_max": 32767, "y_offset": 0, "ba": [1.0, 0, 0, 0, 0]}'
 
-kill $(jobs -p)
+# Probe-run appears to sporadically fail. In that case, the job is no longer active and there's no
+# background process to kill.
+kill $(jobs -p) || true
```
jordens, line 36: With this the job will succeed although it didn't even flash the firmware to be tested.

ryan-summers: What are your proposals? Otherwise, we sporadically see the `probe-run` fail to complete properly and the kill step then fails later because of it. I would propose we just use cargo-flash instead. Thoughts?

jordens: Yeah. I don't know why it fails. I could imagine that a newer probe-rs behind probe-run might handle that st-link better.
But doing for `cargo-flash` in the meantime is fine with me. I would change that in the hitl script explicitly and not the runner `.cargo/config`.

jordens (changes requested):
Looks good.
Do you think we could add a hitl test?

jordens:
And in my tests I can enable streaming exactly once. Disabling and then enabling doesn't send anything.

jordens:
Could you add some notes on the amount of copying that is going on here?
It looks to me like we're copying at the following places:

* In `send()` creation of `block`.
* In `send()`, `enqueue()` is a `memcpy()`. If `spsc` had a closure based API, we could create the block directly in the queue memory and get rid of that copy.
* In `process()`, `dequeue()` is a memcpy in both the discard path and the frame assembly path. That is certainly not needed since we don't do anything with the items. Might get optimized away, I haven't checked.
* In `process()`, nal `send()` calls `send_slice()` which does `copy_slice()`. Could directly build the frame into the smoltcp udp socket buffer returned by `send()`.
* Then the copy form socket buffer into the DMA buffer.

By using closure based APIs (`queue::enqueue_with(|&mut item| {item.adc = *adc...})` and `let size = queue.len() ...; let mut buf = smoltcp_socket::send(size, endpoint)?; for i in 0..batches { queue.dequeue_with(|&item| buf[].copy_from_slice(item)).unwrap()`), we could get rid of two copies.

jordens:
A different approach, it seems to me, is to get the packet buffer mut ref from smoltcp udp `send()` and then have the generator copy batch data directly into it, then let the lower prio swap buffers and do stack update. That also removes the block_id check (simply increase the block ID when there is no packet buffer available), dequeueing when the streaming starts.
I'm just unsure about the semantics of smoltcp though. It seems to return the ref to the packet and immediately enqueue it. It could be sent (copied into DMA) before the packet is full. That would be an annoying design.

jordens:
More context: the reason I'm looking at the copying is because `generator.send()` takes about 500 cycles. For the single copy it must and should do, that's a lot.

ryan-summers:
@jordens The only copy that occurs in the high-priority context is, as you note, a direct copy into the spsc queue. Given that the data previously exists in the DMA ram buffers, there is no escaping this copy. It always will have to exist there. All other data processing/copying occurs at low priority tasks in a best-effort manner, so any optimization we do would not accelerate the DSP process routine (to my knowledge).

I don't think there's any way around the copy, but it _is_ surprising to me how long the copy actually ends up taking. 500 cycles for 8 samples ( = 64 bytes of data + block ID) seems excessive to me. Maybe we can move the queue into DTCM?

jordens:
There are two copies in the high-prio context. One is the assembly of `block` (dereferencing) and the other is in `enqueue()`.

ryan-summers:
> And in my tests I can enable streaming exactly once. Disabling and then enabling doesn't send anything.

This may be a bug in the UDP open/close API. I'll look into it

jordens:
Stack and statics are already in DTCM.

dnadlinger:
Tangentially related: https://github.com/smoltcp-rs/smoltcp/issues/422

jordens (approved):

ryan-summers:
bors merge

ryan-summers:
bors merge

ryan-summers:
bors merge

ryan-summers:
bors retry

ryan-summers:
bors retry

PR #381, opened 2021-06-15
Updating to HAL mainline

ryan-summers:
This PR updates to utilize latest version of the HAL (in anticipation of bumping smoltcp). Updates are only around SPI CS control, and I think the new API, while a little distasteful in syntax, is more expressive and useful.

I've tested this on loopback hardware with a 1KHz 1V sinusoid and confirmed things look as expected.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -52,7 +52,10 @@ mcp23017 = "1.0"
 git = "https://github.com/quartiq/rtt-logger.git"
 rev = "70b0eb5"
 
-# fast double buffered DMA without poisoning and buffer swapping
+# The following modifications of the HAL are being used:
+# * fast double buffered DMA without poisoning and buffer swapping
+# * Utilize `master` branch of smoltcp
+# * Utilize `master` branch of HAL
 [dependencies.stm32h7xx-hal]
 features = ["stm32h743v", "rt", "unproven", "ethernet", "quadspi"]
 # version = "0.9.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -57,7 +57,7 @@ rev = "70b0eb5"
 features = ["stm32h743v", "rt", "unproven", "ethernet", "quadspi"]
 # version = "0.9.0"
 git = "https://github.com/quartiq/stm32h7xx-hal.git"
-rev = "b0b8a93"
+branch = "master"
 
 # link.x section start/end
 [patch.crates-io.cortex-m-rt]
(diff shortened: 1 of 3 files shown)


src/hardware/setup.rs
```diff
@@ -341,13 +341,16 @@ pub fn setup(
...
                 phase: hal::spi::Phase::CaptureOnSecondTransition,
             })
-            .manage_cs()
-            .suspend_when_inactive()
-            .communication_mode(hal::spi::CommunicationMode::Receiver)
-            .cs_delay(design_parameters::ADC_SETUP_TIME);
+            .inter_word_delay(design_parameters::ADC_SETUP_TIME)
+            .hardware_cs(hal::spi::HardwareCS {
+                mode: hal::spi::HardwareCSMode::WordTransaction,
+                assertion_delay: design_parameters::ADC_SETUP_TIME,
```
jordens, line 347: Why is `ADC_SETUP_TIME` in there twice? There should only be the delay between asserting CS and the first bit (MSSI). MIDI should be 0.
If that duplication just a quirk of the new API, can we add a comment to clarify?

Cargo.toml
```diff
@@ -57,7 +57,7 @@ rev = "70b0eb5"
 features = ["stm32h743v", "rt", "unproven", "ethernet", "quadspi"]
 # version = "0.9.0"
 git = "https://github.com/quartiq/stm32h7xx-hal.git"
-rev = "b0b8a93"
+rev = "acd47be"
```
jordens, line 60: Could you update the comment in L55 to refer to the smoltcp bump as well?

jordens (changes requested):

src/hardware/setup.rs
ryan-summers: No, MIDI should not be set to 0. We utilize MIDI because the SPI operates in endless transaction mode (TSIZE = 0), which means that we need to use MIDI as well as MSSI (MSSI is only used right as the transaction starts, then MIDI is used for all further words).

ryan-summers: However, upon probing the lines, it looks like ADC setup time is working even without this line. I'll look into this some more

ryan-summers: Ah, this is because my memory is failing me at an early age apparently. Upon reading docs around the ADCs, we actually removed the endless transaction mode in favor of explicitly clearing the EOT flag (I believe we had to to ensure setup timing). You're right that this line can be removed.

jordens (approved):

PR #387, opened 2021-06-22
Cargo-flash for HITL runner

### LINKED ISSUE #383: HITL runner may spuriously fail programming steps
It appears that `probe-run` spuriously fails to connect to the USB device, which causes the programming step (and later kill) to fail.

Perhaps we could utilize cargo-flash instead

ryan-summers:
This PR utilizes cargo-embed for flashing the HITL runner target. probe-run was giving us spurious faults.

This fixes #383


ryan-summers:
bors try

ryan-summers:
bors retry

PR #391, opened 2021-06-24
setup: use a best-effort lock for rprintln

### LINKED ISSUE #382: Logging messages may cause DSP timing to be missed
Logs via RTT utilize critical sections to preserve the logger from pre-emption. However, this critical section has been measured to be approximately 5uS in some cases. At a 780KHz signal with a batch size of 8, this can correspond to half the DSP processing routine. In some cases, this can result in the DSP routine being delayed and missing timing. This in turn causes the app to panic.

jordens:
close #382

* [x] tested with very long log messages that would otherwise cause the DMA deadline to be missed. (use cargo-embed git to make it work with stlink v3)

### DIFF
--- a/src/hardware/setup.rs
+++ b/src/hardware/setup.rs
@@ -234,13 +234,13 @@ pub fn setup(
 
         let channels = rtt_target::rtt_init_default!();
         // Note(unsafe): The closure we pass does not establish a critical section
-        // as demanded but it ensure synchronization and implements a lock.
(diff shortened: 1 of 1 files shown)


ryan-summers (approved):
Looks good. I agree with the ordering semantics used for the atomic load/store operations being Acquire/Release here.

ryan-summers:
Bors r+

PR #392, opened 2021-06-24
use mutex-trait instead of flatten_closures

jordens:
To have less code to maintain and clean up the API a bit.

* [x] tested on multiple applications (dual-iir, fls)


ryan-summers (approved):

jordens:
bors merge

PR #405, opened 2021-07-19
Accelerating doc CI tests

### LINKED ISSUE #403: doc: TODO
Unsorted docs items after #401:

* [ ] cache deadlinks and ruby stuff (like https://bsmth.de/blog/deploying-jekyll-github-actions-aws-amplify) to speed up CI
* [x] deadlinks `--check-http --check-intra-doc-links`
* [x] add docs to bors requirements

ryan-summers:
This PR fixes #403 by utilize github actions caches to accelerate the doc test CI check. `cargo-deadlinks` is cached, ruby gems are cached, and Rust dependency installations are cached.

Questions:
Should the Rust cache be tagged on a specific key so that we don't pollute the cache with each PR?

### DIFF
--- a/src/configuration.rs
+++ b/src/configuration.rs
@@ -3,7 +3,7 @@
 /// MQTT broker IPv4 address
 ///
 /// In the default configuration, the IP address is defined as 10.34.16.10.
-pub const MQTT_BROKER: [u8; 4] = [10, 35, 16, 10];
+pub const MQTT_BROKER: [u8; 4] = [10, 34, 16, 10];
 
 /// Sampling Frequency
 ///
--- a/src/configuration.rs
+++ b/src/configuration.rs
@@ -3,7 +3,7 @@
 /// MQTT broker IPv4 address
 ///
 /// In the default configuration, the IP address is defined as 10.34.16.10.
-pub const MQTT_BROKER: [u8; 4] = [10, 34, 16, 10];
+pub const MQTT_BROKER: [u8; 4] = [10, 35, 16, 10];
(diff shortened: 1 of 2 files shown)


src/configuration.rs
```diff
@@ -3,7 +3,7 @@
 /// MQTT broker IPv4 address
 ///
 /// In the default configuration, the IP address is defined as 10.34.16.10.
-pub const MQTT_BROKER: [u8; 4] = [10, 34, 16, 10];
+pub const MQTT_BROKER: [u8; 4] = [10, 35, 16, 10];
```
jordens, line 6: Spurious

```suggestion
pub const MQTT_BROKER: [u8; 4] = [10, 34, 16, 10];
```

jordens (changes requested):
Otherwise looks good.

src/configuration.rs
ryan-summers: Woops, looks like my local cache slipped in.

What are your thoughts on adding `src/configuration.rs` to the ignore list? The downside is that you don't see the diff in `git diff` if you modify it

jordens: I'd rather not. Ultimately the hardcoded configuration should go away: replaced by genericism (or just runtime configuration with sufficiently large upper limit of 32 sample static buffers) for those the first two parameters and runtime configuration by some means (DHCP, (m)DNS, flash config) for the broker.

jordens (approved):

ryan-summers:
bors merge

PR #410, opened 2021-07-20
tighten signal_generator symmetry bounds

### LINKED ISSUE #408: signal_generator: validate symmetry
https://github.com/quartiq/stabilizer/blob/cc79deacf7b4efdf9caaacfbb4fc9007661897ee/src/hardware/signal_generator.rs#L67-L76

The conversion to `u32` will overflow if `ftw/symmetry > u32::MAX as f32` or `ftw/(1.0 - symmetry) > u32::MAX as f32`.
The bounds check needs to look at those conditions as well and clip the FTWs. Note that there should not be any limit on `ftw`.

jordens:
close #408

### DIFF
--- a/src/hardware/signal_generator.rs
+++ b/src/hardware/signal_generator.rs
@@ -77,7 +77,7 @@ impl TryFrom<BasicConfig> for Config {
 
         // Validate base frequency tuning word to be below Nyquist.
         const NYQUIST: f32 = (1u32 << 31) as _;
-        if 2.0 * ftw > NYQUIST {
+        if ftw < 0.0 || 2.0 * ftw > NYQUIST {
             return Err(Error::InvalidFrequency);
         }
 
--- a/src/hardware/signal_generator.rs
+++ b/src/hardware/signal_generator.rs
@@ -1,4 +1,7 @@
-use crate::{configuration::ADC_SAMPLE_TICKS_LOG2, hardware::dac::DacCode};
+use crate::{
+    configuration::ADC_SAMPLE_TICKS_LOG2, hardware::dac::DacCode,
+    hardware::design_parameters::TIMER_FREQUENCY,
+};
 use core::convert::{TryFrom, TryInto};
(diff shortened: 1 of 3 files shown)


src/hardware/signal_generator.rs
```diff
@@ -60,13 +63,14 @@ impl TryFrom<BasicConfig> for Config {
...
         let frequency_tuning_word: [u32; 2] = {
-            const LSB_PER_HERTZ: f32 =
-                (1u64 << (31 + ADC_SAMPLE_TICKS_LOG2)) as f32 / 100.0e6;
+            const LSB_PER_HERTZ: f32 = (1u64 << (31 + ADC_SAMPLE_TICKS_LOG2))
+                as f32
+                / (TIMER_FREQUENCY.0 * 1_000_000) as f32;
             let ftw = config.frequency * LSB_PER_HERTZ;
 
-            if config.symmetry <= 0.0 {
+            if config.symmetry <= ftw / u32::MAX as f32 {
```
ryan-summers, line 71: There are two bounds checks we need to satisfy:
* `ftw / symmetry <= u32::MAX as f32` and `ftw / (1 - symmetry) <= u32::MAX as f32`
* `config.symmetry` in [0, 1]

It looks like while trying to combine them into a single comparison (If I'm understanding the changes correctly), the second check for symmetry was lost. IMO it's more clear if we don't try to combine them and leave them as two explicit checks. This code doesn't get executed very often, so the slight loss in efficiency is more than compensated for code legibility.

ryan-summers (changes requested):

src/hardware/signal_generator.rs
jordens: To me it's one and the same check. But I have added a couple more checks which also make what's being checked/clipped/rounded clearer.

jordens:
need to add another check
bors cancel

jordens:
bors merge

PR #412, opened 2021-07-21
Updating IO expander to be configured before AD9959

### LINKED ISSUE #411: Intermittent Pounder setup failure after power on
With pounder installed on stabilizer, there exist an issue where the code always panics at first cold boot (power plugged in), persist for random amount of MCU reset cycles, before the setup process eventually succeeds by repeated resetting of the MCU, and will keep on working through MCU resets, until power to the system is lost.

Error log message:
```
INFO - Starting
INFO - EUI48: 04-91-62-d9-f5-e5
INFO - Found Pounder
panicked at 'called `Result::unwrap()` on an `Err` value: Check', src/hardware/setup.rs:835:14
```
The code panics during the AD9959 setup process in ad9959/src/lib.rs at
```
// Read back the CSR to ensure it specifies the mode correctly.
let mut updated_csr: [u8; 1] = [0];
ad9959.read(Register::CSR, &mut updated_csr)?;
if updated_csr[0] != csr[0] {
    return Err(Error::Check);
}
```
, where the QSPI read back of the CSR register reads 0, which does not match the content written into the register with SPI a few lines before.

The issue is caused by the Pounder setup code initializing the AD9959 chip before the MCP23017 IO expander. Pounder MOSFET T1 (connected to GPB6 of MCP23017) controls power to Pounder OSC1, which is the system clock source for the AD9959 chip. As the MCP23017 IO ports default to floating inputs after power on, the gate of T1 is floating during AD9959 setup, meaning the presence of power to OSC1 and clock source of AD9959 depends on random built up gate charges on T1. 
Screenshot from 2021-07-21 12-32-50

Page 30 of the AD9959 datasheet mentions the IO_UPDATE pin must be held high for one SYNC_CLK cycle before the content in the SPI/QSPI buffer is written into the device register. SYNC_CLK is internally set to 1/4 of the AD9959 system clock, and with the presence of system clock being random, the initial SPI write into CSR register may or may not have updated the CSR register's contents. 
Screenshot from 2021-07-21 12-33-34

If CSR is not updated, QSPI reads back 0, triggering a panic. 
Screenshot from 2021-07-20 15-04-04

If CSR is updated when IO_UPDATE is held high due to random turn on of T1, or previous successful Pounder setup, the AD9959 setup process passes, MCP23017 GPB6 is set low, and the system will work until power is removed.
Screenshot from 2021-07-20 15-05-39

The solution of this issue is to first setup MCP23017, wait 1 ms for OSC1 output to stabilize, then setup AD9959.

ryan-summers:
This PR fixes #411 by refactoring the code to initialize the IO expander before the AD9959. It was discovered that the oscillator to the AD9959 may not be enabled due to a floating MOSFET gate. This was causing register configs to fail.

The fix configures the IO expander first to ensure that the OSC_ENn signal is de-asserted to enable the AD9959 oscillator.

### DIFF
--- a/src/hardware/setup.rs
+++ b/src/hardware/setup.rs
@@ -769,6 +769,26 @@ pub fn setup(
     delay.delay_ms(2u8);
     let pounder = if pounder_pgood.is_high().unwrap() {
         log::info!("Found Pounder");
+
+        let mut io_expander = {
+            let sda = gpiob.pb7.into_alternate_af4().set_open_drain();
+            let scl = gpiob.pb8.into_alternate_af4().set_open_drain();
+            let i2c1 = device.I2C1.i2c(
+                (scl, sda),
+                400.khz(),
+                ccdr.peripheral.I2C1,
+                &ccdr.clocks,
+            );
+            mcp23017::MCP23017::default(i2c1).unwrap()
+        };
+
+        // Configure power-on-default state for pounder. All LEDs are off, on-board oscillator
+        // selected and enabled, attenuators out of reset. Note that testing indicates the
+        // output state needs to be set first to properly update the output registers.
+        io_expander.all_pin_mode(mcp23017::PinMode::OUTPUT).unwrap();
+        io_expander.write_gpio(mcp23017::Port::GPIOA, 0x00).unwrap();
+        io_expander.write_gpio(mcp23017::Port::GPIOB, 0x2F).unwrap();
+
         let ad9959 = {
             let qspi_interface = {
                 // Instantiate the QUADSPI pins and peripheral interface.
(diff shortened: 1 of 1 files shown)


jordens (changes requested):
The setup is done twice now making it hard to maintain. What's the rationale?
Why not just move the `ad9959` creation after the `PounderDevices::new()`?

ryan-summers:
> The setup is done twice now making it hard to maintain. What's the rationale?
> Why not just move the `ad9959` creation after the `PounderDevices::new()`?

Because I was (falsely) under the impression that `PounderDevices` needed the `ad9959` for construction. Will update - I don't like the duplicity either.

jordens:
LGTM
bors r+

PR #414, opened 2021-07-23
Feature/generic stream

### LINKED ISSUE #385: streaming: better buffering
Reduce the number of copies for streaming:

* The high prio process task should directly serialize into a fixed-size buffer A (about 1 kB).
* If A is full and another buffer A' (that's been released from the low prio network stack task) is available it designates A as B and switches over to filling A' (format ID and start batch number and first batch)
* If there is no A' available it drops the batch.
* If the network stack low prio task has a full buffer B available it copies it into the smoltcp socket buffer and designates the now free buffer B as A' closing the loop.
* Maybe this entire thing is easiest with a heapless `Pool` for the three buffers and heapless boxes for the full ones into a N=4 spsc queue (https://github.com/japaric/heapless/issues/210) to implement the concurrency (like [here](https://github.com/stm32-rs/stm32l4xx-hal/blob/master/examples/rtic_frame_serial_dma.rs)).
* Once https://github.com/smoltcp-rs/smoltcp/issues/422 is fixed these three buffers would all be smoltcp UDP packet buffers and the network task would only enqueue.

At the same time we should do the following:

* Make the streaming generic over the batch data format, or better generic over the size of the u8 slice to serialize each batch data into. Applications will want to not only log 2ADC and 2 DAC but other things instead or additionally.
* The frame header hould be `#[repr(C)] {sequence: u16, format: u16}` where `sequence` is the sequence number of the frame (not the first batch in the frame: frame size is fixed and batch format/size is defined elsewhere), format is an opaque format ID defined by the application defining the contents of a batch. `format` implies batch data size. Frame size minus header size implies number of batches contained.
* Endianess should be device-endian for everything.
* Rewrite streaming docs afterwards

ryan-summers:
This PR fixes #385 by improving the semantics by which streams are generated. Specifically, serialization format is now application-defined and data is only copied once.

TODO:
- [x] Add documentation for streaming architecture
- [x] Add format specifiers
- [x] Clean up reception script
- [x] Update lockin serialization

### DIFF
--- a/scripts/stream_throughput.py
+++ b/scripts/stream_throughput.py
@@ -14,6 +14,11 @@ import logging
 # Representation of a single data batch transmitted by Stabilizer.
 Packet = collections.namedtuple('Packet', ['index', 'data'])
 
+# Specifies a known format for incoming packets.
+#
+# * `sample_size_bytes` is the number of bytes required for each sample in the batch.
+# * `batch_format` is a `struct` format string that will be provided the `batch_size` as an named
+# argument. This format string will be used to deserialize each batch of data from the frame.
 Format = collections.namedtuple('Format', ['sample_size_bytes', 'batch_format'])
 
 # All supported formats by this reception script.
@@ -158,7 +163,7 @@ def main():
 
     while True:
         # Receive any data over UDP and parse it.
-        data = connection.recv(4096 * 4)
+        data = connection.recv(4096)
         if data and not timer.is_started():
             timer.start()
 
--- a/scripts/stream_throughput.py
+++ b/scripts/stream_throughput.py
@@ -13,6 +13,12 @@ import logging
 # Representation of a single UDP packet transmitted by Stabilizer.
 Packet = collections.namedtuple('Packet', ['index', 'adc', 'dac'])
 
+Format = collections.namedtuple('Format', ['batch_size', 'batches_per_frame'])
+
+FORMAT = {
+    0: Format(8, 255)
+}
+
 class Timer:
     """ A basic timer for measuring elapsed time periods. """
 
@@ -88,9 +94,10 @@ class PacketParser:
         if len(self.buf) < 4:
             return None
 
-        start_id, num_blocks, data_size = struct.unpack_from('!HBB', self.buf)
+        start_id, format_id = struct.unpack_from('!HH', self.buf)
 
-        packet_size = 4 + data_size * num_blocks * 8
+        frame_format = FORMAT[format_id]
+        packet_size = 4 + frame_format.batch_size * frame_format.batches_per_frame * 8
 
         if len(self.buf) < packet_size:
             return None
--- a/scripts/stream_throughput.py
+++ b/scripts/stream_throughput.py
@@ -32,7 +32,7 @@ def parse_packet(buf):
     """ Attempt to parse packets from the received buffer. """
     # Attempt to parse a block from the buffer.
     if len(buf) < struct.calcsize(HEADER_FORMAT):
-        return None
+        return []
 
     # Parse out the packet header
     magic, format_id, batch_size, sequence_number = struct.unpack_from(HEADER_FORMAT, buf)
@@ -40,14 +40,14 @@ def parse_packet(buf):
 
     if magic != MAGIC_HEADER:
         logging.warning('Encountered bad magic header: %s', hex(magic))
-        return None
+        return []
 
     if format_id not in FORMAT:
         raise Exception(f'Unknown format specifier: {format_id}')
 
     frame_format = FORMAT[format_id](batch_size)
 
-    batch_count = len(buf) / struct.calcsize(frame_format)
+    batch_count = int(len(buf) / struct.calcsize(frame_format))
 
     packets = []
     for offset in range(batch_count):
@@ -135,9 +135,7 @@ def main():
 
         # Handle any received packets.
         total_bytes += len(data)
-        packet = parse_packet(data)
-
-        if packet:
+        for packet in parse_packet(data):
             # Handle any dropped packets.
             drop_count += sequence_delta(last_index, packet.index)
             last_index = packet.index
--- a/scripts/stream_throughput.py
+++ b/scripts/stream_throughput.py
@@ -14,18 +14,49 @@ import logging
 # Representation of a single data batch transmitted by Stabilizer.
 Packet = collections.namedtuple('Packet', ['index', 'data'])
 
-# Specifies a known format for incoming packets.
-#
-# * `sample_size_bytes` is the number of bytes required for each sample in the batch.
-# * `batch_format` is a `struct` format string that will be provided the `batch_size` as an named
-# argument. This format string will be used to deserialize each batch of data from the frame.
-Format = collections.namedtuple('Format', ['sample_size_bytes', 'batch_format'])
+# The magic header half-word at the start of each packet.
+MAGIC_HEADER = 0x057B
+
+# The struct format of the header.
+HEADER_FORMAT = '<HBBI'
 
 # All supported formats by this reception script.
+#
+# The items in this dict are functions that will be provided the sampel batch size and will return
+# the struct deserialization code to unpack a single batch.
 FORMAT = {
-    0: Format(sample_size_bytes=8,
-              batch_format='<{batch_size}H{batch_size}H{batch_size}H{batch_size}H')
+    1: lambda batch_size: f'<{batch_size}H{batch_size}H{batch_size}H{batch_size}H'
 }
+
+def parse_packet(buf):
+    """ Attempt to parse packets from the received buffer. """
+    # Attempt to parse a block from the buffer.
+    if len(buf) < struct.calcsize(HEADER_FORMAT):
+        return None
+
+    # Parse out the packet header
+    magic, format_id, batch_size, sequence_number = struct.unpack_from(HEADER_FORMAT, buf)
+    buf = buf[struct.calcsize(HEADER_FORMAT):]
+
+    if magic != MAGIC_HEADER:
+        logging.warning('Encountered bad magic header: %s', hex(magic))
+        return None
+
+    if format_id not in FORMAT:
+        raise Exception(f'Unknown format specifier: {format_id}')
+
+    frame_format = FORMAT[format_id](batch_size)
+
+    batch_count = len(buf) / struct.calcsize(frame_format)
+
+    packets = []
+    for offset in range(batch_count):
+        data = struct.unpack_from(frame_format, buf)
+        buf = buf[struct.calcsize(frame_format):]
+        packets.append(Packet(sequence_number + offset, data))
+
+    return packets
+
 
 class Timer:
     """ A basic timer for measuring elapsed time periods. """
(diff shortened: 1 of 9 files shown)


src/net/data_stream.rs
```diff
@@ -73,216 +107,143 @@ pub struct AdcDacData {
...
-        // Copy the samples into the buffer.
-        for device in &[batch.adcs, batch.dacs] {
-            for channel in device {
-                for sample in channel.iter().step_by(self.subsample_rate) {
-                    self.buf[self.write_index..self.write_index + 2]
-                        .copy_from_slice(&sample.to_be_bytes());
-                    self.write_index += 2;
-                }
+        assert!(
+            format == self.current_frame.as_ref().unwrap().format,
```
jordens, line 230: This should go away if the `FrameGenerator` is generic over the format ID and serialized batch size.

src/net/data_stream.rs
```diff
@@ -73,216 +107,143 @@ pub struct AdcDacData {
...
+        result
+    }
+
+    pub fn is_full<const T: usize>(&self) -> bool {
+        self.offset + T >= self.buffer.len()
+    }
+
+    pub fn finish(&mut self) -> &[u8] {
+        let offset = self.offset;
+        self.buffer[0..2].copy_from_slice(&self.sequence_number.to_ne_bytes());
```
jordens, line 175: Ah. `ne` is not "network" but "native"...

src/net/data_stream.rs
```diff
@@ -73,216 +107,143 @@ pub struct AdcDacData {
...
-pub struct BlockGenerator {
-    queue: Producer<'static, AdcDacData, BLOCK_BUFFER_SIZE>,
-    current_id: u16,
+struct StreamFrame {
+    format: StreamFormat,
+    sequence_number: u16,
+    buffer: Box<[u8; FRAME_SIZE], Init>,
+    offset: usize,
+    batch_count: u16,
+    batch_size: u8,
```
jordens, line 136: I think this is more complicated and heavy then necessary.
I would make the `FrameGenerator` generic (or by some other means knowledgable) over the (all compile-time constants) length of a frame, the number of batches per frame and the format. Also make the frame size an even multiple of the batch size (plus header). Full checking is easier that way and offset and batch count become irrelevant.
Then there is no need to put all this additional data (format, sequence_number, offset, batch_count, batch_size) into each item in the Queue. Remember this all gets copied. Everything apart from the frame sequence number is constant.
The sequence number together with the format should be added to the frame after the `alloc()` in `add()` (or just before `enqueue()`).

Also see the comment on `setup_streaming()`.

src/net/data_stream.rs
```diff
@@ -73,216 +107,143 @@ pub struct AdcDacData {
...
-                    self.write_index += 2;
-                }
+        assert!(
+            format == self.current_frame.as_ref().unwrap().format,
+            "Unexpected stream format encountered"
+        );
+
+        self.current_frame.as_mut().unwrap().add_batch::<_, T>(f);
+
+        if self.current_frame.as_ref().unwrap().is_full::<T>() {
```
jordens, line 236: Beyond the clippy lint I suspect this can be simplified further with some `if let` or by merging it with the `as_ref()` above. At least there should be a way to cancel a few of the 4 `unwrap()`s here (L230, L234, L236, L239).

src/net/data_stream.rs
```diff
@@ -73,216 +107,143 @@ pub struct AdcDacData {
...
+        format: StreamFormat,
+        sequence_number: u16,
+    ) -> Self {
         Self {
-            queue,
-            current_id: 0,
+            format,
+            offset: 7,
+            sequence_number,
+            buffer: unsafe { buffer.assume_init() },
```
jordens, line 149: I wonder whether it would be better to follow through with the notion that this is uninitialized data. I would not allow the closure (to `add()`) to read stale/previous data in the buffers.
The closure is the recipient of the "out-pointer" in https://doc.rust-lang.org/std/mem/union.MaybeUninit.html#out-pointers. And the closure will be doing unsafe stuff in general.

src/net/data_stream.rs
```diff
@@ -73,216 +107,143 @@ pub struct AdcDacData {
 /// `stream` is the logically consumer (UDP transmitter) of the enqueued data.
 pub fn setup_streaming(
     stack: NetworkReference,
-) -> (BlockGenerator, DataStream) {
-    let queue = cortex_m::singleton!(: Queue<AdcDacData, BLOCK_BUFFER_SIZE> = Queue::new()).unwrap();
-
+) -> (FrameGenerator, DataStream) {
+    let queue =
+        cortex_m::singleton!(: Queue<StreamFrame, FRAME_COUNT> = Queue::new())
```
jordens, line 112: The `Queue` items should not be more than `Box<[u8; FRAME_SIZE]>`. I'd make `setup_streaming` generic over batch size, batches_per_frame and add `format` to `FrameGenerator::new()`. That also allows making the frame an exact multiple of the batch size plus header size.

jordens (changes requested):
Nice work. Let me know what you think about the two bigger things (`MaybeUninit` and the genericism/light-weight queue items).

src/net/data_stream.rs
ryan-summers: This is not a `MaybeUninit` structure, but rather a `heapless::pool::Box<T, Uninit>`. The only way to get a `Box<T, Init>` is to call `Box::init(t: T)`, where a copy from `t` -> inner data is performed. If we actually did that, we'd have to enforce a 1KB copy (of zero'd data) into the buffer. I wanted to avoid that overhead, since we never read the buffer anyways.

src/net/data_stream.rs
ryan-summers: Yeah, I had originally implemented it similar to this, but I didn't like how the generic parameters started propagating through a lot of the code base. In any case, it still seems to make more sense to make it a const-generic parameter. Will update.

src/net/data_stream.rs
ryan-summers: The batch count and batch size are not for this driver, but rather for the UDP receiver. Is your comment more around just removing them from this struct and having them only exist in the header (since they're const generics, we don't have to carry them around, as they're part of the type signature)?

src/net/data_stream.rs
ryan-summers: This threw me through a loop as well.

src/net/data_stream.rs
jordens: Both.
The receiver knows batch size and content from batch format ID. It knows batches_per_frame either from the format ID (if we let the application define that) or simply by the size of the UDP packet (they are not split).
The other aspect is that `StreamFrame` is the queue item and is being copied into and out of it. None of the data apart from `buffer` should be in the queue because it is either known/constant or has no bearing on the processing at the other end of the queue.

src/net/data_stream.rs
jordens: Are you sure. The only example I'm aware of is this: https://github.com/stm32-rs/stm32l4xx-hal/blob/master/examples/rtic_frame_serial_dma.rs#L92
`DMAFrame::new()` has a `MaybeUninit::uninit()` member. The `dma_buf.init()` would copy uninitialized data?

src/net/data_stream.rs
jordens: Maybe with a few associated types as shorthands?

src/net/data_stream.rs
ryan-summers: `core::ptr::write()` is just a call to memcpy underneath. Thus, even if we have a `MaybeUninit::uninit()` object, we're still copying the entire MaybeUninit structure, which contains the underlying data + some. Check out goldbolt for the compiler output: https://godbolt.org/z/n8r6EexK6 - both do a memcpy of 1024 bytes.

jordens: Hmm. Right.
Would be interesting to discuss with @korken89 what the intended usage pattern is here without assume_init.

src/net/data_stream.rs
ryan-summers: I'm going to elect to keep things not templated on batch size and batch count per packet. The template parameters start to get really messy and cause a lot of pain in other areas of the codebase. I think it would make more sense to make them just constant struct members. I'll get a small refactor going to clean things up generally in the aspects mentioned above though.

src/net/data_stream.rs
ryan-summers: I'm electing to leave this without all the generic parameters - it started getting pretty out-of-hand and didn't support use cases where streaming wasn't desired at all. The `Queue` is now just the box'd array + a length, so it's not much overhead.

jordens: Why length if the frame is always full?

scripts/stream_throughput.py
```diff
@@ -52,134 +97,60 @@ def elapsed(self):
...
     good_blocks = 0
+    total_bytes = 0
 
     timer = Timer()
-    parser = PacketParser()
 
     while True:
         # Receive any data over UDP and parse it.
-        data = connection.recv(4096)
+        data = connection.recv(1024)
```
jordens, line 132: That's wrong. It only works due to the off-by-one error below.

```suggestion
        data = connection.recv(4096)
```

is good.

src/net/data_stream.rs
```diff
@@ -73,216 +119,141 @@ pub struct AdcDacData {
...
+
+        let result = f(&mut self.buffer[self.offset..self.offset + T]);
+
+        self.offset += T;
+
+        result
+    }
+
+    pub fn is_full<const T: usize>(&self) -> bool {
+        self.offset + T >= self.buffer.len()
```
jordens, line 180:
```suggestion
        self.offset + T > self.buffer.len()
```

scripts/stream_throughput.py
```diff
@@ -4,14 +4,59 @@
...
+
+    frame_format = FORMAT[format_id](batch_size)
+
+    batch_count = int(len(buf) / struct.calcsize(frame_format))
+
+    packets = []
+    for offset in range(batch_count):
+        data = struct.unpack_from(frame_format, buf)
+        buf = buf[struct.calcsize(frame_format):]
+        packets.append(Packet(sequence_number + offset, data))
```
jordens, line 56:
```suggestion
        yield Packet(sequence_number + offset, data)
```

and remove the empty lists in `return []` and the `return packets` below.

src/net/mod.rs
```diff
@@ -113,8 +113,16 @@ where
     }
 
     /// Enable live data streaming.
-    pub fn enable_streaming(&mut self) -> BlockGenerator {
-        self.generator.take().unwrap()
+    ///
+    /// # Args
+    /// * `format` - A unique u8 code indicating the format of the data.
+    pub fn enable_streaming(
```
jordens, line 119: This doesn't really enable streaming.
More `get_streaming_generator(format: impl Into<u8>)`?

jordens (changes requested):

src/net/data_stream.rs
```diff
@@ -73,216 +119,141 @@ pub struct AdcDacData {
...
+    where
+        F: FnMut(&mut [u8]),
+    {
+        assert!(!self.is_full::<T>(), "Batch cannot be added to full frame");
+
+        let result = f(&mut self.buffer[self.offset..self.offset + T]);
+
+        self.offset += T;
+
+        result
```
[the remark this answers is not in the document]
ryan-summers: This is vestigial. The function (and the FnMut() type) do not return anything, so this is just returning the empty () type. I'll remove it.

src/net/data_stream.rs
```diff
@@ -73,216 +119,141 @@ pub struct AdcDacData {
...
-            dacs: [*dacs[0], *dacs[1]],
-        };
-
-        self.current_id = self.current_id.wrapping_add(1);
-        self.queue.enqueue(block).ok();
+    pub fn add_batch<F, const T: usize>(&mut self, mut f: F)
+    where
+        F: FnMut(&mut [u8]),
+    {
+        assert!(!self.is_full::<T>(), "Batch cannot be added to full frame");
```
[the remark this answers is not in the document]
ryan-summers: Agreed - I think this is not worth checking. If it does fail, there's a firmware issue, and the slice indexing panic will be more helpful.

src/net/data_stream.rs
```diff
@@ -2,10 +2,31 @@
...
+//! * Format Code <u8>
+//! * Batch Size <u8>
+//! * Sequence Number <u32>
+//!
+//! The "Magic word" is a constant field for all packets. The value is alway 0x057B.
+//!
+//! The "Format Code" is a unique specifier that indicates the serialization format of each batch of
+//! data in the frame. Refer to [StreamFormat] for further information.
+//!
+//! The "Batch size" is the value of [SAMPLE_BUFFER_SIZE].
```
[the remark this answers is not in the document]
ryan-summers: This was largely to facilitate the parameterization of SAMPLE_BUFFER_SIZE in the future as a run-time constant. I'll bubble it all the way out to the public API though.

src/net/data_stream.rs
```diff
@@ -15,14 +36,25 @@ use miniconf::MiniconfAtomic;
...
 
-// The number of data blocks that we will buffer in the queue.
-const BLOCK_BUFFER_SIZE: usize = 30;
+use heapless::pool::{Box, Init, Pool, Uninit};
+
+use super::NetworkReference;
+
+const MAGIC_WORD: u16 = 0x057B;
+
+const HEADER_SIZE: usize = 8;
```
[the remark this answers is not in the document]
ryan-summers: I want to avoid usage of `repr(packed)` due to all of the unsoundness discussed around it. We need to think carefully about alignment of the structure or we could cause unexpected hardfault's of the CPU due to unaligned access. As such, I think it's easier and less painful to just define this as a static value.

scripts/stream_throughput.py
ryan-summers: That'll teach me to push changes without testing them.

jordens (approved):
Nice

PR #418, opened 2021-07-27
rj/fixes

jordens:
- remove old GPL license text
- signal_generator: actually catch the extreme symmetry cases (0 and 1)
- stream_throughput: fix port argument


jordens:
bors merge

PR #420, opened 2021-07-28
Miscellaneous

jordens:
* dds: name consistently
* dds: inline
* refactor the broker ip parsing to be panicing
* speed up qspi to 100 MHz (tested)
* tweak pounder qspi xfer assembly padding for speed
* fix some clippy lints
* try 32 byte max qspi xfer size instead of 16 but revert that until the memclr that's being pulled is can be avoided.


src/bin/dual-iir.rs
```diff
@@ -200,7 +200,7 @@ const APP: () = {
             stabilizer.cycle_counter,
             env!("CARGO_BIN_NAME"),
             stabilizer.net.mac_address,
-            net::parse_or_default_broker(option_env!("BROKER")),
+            net::parse_ipv4(option_env!("BROKER").unwrap_or("10.34.16.10")),
```
ryan-summers, line 203: If all `parse_ipv4()` does is call the `parse()` function, why not just inline as:
```rust
option_env!("BROKER").unwrap_or("10.34.16.10").parse().unwrap()
```

src/net/mod.rs
```diff
@@ -209,29 +209,17 @@ pub fn get_device_prefix(
...
-                 Some,
-             )
-         })
-         .unwrap_or_else(|| DEFAULT_MQTT_BROKER.into())
+/// The IP address.
+///
+/// # Panic
+/// This panics if parsing fails.
+pub fn parse_ipv4(input: &str) -> IpAddr {
+    input.parse::<minimq::embedded_nal::IpAddr>().unwrap()
```
ryan-summers, line 224: 1. I don't think you need to tell `parse()` the type here because it should be able to infer it via the function return type - or was it giving you errors otherwise?
2. This implementation will panic if the BROKER command-line argument was invalid - I was debating that behavior originally, but opted for defaulting back to the default broker. I can understand why just straight panicking may be more desirable, but I just want to make sure this is intended

src/bin/dual-iir.rs
jordens: Ah right! I had thought I had to wrestle to get some trait into scope.

src/net/mod.rs
jordens: Right. I just hadn't noticed I can remove the type.
I'm for panicing here. If the user specifies a broker IP at compile time that means the default broker will not work for them. If the user specifies an invalid IP it doesn't make much sense to proceed with a default broker because we know it won't work. And since the intention was to get a working broker connection, we should rather fail/panic than try something that we know won't work.

ad9959/src/lib.rs
```diff
@@ -597,20 +599,16 @@ impl ProfileSerializer {
...
-                self.add_write(Register::CSR, &[(self.mode as u8) << 1]);
-                self.add_write(Register::LSRR, &[0, 0]);
-            }
-            2 => self.add_write(Register::CSR, &[(self.mode as u8) << 1]),
-            1 => self.add_write(Register::LSRR, &[0, 0]),
-            0 => {}
-
-            _ => unreachable!(),
+        if self.index & 1 != 0 {
+            self.add_write(Register::LSRR, &[0, 0]);
```
ryan-summers, line 606: Mind adding some comments that adding an LSRR write pads us to even size-alignment, and then the CSR pads us to 4-byte (word) alignment?

ryan-summers (commented):
LGTM - requesting one minor comment addition, but otherwise all looks good. I'm assuming you've worked successfully with the 100MHz frequency and I have not personally tested this, but would be happy to if desired.

ad9959/src/lib.rs
jordens: I added more comments explaining the strategy.

jordens:
bors merge

PR #422, opened 2021-08-02
Updating Stabilizer to remove compile-time config parameters

### LINKED ISSUE #417: Move hardcoded constants out of library
The constants in `src/configuration.rs` should all be supplied by the application (via compile time generics or runtime parameters) such that using stabilizer as a library in an application doesn't require touching the source code library.

Until either mDNS or flash-based configuration is developed for the broker IP, this could also allow setting the broker IP on the `cargo {run|build}` command line.

It also has a significant effect on the on-boarding doc complexity.

ryan-summers:
This PR removes `src/configuration.rs` in favor of application-level constants. It also updates the MQTT broker IP to be specified via the command-line during compilation, and will default to 10.34.16.10

This should pave the way to storing some of these parameters potentially in flash for loading on boot.

This fixes #417


src/bin/dual-iir.rs
```diff
@@ -183,15 +187,34 @@ const APP: () = {
...
+                                "{:?}: Failed to parse broker IP ({:?}) - Falling back to default",
+                                err,
+                                data
+                            );
+                            None
+                        },
+                        |ip| Some(ip),
+                    )
+                })
+                .unwrap_or(DEFAULT_MQTT_BROKER.into()),
```
jordens, line 217: What about making this a `net::parse_or_default_broker(option_env!("BROKER"))` helper function and move it together with `DEFAULT_MQTT_BROKER` into `net`? Until we have a runtime way to get or configure this every app will likely use this.

There are also two clippy complaints about the code.

src/bin/dual-iir.rs
```diff
@@ -64,6 +61,13 @@ const SCALE: f32 = i16::MAX as _;
 // The number of cascaded IIR biquads per channel. Select 1 or 2!
 const IIR_CASCADE_LENGTH: usize = 1;
 
+// The number of samples in each batch process
+const SAMPLE_BUFFER_SIZE: usize = 8;
```
jordens, line 65: Can we seize the opportunity to rename this to a more accurate `BATCH_SIZE`? (It not isn't the size of `SampleBuffer` anymore and it is always the batch size both for ADCs and DACs).

Nice touch of showing the distinction of using a log2 batch size for lockin and that the ticks doesn't need to be log2 unless the app demands it!

jordens (changes requested):
Great!

src/bin/dual-iir.rs
```diff
@@ -228,10 +235,14 @@ const APP: () = {
...
                 SignalGenerator::new(
-                    settings.signal_generator[0].try_into().unwrap(),
+                    settings.signal_generator[0]
+                        .try_into_config(SAMPLE_TICKS_LOG2)
+                        .unwrap(),
                 ),
                 SignalGenerator::new(
-                    settings.signal_generator[1].try_into().unwrap(),
+                    settings.signal_generator[1]
+                        .try_into_config(SAMPLE_TICKS_LOG2)
```
ryan-summers, line 244: @jordens There's no other reason that `dual-iir` needs to restrict the sample rate to LOG2 values except for the signal generator, where we do an optimized division to calculate the FTW. Should we refactor the signal generator FTW calculation to support other tick sizes as well, or leave it for now?

jordens: Leave it for now. That restriction can be removed later.

jordens (commented):
LGTM, just the pounder code needs adapting as well.

I think the prescaler there should be a fixed `Div8` (250/8 < 100/3 as required for oversampling), not the batch size log2.

src/hardware/pounder/timestamp.rs
```diff
@@ -56,11 +57,16 @@ impl Timestamper {
         let mut input_capture = capture_channel
             .into_input_capture(timers::tim8::CaptureSource1::TRC);
 
+        let prescaler = match batch_size {
```
jordens, line 60: Just fix it to `Div8` as required for oversampling. The batch size shouldn't factor into it. I think this is an error that didn't have any consequence so far.

jordens (changes requested):

src/hardware/pounder/timestamp.rs
ryan-summers: I thought the intent here was to have a single timestamp per batch coming from pounder?

jordens: You are right!
This one should be the batch size.
The prescaler I had in mind was the `etps()` that's (correctly) set in `setup()` to `Div4`. (Both are on the same timer.)

jordens (approved):

ryan-summers:
bors merge

PR #425, opened 2021-08-05
Feature/loopback hitl

### LINKED ISSUE #354: Implement loop-back HITL ADC/DAC tests
Implement HITL tests for stabilizer in a loop-back configuration.

Specifically, the IIR filter should be configured to output static values, and then telemetry should be used to verify that the ADC samples then appear at the proper voltage.

We should also exercise the AFE programmable gain amplifiers in this test as well. Sample tests:
* Generate 1V output, AFE = 1x, verify 1V input
* Generate 1V output, AFE = 2x, verify 1V
* Generate 0V output, verify 0V input

This should give us a good idea that the ADC-> DAC processing routing is working as anticipated. Once data streaming is implemented, we should also be able to detect the latency from output -> input as well.

ryan-summers:
This PR fixes #354 by implementing loop-back HITL tests for Stabilizer.


ryan-summers:
bors try

ryan-summers:
bors try

ryan-summers:
bors try

ryan-summers:
bors retry

ryan-summers:
bors cancel

ryan-summers:
bors try

hitl/loopback.py
```diff
@@ -0,0 +1,148 @@
...
+        self._telemetry = []
+        self.client.on_message = self.handle_telemetry
+        self._telemetry_topic = f'{prefix}/telemetry'
+        self.client.subscribe(self._telemetry_topic)
+
+
+    def handle_telemetry(self, _client, topic, payload, _qos, _properties):
+        """ Handle incoming telemetry messages over MQTT. """
+        assert topic == self._telemetry_topic
+        self._telemetry.append(json.loads(payload))
```
jordens, line 74: I'd pipe this through a [queue](https://docs.python.org/3/library/asyncio-queue.html) (`get_latest` becomes async as well).
Or even LiFoQueue as that seems to be what you are after here.
Or even, since you are only ever accessing the latest, simply only store the latest.

hitl/loopback.py
```diff
@@ -0,0 +1,148 @@
...
+        """ The actual testing being completed. """
+        interface = await Miniconf.create(args.prefix, args.broker)
+        telemetry = await TelemetryReader.create(args.prefix, args.broker)
+
+        # Test loopback with a static 1V output of the DACs.
+        await test_loopback(interface, telemetry, 1.0)
+
+        # Repeat test with AFE = 2x
+        print('Configuring AFEs to 2x input')
+        await interface.command('afe/0', "G2", retain=False)
```
jordens, line 136: The setup should most of the mqtt settings, not just the non-default ones. And I would inline `test_loopback` as there seems to be no clear distinctions which settings are in there and which are here.

hitl/loopback.py
```diff
@@ -0,0 +1,148 @@
...
+    """ Main program entry point. """
+    parser = argparse.ArgumentParser(description='Loopback tests for Stabilizer HITL testing',)
+    parser.add_argument('prefix', type=str,
+                        help='The MQTT topic prefix of the target')
+    parser.add_argument('--broker', '-b', default='mqtt', type=str,
+                        help='The MQTT broker address')
+
+    args = parser.parse_args()
+
+    async def test():
```
jordens, line 126: I would run this only on one channel pair (adc 0 and dac 0) for now. That provides a lot of coverage already and the second adc/dac pair could/should later be used for some other kind of test (that is not loopback e.g.).

jordens (changes requested):

jordens (approved):

ryan-summers:
bors merge

PR #426, opened 2021-08-05
Feature/streaming hitl

### LINKED ISSUE #386: HITL for streaming
* Set up stream
* Parse a few incoming frames and verify that format and batch ID makes sense
* Disable stream

ryan-summers:
This PR fixes #386 by implementing HITL tests for the data livestream functionalities of Stabilizer.

- [x] This should be merged after https://github.com/quartiq/stabilizer/pull/425


ryan-summers:
bors try

ryan-summers:
bors try

ryan-summers:
@jordens Do you have any ideas in regards to the shared code between this and `scripts/stream_throughput.py`? I'm not a fan of duplicating the code in both places, but I'm not sure what a good way to structure it would be otherwise.

hitl/streaming.py
```diff
@@ -0,0 +1,165 @@
...
+def sequence_delta(previous_sequence, next_sequence):
+    """ Check the number of items between two sequence numbers. """
+    if previous_sequence is None:
+        return 0
+
+    delta = next_sequence - (previous_sequence + 1)
+    return delta & 0xFFFFFFFF
+
+
+class StabilizerStream:
```
ryan-summers, line 45: Should this be moved into a common `stabilizer` python package that we place somewhere else? Then, do some `from stabilizer import StabilizerStream`?

jordens: I'd leave it for now. Pretty sure we'll refactor/port/rewrite it when writing a web app or porting the stream decoding lib to rust.

jordens: But in light of the upcoming CLI tool for the iir filters, yes, make this `/py/stream.py` and add `/py/__init__.py`. The CLI tool would also go there.

Right: `streaming_throughput` and this hitls test should both use a common lib for stream decoding.

hitl/streaming.py
```diff
@@ -0,0 +1,165 @@
...
+    }
+
+    def __init__(self, port):
+        """ Initialize the stream. """
+        self.socket = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
+        self.socket.bind(("", port))
+        self.socket.settimeout(0.5)
+
+
+    def clear(self, duration=5):
```
jordens, line 69: Hmm. Just sleep 5 seconds and then do `while len(self.socket.recv(4096)): pass` to drain?

jordens (changes requested):

hitl/run.sh
```diff
@@ -11,10 +11,11 @@
 set -eux
 
 # Set up python for testing
-python3 -m venv --system-site-packages py
+python3 -m venv --system-site-packages py-venv
 . py/bin/activate
```
jordens, line 15:
```suggestion
. py-venv/bin/activate
```

jordens (changes requested):

ryan-summers:
bors try

ryan-summers:
bors try

jordens (approved):

ryan-summers:
bors merge

PR #434, opened 2021-08-10
Adding optimized signal generator

### LINKED ISSUE #415: Optimize signal generator
The signal generator has been observed to introduce some higher-than-desired overheads in the DSP process routine. Refactor the phase accumulation to minimize execution time.

@jordens has proposed the following refactor after using [godbolt](https://godbolt.org/z/vb3x8rPnM):
```rust
impl core::iter::Iterator for SignalGenerator {
    type Item = i16;

    /// Get the next value in the generator sequence.
    fn next(&mut self) -> Option<i16> {
        let sign = self.phase_accumulator.is_negative();
        self.phase_accumulator = self.phase_accumulator.wrapping_add(
            self.config.frequency_tuning_word[sign as usize]
        );

        let scale = match self.config.signal {
            Signal::Square => if sign { -1 << 15 } else { 1 << 15 },
            Signal::Triangle => (self.phase_accumulator >> 15).abs() - (1 << 15),
        };
        Some(((self.config.amplitude as i32 * scale) >> 15) as _)
    }
}
```

ryan-summers:
This PR fixes #415 by implementing the proposed changes to reduce processing overhead.

Unfortunately, even with these changes, it appears to not be possible to stream at full data rate even when using square waves - there are still losses of ~0.5% of frames.

These changes were run with a signal frequency/amplitude/symmetry sweep from 500Hz-1KHz, Symmetry 0-1, and amplitude 1V -> 2V and output waveform on a scope looked well-formed.

### DIFF
--- a/src/hardware/signal_generator.rs
+++ b/src/hardware/signal_generator.rs
@@ -93,12 +93,12 @@ impl BasicConfig {
                 ftw / self.symmetry
             } else {
                 NYQUIST
-            } as u32,
+            } as i32,
             if symmetry_complement * NYQUIST > ftw {
                 ftw / symmetry_complement
             } else {
                 NYQUIST
-            } as u32,
+            } as i32,
         ];
 
         Ok(Config {
@@ -120,7 +120,7 @@ pub struct Config {
     pub amplitude: i16,
 
     /// The frequency tuning word of the signal. Phase is incremented by this amount
-    pub frequency_tuning_word: [u32; 2],
+    pub frequency_tuning_word: [i32; 2],
 }
 
 impl Default for Config {
@@ -135,7 +135,7 @@ impl Default for Config {
 
 #[derive(Debug)]
 pub struct SignalGenerator {
-    phase_accumulator: u32,
+    phase_accumulator: i32,
     config: Config,
 }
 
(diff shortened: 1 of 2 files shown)


src/hardware/signal_generator.rs
```diff
@@ -174,32 +174,26 @@ impl core::iter::Iterator for SignalGenerator {
...
 
-        let amplitude: i16 = match self.config.signal {
-            Signal::Cosine => (dsp::cossin(phase).0 >> 16) as i16,
+        let scale = match self.config.signal {
+            Signal::Cosine => (dsp::cossin(self.phase_accumulator).0 >> 16),
             Signal::Square => {
-                if phase.is_negative() {
-                    i16::MAX
+                if sign {
+                    -i16::MAX as i32
```
jordens, line 186: Stick to what I had in godbolt. It won't overflow due correct bounds on `amplitude`.

src/hardware/signal_generator.rs
```diff
@@ -174,32 +174,26 @@ impl core::iter::Iterator for SignalGenerator {
...
                 } else {
-                    -i16::MAX
+                    i16::MAX as i32
                 }
             }
-            Signal::Triangle => i16::MAX - (phase.abs() >> 15) as i16,
+            Signal::Triangle => {
+                i16::MAX as i32 - (self.phase_accumulator.abs() >> 15)
```
jordens, line 192: Same here.

jordens (changes requested):

jordens:
Is the frame loss at the sender or at the receiver?

ryan-summers:
I just did a quick Wireshark capture during a drop event, and it looks like the drops may actually be on the Python side here, which is good news.

jordens (approved):

PR #437, opened 2021-08-16
RTIC/Scheduling refactor

### LINKED ISSUE #400: Telemetry unexpectedly stops transmitting
This is the traffic dump from before and after it gets stuck:
image

Then it gets unstuck a couple hours later, reason unknown.
image

Due to sniffing location I'm only seeing the DHCP ACKs.
This only happens roughly once a week. I've seen it four times now.
It might be a timer overflow and/or race. The timescale for becoming unstuck was similar twice.

* I also wonder why all the telemetry messages are fragmented into 63 bytes.
* The MQTT connection survives the day-long stall just fine (even mid-message). But for some reason the device still resets the TCP connection after the first (successful!) telemetry message following the stall.
* No messages from the broker when the interruption start. But when it ends: 

```
1627728003: Socket error on client fls-04-91-62-d9-81-ff-tlm, disconnecting.
1627728003: New connection from 10.34.16.130 on port 1883.
1627728003: New client connected from 10.34.16.130 as fls-04-91-62-d9-81-ff-tlm (p5, c0, k0).
```

* Streaming continues unaffected throughout the interruption.
* It seems to simply mishandle the DHCP stuff. The lease time is 30 m but there is only roughly one renewal per day. When it hangs, then there are no renewals. Also when it clears there are no renewals. In fact, there are no renewals at all. Everything is just discover-off-fer-request-ack every roughly 20 hours.

ryan-summers:
This PR refactors the applications to utilize RTIC 0.6

This PR fixes #400 

TODO:
 - [x] Revert timer back to 10KHz (need to test overflow operation first)
 - [x] Test telemetry scheduling

## Testing
* SystemTimer frequency set to 10MHz, telemetry output observed over ~8 minutes of duration to schedule regularly (tick overflows in 7.15 minutes at this frequency, (2^32 * 100e-9 / 60 = 7.15 mins) No sporadic behavior occurred as the `Instant` wrapped
* SystemTimer verified to schedule telemetry at 10 second intervals with default settings and 10KHz clock


ryan-summers:
bors try

src/hardware/system_timer.rs
```diff
@@ -30,42 +36,22 @@ impl SystemTimer {
...
-    /// for use in calculating time delta, and should not be used for timestamping purposes due to
-    /// roll-over.
-    fn now() -> i32 {
+    // The duration of each tick in seconds.
+    const SCALING_FACTOR: Fraction = Fraction::new(1, 10_000);
+
+    fn try_now(&self) -> Result<Instant<Self>, Error> {
         // Note(unsafe): Multiple interrupt contexts have access to the underlying timer, so care
         // is taken when reading and modifying register values.
         let regs = unsafe { &*hal::device::TIM15::ptr() };
```
jordens, line 57: Can we swap the loop and the CS (both below) to limit the duration of the CS?

src/hardware/system_timer.rs
```diff
@@ -93,24 +79,80 @@ impl rtic::Monotonic for SystemTimer {
...
 
-    /// Get a timestamp correlating to zero time.
-    fn zero() -> i32 {
-        0
+    fn set_compare(&mut self, instant: &Instant<Self>) {
+        let regs = unsafe { &*hal::device::TIM15::ptr() };
+
+        // Note(cs): When operating on the timer registers, we must be atomic to prevent
+        // pre-emption.
+        cortex_m::interrupt::free(|_cs| {
```
jordens, line 107: Same here. Can we push the CS further down to only cover the register accesses (or the `match`)?
I am unsure what atomicicity you mean here. If two tasks are racing to do `set_compare()` they would first need to check that they only ever decrease the `ccr` (make it fire earlier). This is an external constraint independent of the atomicity here but I can't find where they do that.
Now even then, if we let them race, then what atomicity is in danger? Retrieving/computing `now` is already atomic. The racing `ccr` writes, the `cc1ie` sets, and the `pend()` are all atomic and can be done in any order among the tasks with the same result.

src/net/network_processor.rs
```diff
@@ -4,13 +4,15 @@
...
+use crate::hardware::{system_timer::SystemTimer, EthernetPhy};
+use core::convert::TryFrom;
+use rtic::time::{duration::Milliseconds, fraction::Fraction, Clock};
 
 /// Processor for managing network hardware.
 pub struct NetworkProcessor {
     stack: NetworkReference,
     phy: EthernetPhy,
-    clock: CycleCounter,
+    clock: SystemTimer,
```
jordens, line 15: Nice that we can completely free up cyclecounter.

src/net/network_processor.rs
```diff
@@ -72,9 +74,18 @@ impl NetworkProcessor {
     /// An update state corresponding with any changes in the underlying network.
     pub fn update(&mut self) -> UpdateState {
         // Service the network stack to process any inbound and outbound traffic.
-        let now = self.clock.current_ms();
+        let current_ms: Milliseconds<u32> = {
+            // Note(unwrap): The system timer is infallible.
+            let now = self.clock.try_now().unwrap();
+
+            // Note(unwrap): The underlying timer operates at a tick frequency greater than 1 time
+            // per millisecond. Therefore, at the maximum timer tick count, the duration since the
```
jordens, line 82:
```suggestion
            // per millisecond (see the assertion in `new()`). Therefore, at the maximum timer tick count, the duration since the
```

src/hardware/system_timer.rs
```diff
@@ -30,42 +36,22 @@ impl SystemTimer {
     /// * `timer` - The hardware timer used for implementing the RTIC monotonic.
     pub fn initialize(mut timer: hal::timer::Timer<hal::device::TIM15>) {
         timer.pause();
-        // Have the system timer operate at a tick rate of 10KHz (100uS per tick). With this
+        // Have the system timer operate at a tick rate of 10 KHz (100 uS per tick). With this
```
jordens, line 39:
```suggestion
        // Have the system timer operate at a tick rate of 10 KHz (100 µs per tick). With this
```

jordens (changes requested):

src/bin/dual-iir.rs
```diff
@@ -29,162 +30,171 @@
...
-    /// `signal_generator/<n>`
-    ///
-    /// * <n> specifies which channel to configure. <n> := [0, 1]
-    ///
-    /// # Value
-    /// See [signal_generator::BasicConfig#miniconf]
-    signal_generator: [signal_generator::BasicConfig; 2],
-}
+#[rtic::app(device = stabilizer::hardware::hal::stm32, peripherals = true, dispatchers=[DCMI, JPEG, SDMMC])]
+mod app {
```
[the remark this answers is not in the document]
ryan-summers: I tried this originally and it was throwing errors. However, it was likely due to proc-macro shenanigans, as it's working now. I've updated it - I'm a fan of this approach as well.

src/hardware/system_timer.rs
ryan-summers: I actually think we want to explicitly ignore a potential race on `set_compare()` - it's RTIC's job to handle scheduling. If two `set_compare()`s override eachother in different contexts, I'd consider that an RTIC bug. I don't think we can actually protect against that using atomic access, so I'm unsure if we need the critical sections at all.

jordens: Agreed. But this also means we just can't use `set_compare()` in `NetworkProcessor` (in the future).

jordens (approved):

src/hardware/system_timer.rs
ryan-summers: I don't think we ever want to use `set_compare()` - this is an API explicitly provided for the RTIC scheduling purposes. Definitely worth noting though - I'll add some docs.

PR #440, opened 2021-08-18
process: rtic 0.6 escalates link_section attrs

jordens:
https://github.com/rtic-rs/cortex-m-rtic/issues/462


ryan-summers:
Reviewd RTICs updates, and it looks like this should accomplishw hat we want (link section attribute applied to RTIC-generated task, instead of calling from RAM into ITCM. Should be a (mild) performance boost as well

bors r+

Issue #449, opened 2021-09-29
mosquitto.conf fails with mosquitto v2.0.12


svrotter:
When I tried to get mosquitto running in docker on my new machine (win10), I encountered the following issue:
The command proposed in the Stabilizer getting started guide
`docker run -p 1883:1883 --name mosquitto -v ${pwd}/mosquitto.conf:/mosquitto/config/mosquitto.conf -v /mosquitto/data -v /mosquitto/log eclipse-mosquitto:2`
exits with error in mosquitto.conf (see below).
![mosquitto-error](https://user-images.githubusercontent.com/85298112/135326340-338da40b-a559-45cf-90ce-fd07757257ac.PNG)

Turns out, mosquitto is now in v2.0.12 which seems to have some problems with the config file, unlike v2.0.11 as the older version by using
`docker run -p 1883:1883 --name mosquitto -v ${pwd}/mosquitto.conf:/mosquitto/config/mosquitto.conf -v /mosquitto/data -v /mosquitto/log eclipse-mosquitto:2.0.11`
exited without errors.

jordens:
Works fine here. Sure the conf file is not corrupted (i.e. windows line-endings)? I can see the closing parenthesis at the beginning of the line.

```
rj@sherry:~/src/stabilizer (master)$ docker run -v `pwd`/mosquitto.conf:/mosquitto/config/mosquitto.conf eclipse-mosquitto:2
Unable to find image 'eclipse-mosquitto:2' locally
2: Pulling from library/eclipse-mosquitto
a0d0a0d46f8b: Pull complete
dfd7434e63cc: Pull complete
0923389132a8: Pull complete
Digest: sha256:ce08d3fe69d4170cea2426739af86ac95e683f01dd2c4141da661983a2401364
Status: Downloaded newer image for eclipse-mosquitto:2
1632942900: mosquitto version 2.0.12 starting
1632942900: Config loaded from /mosquitto/config/mosquitto.conf.
1632942900: Opening ipv4 listen socket on port 1883.
1632942900: Opening ipv6 listen socket on port 1883.
1632942900: mosquitto version 2.0.12 running
```

svrotter:
You are right about the closing parenthesis. But that was just the try for the screenshot and I tried it multiple times, on multiple consoles with different commands (copy pasted, typed, with pwd, with direct path specification).
Conf file corruption shouldn't be the reason because I used the same config file for v2.0.11 and v2.0.12 and one command worked and the other didn't.
[...]

Issue #339, opened 2021-04-16
Documentation, Manual, CLI


jordens:
Just found https://github.com/airwoodix/sinara-stabilizer-notes from @airwoodix.

* IMO `filter_design.py` can be polished up a little and upstreamed here as a first-class user facing CLI for dual-iir. It's also a nice show case for rough edges or missing pieces in the API.
* The documentation should be integrated into the different components (i.e. Usage/dual-iir -> `filter_design` and and Preparation -> README or other docs structure rustdoc?)

jordens:
User manual and infrastructure added as of #401.
Remaining:

* [ ] Hardware setup (powering, PoE etc)
* [ ] More detailed MQTT setup (let's try [docker](https://hub.docker.com/_/eclipse-mosquitto) plus our config file)
* [ ] Some nicer library for IIR parameters

jordens:
@svrotter You took notes on a couple bugs in the current manual (divergences between docs and code) and missing/tricky bits in the setup instructions, especially on windows. Could you post them here?

svrotter:
We found some issues while trying to set up Stabilizer for local use on my Windows 10 machine with the Getting Started Guide. I try to give a brief overview of what we encountered:

1. windows cmd and powershell don't seem to understand the BROKER=X part of the commands in 'Update Stabilizer' section but my MINGW64 bash apparently does
2. windows cmd and powershell are fine with cargo objcopy --release --bin dual-iir -- -O binary dual-iir.bin` in section 'Uploading firmware' but MINGW64 chokes when using the "`" at the end of the command
3. the docker run ... command in 'Set up the Broker' section also behaves differently on different tools. While powershell acts likes expected, cmd does not allow pwd (chdir could do the trick) and MINGW64 starts the path with /c/... which corrupts the command

We wanted to set everything up for local use, as I don't have access to "10.34.16.10" as a broker. I try to give a quick rundown of what we needed to do in order to make it work. Forgive me, if my terminology may be a bit inaccurate here but networking is really not my strength.
1. Activate vEthernet (WSL) network adapter sharing: Go to vEthernet (WSL) network adapter Properties -> Sharing -> activate "Allow other network users to connect through this computer's Internet connection" -> choose "Ethernet" as Home networking connection. There is no need to have a specific IP address here
[...]

PR #452, opened 2021-10-08
Documentation updates

ryan-summers:
This PR fixes a variety of issues noted in https://github.com/quartiq/stabilizer/issues/339#issuecomment-909457087

This PR also fixes https://github.com/quartiq/stabilizer/issues/449 by adding a `.gitattributes` file that forces the mosquitto config file to be checked out with LF line endings. It appears more recent versions of mosquitto require this.

### DIFF
--- a/book/README.md
+++ b/book/README.md
@@ -13,8 +13,7 @@ directory, and then serve the book:
 ```
 cargo doc --no-deps -p miniconf -p stabilizer -p idsp -p ad9959
 mv target/thumbv7em-none-eabihf/doc book/src/firmware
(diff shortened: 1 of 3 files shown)


svrotter:
I think this should fix most of the issues I pointed out in #339 regarding the getting started section.
As @jordens pointed out in the matrix chat he doesn't consider the setup for a broker on the local machine as relevant at his point, but is there a reason that you didn't touch the miniconf installation in the usage section?

I think we also had some problems there, but @jordens could have to say more about it than me

PR #457, opened 2021-11-02
pounder: warn only if fifo more than half full

jordens:
In regular operation when the process task frequently queues large ~14 byte write,
and if there is a miniconf update, the second will see a non-empty fifo.
It only becomes problematic if this would lead to
significant stalls. As long as updates are smaller than half the fifo
level, this is benign.

### DIFF
--- a/src/hardware/pounder/dds_output.rs
+++ b/src/hardware/pounder/dds_output.rs
@@ -33,8 +33,8 @@
 ///! # Limitations
 ///!
 ///! The QSPI output FIFO is used as an intermediate buffer for holding pending QSPI writes. Because
-///! of this, the implementation only supports up to 16 serialized bytes (the QSPI FIFO is 4 32-bit
-///! words wide) in a single update.
+///! of this, the implementation only supports up to 16 serialized bytes (the QSPI FIFO is 8 32-bit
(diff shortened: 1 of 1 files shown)


src/hardware/pounder/dds_output.rs
```diff
@@ -113,7 +113,8 @@ impl DdsOutput {
         // fashion.
         let regs = unsafe { &*hal::stm32::QUADSPI::ptr() };
 
-        if regs.sr.read().flevel() != 0 {
+        // Warn if the fifo is still at least half full.
+        if regs.sr.read().flevel().bits() >= 16 {
```
ryan-summers, line 117: 16 bytes is indeed half of the FIFO, which docs say is 32 bytes wide. However, thoughout this file, there are references to the FIFO being 4 32-bit words, which is only 16 bytes. Can you please correct the references in this file to reference 8 32-bit words wide, or 32 bytes?

ryan-summers (changes requested):

src/hardware/pounder/dds_output.rs
jordens: done

PR #460, opened 2021-11-08
Updating stabilizer streaming API to decode, use async

ryan-summers:
This PR refactors Stabilizer's `stream` package to efficiently convert a stream frame into a set of traces. It also refactors the API to expose a single `StreamFrame` object that the traces can be extracted from, so the smallest logical unit is no longer a single "batch".

As a result, the concept of "batch" has been entirely removed from the end-user API.

This PR was used with a python-based stream viewer to ensure that the stream could be deserialized efficiently.


hitl/streaming.py
```diff
@@ -80,27 +80,31 @@ async def test():
...
-        assert total_batches, 'Stream did not receive any frames'
-        stream_efficiency = 1.0 - (dropped_batches / total_batches)
+            num_samples = len(frame.traces[list(frame.traces.keys())[0]])
+
+            total_samples += num_samples + num_dropped
+
+            last_sequence = frame.sequence_number + num_samples - 1
+
+        assert total_samples, 'Stream did not receive any frames'
+        stream_efficiency = 1.0 - (dropped_samples / total_samples)
```
jordens, line 103: This all looks a bit convoluted still. It should use the same logic and preferably code with: https://github.com/quartiq/stabilizer/blob/master/scripts/stream_throughput.py

jordens (changes requested):

py/stabilizer/stream.py
```diff
@@ -4,11 +4,54 @@
...
+from . import DAC_VOLTS_PER_LSB, ADC_VOLTS_PER_LSB
+
+StreamFrame = collections.namedtuple('StreamFrame', ['sequence_number', 'format', 'traces'])
+"""
+A single frame received over the livestream interface.
+
+# Args
+* `sequence_number` - The starting sample sequence number.
+* `format` - The format code of the frame.
+* `traces` - A dictionary of traces, each entry of which contain the raw data associated with the
```
jordens, line 28: I don't think a generic frame payload will always be `traces`.

ryan-summers: The main thought here is that we're using the stream to visualize time-series data. What are some use cases where we don't anticipate the stream frame to have that type of data?

hitl/streaming.py
ryan-summers: The intent here is that we're interating across each received stream frame instead of each batch - I have forgotten to update `stream_throughput.py`, but that script would no longer work with these changes.

py/stabilizer/stream.py
```diff
@@ -4,11 +4,54 @@
...
+INT16_MIN = np.iinfo(np.int16).min
+
+
+def format_adc_dac_data(buf: ByteString, batch_size: int):
+    """ Format ADC/DAC data into buffers. """
+    frame = np.frombuffer(buf, '<i2')
+
+    frame = frame.reshape(-1, 4, batch_size).swapaxes(0, 1).reshape(4, -1)
+
+    return {
```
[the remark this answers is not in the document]
ryan-summers: Yeah, it looks like Bokeh has a `y-scale` and `x-scale` that we can apply to transform the data into the coordinate grid. In that case, I believe all we actually have to do is convert the DAC codes to i16 by adding INT16_MIN

py/stabilizer/stream.py
```diff
@@ -72,21 +115,34 @@ def read_frame(self):
...
+        return len(readable) > 0
 
-        frame_format = self.FORMAT[format_id](batch_size)
 
-        batch_count = int(len(buf) / struct.calcsize(frame_format))
+    async def next_frame(self) -> StreamFrame:
+        """ Get the next frame asynchronously. """
+        while True:
+            while not self.frame_available():
+                await asyncio.sleep(0)
```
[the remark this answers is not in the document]
ryan-summers: While I agree, the protocol for actually getting an async UDP socket going is extremely involved (See https://gist.github.com/vxgmichel/e47bff34b68adb3cf6bd4845c4bed448), and I was avoiding it for clarity/simplicity. Happy to go that route though

py/stabilizer/stream.py
```diff
@@ -40,28 +83,28 @@ def __init__(self, port, timeout=None):
...
-        try:
-            while self.socket.recv(4096):
-                pass
-        except socket.timeout:
-            pass
 
+    def add_formatter(self, code: int,
+                      factory: Callable[[ByteString, int], Mapping[str, List[float]]]):
+        """ Add a formatter for recognizing a specific stream format. """
+        assert code not in self._formatters, f'Formatter for {code} already exists'
```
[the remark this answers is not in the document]
ryan-summers: Can you elaborate on what you'd prefer? As far as I'm aware, this is an assert on a coding error (user trying to add a formatter for the same code twice)

py/stabilizer/stream.py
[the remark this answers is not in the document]
jordens: I did that a while ago for a different protocol and it's really not that hard. It's just [`create_datagram_endpoint`](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.create_datagram_endpoint) with a very slim protocol (which is probably exactly what we want). Most of the stuff in that gist is not needed here.

py/stabilizer/stream.py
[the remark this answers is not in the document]
jordens: Sure. But then you don't need to bother constructing a user-friendly message.

py/stabilizer/stream.py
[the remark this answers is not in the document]
jordens: Or in this case for our ADC-DAC format, leave the ADC in i16 and the DAC in u16. That may require a record array which is a bit annoying.
Doing the `^ 0x8000` is simplest.

hitl/streaming.py
jordens: I had the impression that we could have merged the code. The `stream_throughput` seemed latest and greatest.

py/stabilizer/stream.py
jordens: I'm a bit torn about where the payload parsing should happen. I'll think about it.

ryan-summers: There's certainly something to be said about offloading the parsing of the frame data... The downside if we don't unify it is that we would potentially be copying code in various places. Alternatively, we could just provide some parsing functions and not explicitly parse the frame, and just provide the raw data that the end user can pass to the provided parsing functions. That would definitely simplify the `formatter` API.

Downside of that approach is that we have to carry around the concept of "batches" until the frame has been deserialized.

jordens:
@ryan-summers I refactored everything to be asyncio with compact code. I can't request a review from you since you are the author of the PR. Feel free to comment.

bors try

jordens:
bors try

jordens:
bors try

jordens:
bors try

jordens:
bors try

py/stabilizer/stream.py
```diff
@@ -1,92 +1,159 @@
...
-        for offset in range(batch_count):
-            data = struct.unpack_from(frame_format, buf)
-            buf = buf[struct.calcsize(frame_format):]
-            yield (sequence_number + offset, data)
+        try:
+            parser = self.parsers[header.format_id]
+        except KeyError:
+            logger.warning("No parser for format %s, ignoring", header.format_id)
+            return
+        frame = parser(header, data[self.header_fmt.size:])
```
ryan-summers, line 94: My only qualm with this approach is that this implies that the user gets a variable type back from the `StabilizerStream.queue`. Essentially, there's no guarantee on what the type is, and the type changes based on the format_id of the received frame.

In Rust, we'd solve this by returning some `impl StreamFrame`. In our case, I guess we could use an ABC to guarantee that the returned object implements some common interface?

We need to have some known API so something like the stream visualizer can know what to do for any arbitrary stream frame received.

jordens: I'm not sure there is a common meaningful `impl StreamFrame` trait. Apart from the header data, what's guaranteed to be common among stream frame formats?

The user can use any of the following approaches:

* adjust `parsers` so they are sure to only receive what they want
* check the `.header.format_id` attribute of the object they get form the queue
* use `is AdcDac` to check for class
* use duck typing to check for behavioral equivalence

jordens (approved):

PR #468, opened 2021-12-07
Indicating stabilizer status via MQTT will

### LINKED ISSUE #302: MQTT: Status indication via retained will
It would be nice to provide a general status/connection health indication via a will specified during connection, causing a retained message to be written when the broker loses the device connection, e.g. `dt/sinara/stabilizer/<…>/status` `{"online": false, "message": "Stabilizer connection lost"}` or something like that, which would be cleared (or set to an positive message instead) during startup.

ryan-summers:
This PR fixes #302 by utilizing a miniconf version that uses the <prefix>/alive topic to indicate life status of stabilizer.

This branch was run locally and ethernet was pulled. It was observed that after 2 and a half minutes, the `<prefix>/alive` topic was published as 0.

We may want to reduce the keep-alive interval, as it appears the will is only published after 2.5x the keepalive interval has elapsed.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -224,8 +224,7 @@ dependencies = [
 [[package]]
 name = "derive_miniconf"
 version = "0.1.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a09410a9472600744ed45ab245fc56bce9c44a422434cf91cad613ac5857f93c"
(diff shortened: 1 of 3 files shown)


jordens (approved):

ryan-summers:
bors merge

ryan-summers:
bors merge

PR #471, opened 2021-12-10
dual_iir_configuration: fix pid parameter scaling

jordens:


scripts/dual_iir_configuration.py
```diff
@@ -188,7 +188,7 @@ def calculate_pid_coefficients(sampling_period, args):
         [1, -2, 1]
     ]
 
-    digital_conversion = (2 * pi) / sampling_period
+    digital_conversion = 2 * pi * sampling_period
```
ryan-summers, line 191: I'm confused. I thought that `omega_s = 2 pi / T`.

Is this conversion type not actually `omaga_s`?

jordens: In the definitions of the primer, `kd*2*pi` is the D gain at the sample frequency. Or `ki/tau` is the I gain "slope". Then
`kd*2*pi*f/fs` and `ki/(2*pi*f/fs)` are gains at f.

ryan-summers: Ah gotcha. That makes more sense, thanks for the clarification. This is a scaling factor to convert some given `f` input.

PR #473, opened 2021-12-10
rj/py dependency

jordens:
- py, hitl: miniconf dependency


ryan-summers:
bors cancel

ryan-summers (changes requested):
Can you also update the `book/src/usage.md` document to point users to installing the stabilizer package instead of using `requirements.txt`?

PR #475, opened 2021-12-13
Adding settings republish support

ryan-summers:
This PR updates Stabilizer to use miniconf 0.3, which automatically publishes the settings structure on startup over MQTT.

TODO:
- [x] Release `minimq` as a patched 0.5.1 version to incorporate handling of full network buffering.
- [x] Release idsp to support new miniconf versions, remove patch

### DIFF
--- a/src/bin/dual-iir.rs
+++ b/src/bin/dual-iir.rs
@@ -50,7 +50,6 @@ use stabilizer::{
     net::{
         data_stream::{FrameGenerator, StreamFormat, StreamTarget},
         miniconf::Miniconf,
-        serde::Deserialize,
         telemetry::{Telemetry, TelemetryBuffer},
         NetworkState, NetworkUsers,
     },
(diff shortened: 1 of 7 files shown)


src/bin/dual-iir.rs
```diff
@@ -68,7 +67,7 @@ const BATCH_SIZE: usize = 8;
 // 128, there is 1.28uS per sample, corresponding to a sampling frequency of 781.25 KHz.
 const SAMPLE_TICKS_LOG2: u8 = 7;
 
-#[derive(Clone, Copy, Debug, Deserialize, Miniconf)]
```
jordens, line 71: Just a note: I was about to ask why we don't eagerly derive things. But then I looked closer and the serde traits are actually not "common traits" [Types eagerly implement common traits (C-COMMON-TRAITS)](https://rust-lang.github.io/api-guidelines/interoperability.html#types-eagerly-implement-common-traits-c-common-traits)

jordens (approved):

jordens:
Ah. While this was pretty small code-wise, it is big on user impact. Should definitely get some mention in the CHANGELOG. @ryan-summers

PR #499, opened 2022-02-02
Updating RTIC to 1.0

ryan-summers:
This PR updates RTIC to 1.0

It also makes some modifications to the `Monotonic` implementation to address issues with `embedded-time` not being able to satisfy RTIC's requirements. `fugit` is now used for the rtic::Monotonic implmentation and `embedded-time` is used for libraries.

TODO:
- [x] Tested on Stabilizer hardware

### DIFF
--- a/src/bin/dual-iir.rs
+++ b/src/bin/dual-iir.rs
@@ -33,8 +33,8 @@ use core::sync::atomic::{fence, Ordering};
 
 use mutex_trait::prelude::*;
 
+use fugit::ExtU32;
 use idsp::iir;
-use rtic::time::duration::Extensions;
 use stabilizer::{
     hardware::{
         self,
@@ -441,7 +441,7 @@ mod app {
 
         // Schedule the telemetry task in the future.
         telemetry_task::Monotonic::spawn_after(
-            (telemetry_period as u32).seconds(),
+            (telemetry_period as u32).secs(),
         )
         .unwrap();
     }
@@ -451,7 +451,7 @@ mod app {
         c.shared
             .network
             .lock(|network| network.processor.handle_link());
-        ethernet_link::Monotonic::spawn_after(1u32.seconds()).unwrap();
+        ethernet_link::Monotonic::spawn_after(1u32.secs()).unwrap();
     }
 
     #[task(binds = ETH, priority = 1)]
--- a/src/hardware/system_timer.rs
+++ b/src/hardware/system_timer.rs
@@ -6,11 +6,9 @@
 ///! than the CPU clock. This allows for longer scheduling periods with less resolution. This is
 ///! needed for infrequent (e.g. multiple second) telemetry periods.
 use hal::prelude::*;
-use rtic::{
-    time::{clock::Error, fraction::Fraction, Clock, Instant},
-    Monotonic,
-};
 use stm32h7xx_hal as hal;
+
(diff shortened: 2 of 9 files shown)


ryan-summers:
bors try

jordens:
Also needs a changelog entry at least for the MSRV change.

jordens:
Also tested on `fls` and works fine.

src/hardware/system_timer.rs
```diff
@@ -103,15 +116,18 @@ impl Monotonic for SystemTimer {
...
+    fn now(&mut self) -> Self::Instant {
+        Self::Instant::from_ticks(self.get_ticks())
+    }
+
+    fn set_compare(&mut self, instant: Self::Instant) {
         let regs = unsafe { &*hal::device::TIM15::ptr() };
 
-        let now = self.try_now().unwrap();
-        match now.checked_duration_until(instant) {
+        match instant.checked_duration_since(self.now()) {
```
jordens, line 126: Instead of that entire match I would just:

```rust
let ticks = instant
    .checked_duration_since(self.now())
    .map_or(0, |t| t.ticks())
    .max(1)
    .min(0xffff);
regs.ccr1.write(|w| w.ccr().bits(ticks as _));
regs.dier.modify(|_, w| w.cc1ie().set_bit());
```

src/bin/dual-iir.rs
```diff
@@ -33,8 +33,8 @@ use core::sync::atomic::{fence, Ordering};
 
 use mutex_trait::prelude::*;
 
+use fugit::ExtU32;
```
jordens, line 36:
```suggestion
pub use fugit::ExtU32;
```

ryan-summers: Why does this need to be `pub`? This is in a final `bin` application, so there shouldn't be anyone importing it as a library

jordens: Right. I thought I was looking at `system_timer`.

src/hardware/system_timer.rs
ryan-summers: That match is actually not equivalent to what's written here. We load the CCR1 (compare-capture-reg 1) with the u16 compare timestamp - the `map_or` function maps the Option to the tick delta between `instant` and `now`, which is not equivalent to the timestamp that we want to wake up at, but rather the time difference between the two instants.

We only use the delta-tick calculated value to determine whether or not the `instant` will occur within a single overflow of the timer. If it does, we just truncate the `instant` to a u16 tick count (since we guarantee that the 16 MSBs are irrelevant via the earlier check) and store that in CCR1. If it doesn't, we store u16::MAX in CCR1 to re-evaluate at the next timer overflow.

I think the correct chain would be:
```rust
let ticks = if instant.checked_duration_since(self.now()).unwrap_or(1) < (1 << 16) {
    instant.ticks() as u16
} else {
    u16::MAX
}
```

ryan-summers: However, after staring at the code for a while, I think this can be simplified further. Give me a moment.

ryan-summers: In any case, given that we plan on replacing this code soon with `systick-monotonic`, I'd recommend we leave it as-is.

jordens: Ah. Yes. You have TIM15 free running and interrupt on match, not interrupt on wrap and reload.

jordens (approved):

PR #502, opened 2022-02-04
use systick-monotonic

### LINKED ISSUE #501: use `systick-monotonic`
I made https://github.com/rtic-rs/dwt-systick-monotonic/pull/6 and tested on fls (just for the RTIC monotonic, not the SystemTimer Clock for NAL etc).
But I think we should just use `systick-monotonic` at 1 kHz (which also works fine). wrapping `monotonics::now().ticks()` might be all we need for `Clock`.
https://github.com/rtic-rs/systick-monotonic

jordens:
Closes #501 

* [x] consolidate CPU frequency and Systick rate usage
* [x] Test


SingularitySurfer:
Here 
https://github.com/quartiq/stabilizer/blob/45ac73b252637fb6d9312925c896afdc4c1a3f9b/src/net/mod.rs#L64-L69
and here
https://github.com/quartiq/stabilizer/blob/45ac73b252637fb6d9312925c896afdc4c1a3f9b/src/net/telemetry.rs#L98-L101
the clock is missing in the docstring.

jordens:
bors try

SingularitySurfer:
Just noticed that here https://github.com/quartiq/stabilizer/blob/39f2133f6a803f3eaf9924c00a1116815aaf4af1/src/net/network_processor.rs#L22
the clock is there where it shouldn't be, even tho that's older code.

jordens:
bors try

src/hardware/mod.rs
```diff
@@ -44,17 +43,22 @@ const RX_DESRING_CNT: usize = 4;
...
-    system_timer::SystemTimer,
+    SystemTimer,
 >;
 
 pub type EthernetPhy = hal::ethernet::phy::LAN8742A<hal::ethernet::EthernetMAC>;
 
+/// System timer (RTIC Monotonic) tick frequency
+pub const HZ: u32 = 1_000;
+pub type Systick = systick_monotonic::Systick<HZ>;
+pub type SystemTimer = mono_clock::MonoClock<u32, HZ>;
```
ryan-summers, line 60:
```suggestion
pub const MONOTONIC_FREQUENCY: u32 = 1_000;
pub type Systick = systick_monotonic::Systick<MONOTONIC_FREQUENCY>;
pub type SystemTimer = mono_clock::MonoClock<u32, MONOTONIC_FREQUENCY>;
```

I'm not a fan of just `HZ` due to its vagueness

src/hardware/setup.rs
```diff
@@ -255,24 +258,19 @@ pub fn setup(
...
-    {
-        let tim15 =
-            device
-                .TIM15
-                .timer(10.khz(), ccdr.peripheral.TIM15, &ccdr.clocks);
-        system_timer::SystemTimer::initialize(tim15);
-    }
+    let systick = Systick::new(core.SYST, ccdr.clocks.sysclk().0);
+
+    core.SCB.enable_icache();
```
ryan-summers, line 273: Is there any particular reason we're enabling the ICACHE earlier? Just to speed up INIT? If there's an important reason, we should document it. Otherwise, this is fine.

src/hardware/setup.rs
```diff
@@ -761,10 +767,7 @@ pub fn setup(
             data
         };
 
-        let mut stack = smoltcp_nal::NetworkStack::new(
-            interface,
-            system_timer::SystemTimer::default(),
-        );
+        let mut stack = smoltcp_nal::NetworkStack::new(interface, clock);
```
ryan-summers, line 770: I'm not a fan of constructing the clock just to pass it in here. Could we construct the clock locally?

I dislike having to construct the `clock` externally and passing it in - it's just a strange API.

```rust
let clock = SystemTimer::new(|| crate::app::monotonics::now().ticks() as u32);
```
Not sure if that will work though.

src/bin/dual-iir.rs
```diff
@@ -243,21 +254,25 @@ mod app {
...
         local.dacs.0.start();
         local.dacs.1.start();
 
-        // Start sampling ADCs.
-        stabilizer.adc_dac_timer.start();
+        // Spawn a settings update for default settings.
+        settings_update::spawn().unwrap();
+        telemetry::spawn().unwrap();
+        ethernet_link::spawn().unwrap();
+        start::spawn_after(100.millis()).unwrap();
```
ryan-summers, line 267: Any reason why we need to start the ADC sampling timer so late? Is it the `init::Monotonics` construction? If so, we can perform that call before starting the timer. I dislike having this additional complexity.

ryan-summers (changes requested):

src/hardware/setup.rs
jordens: It moved around a couple times. No particular reason. It should just be after the ITCM load.

src/hardware/setup.rs
jordens: Doesn't work because the binaries are not part of the library crate.

src/bin/dual-iir.rs
jordens: That and a lot of other things I tried all didn't help. I assume it's some unfortunate sideffect of the other tasks being scheduled all at once here and the copious `interrupt::free` that the timer queue does, and maybe cold caches plus the increased timer ISR frequency as well. I couldn't figure it out immediately without instrumenting and tracing the entire init to idle transition.
A nice project would be to try `rtic-scope` on this. That would provide a lot of insight into the underlying timing of the various timers and delays in that area.

PR #503, opened 2022-02-07
rj/bin homogenize

jordens:
This is mostly a cleanup to variable and constant names in the applications. But it fixes the RTT timeout issue on my machine and also sorts the initialization a bit differently to start DMA later and provide less chance for timeouts.

- dual-iir/lockin: harmonize code, naming and comments
- move rtt init early so debugger doesn't time out
- bins: renames, align code

### DIFF
--- a/src/bin/lockin.rs
+++ b/src/bin/lockin.rs
@@ -1,7 +1,7 @@
 //! # Lockin
 //!
 //! The `lockin` application implements a lock-in amplifier using either an external or internally
-//! generated
+//! generated reference.
 //!
 //! ## Features
 //! * Up to 800 kHz sampling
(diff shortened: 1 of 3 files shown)


src/bin/lockin.rs
```diff
@@ -1,6 +1,6 @@
 //! # Lockin
 //!
-//! THe `lockin` application implements a lock-in amplifier using either an external or internally
+//! The `lockin` application implements a lock-in amplifier using either an external or internally
 //! generated
```
ryan-summers, line 4: This comment is an incomplete sentence - mind finishing it?

ryan-summers (approved):

PR #507, opened 2022-02-09
add sine to signal generator

svrotter:
Added sine function to signal generator to have the full variety of available signals

### DIFF
--- a/src/hardware/signal_generator.rs
+++ b/src/hardware/signal_generator.rs
@@ -4,6 +4,7 @@ use serde::{Deserialize, Serialize};
 /// Types of signals that can be generated.
 #[derive(Copy, Clone, Debug, Deserialize, Serialize, Miniconf)]
 pub enum Signal {
+    Sine,
     Cosine,
     Square,
     Triangle,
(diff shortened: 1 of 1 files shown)


jordens:
What does this offer?
Isn't that already possible by adding -0.25 turns phase_offset?

svrotter:
It is.
But sometimes life is easier when you also give things the names that are easier.
If I'd demodulate in I&Q for example and need to setup LUTs, it is way more intuitive to set one up with a sine and one with a cosine, especially when they need to be phase shifted according to the signal that should be demodulated.

jordens:
For demodulation you want to use the `cossin` stuff directly, or add complex output to the signal generator. Using two signal generators to do it is inefficient.

jordens:
I still don't see when and why this would help. If you want names, then you can make a `const SINE_PHASE_OFFSET: f32 = -0.25;` and use that. Adding more code that does the same just with different names just means more code to test, check, maintain, document.

svrotter:
The IQ thing was just the first thing that came to my mind. There are surely other situation I'd like to have that. At the end I'm replicating what I get offered by the AWGs I personally use and like.
If you don't want to have a naming that makes the device more user friendly outside of the rust code in interfaces like mine, where a `const SINE_PHASE_OFFSET` doesn't really help, it's fine. 
I'm not passionate about it, so just reject the request

jordens:
As mentioned, a good design for IQ would give you IQ, not at twice the price I and Q separately. That doesn't appear to be a good argument. And having just looked at the first AWG that came to my mind, it's already a counterexample. It doesn't have two output options for sine and cosine but just one (plus the obvious phase shift). I'm always happy to make this more user friendly but I don't see where and how this achieves it.

PR #509, opened 2022-02-09
Add static IP address as a build feature

svrotter:
In some cases it might be advantageous to set a static IP address for Stabilizer e.g. if it is used in an embedded environment that doesn't support DHCP or troubleshooting problems with DHCP negotiation.
This PR add this functionality as a build feature.

### DIFF
--- a/src/hardware/setup.rs
+++ b/src/hardware/setup.rs
@@ -23,11 +23,20 @@ const NUM_TCP_SOCKETS: usize = 4;
 const NUM_UDP_SOCKETS: usize = 1;
 const NUM_SOCKETS: usize = NUM_UDP_SOCKETS + NUM_TCP_SOCKETS;
 
+const IPADDR: [u8; 4] = [192, 168, 137, 17];
+
+// Omit DHCP socket, if static IP is used
+#[cfg(feature = "static_ip")]
+const NUM_DHCP_SOCKETS: usize = 0;
+#[cfg(not(feature = "static_ip"))]
+const NUM_DHCP_SOCKETS: usize = 1;
+
 pub struct NetStorage {
     pub ip_addrs: [smoltcp::wire::IpCidr; 1],
 
     // Note: There is an additional socket set item required for the DHCP socket.
-    pub sockets: [smoltcp::iface::SocketStorage<'static>; NUM_SOCKETS + 1],
+    pub sockets: [smoltcp::iface::SocketStorage<'static>;
+        NUM_SOCKETS + NUM_DHCP_SOCKETS],
     pub tcp_socket_storage: [TcpSocketStorage; NUM_TCP_SOCKETS],
     pub udp_socket_storage: [UdpSocketStorage; NUM_UDP_SOCKETS],
     pub neighbor_cache:
@@ -84,7 +93,8 @@ impl Default for NetStorage {
             )],
             neighbor_cache: [None; 8],
             routes_cache: [None; 8],
-            sockets: [smoltcp::iface::SocketStorage::EMPTY; NUM_SOCKETS + 1],
+            sockets: [smoltcp::iface::SocketStorage::EMPTY;
+                NUM_SOCKETS + NUM_DHCP_SOCKETS],
             tcp_socket_storage: [TcpSocketStorage::new(); NUM_TCP_SOCKETS],
             udp_socket_storage: [UdpSocketStorage::new(); NUM_UDP_SOCKETS],
         }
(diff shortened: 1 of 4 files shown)


src/hardware/setup.rs
```diff
@@ -719,7 +744,9 @@ pub fn setup(
         .routes(routes)
         .finalize();
 
-        interface.add_socket(smoltcp::socket::Dhcpv4Socket::new());
+        if NUM_DHCP_SOCKETS == 1 {
+            interface.add_socket(smoltcp::socket::Dhcpv4Socket::new());
```
ryan-summers, line 748:
```suggestion
            #[cfg(not(feature = "static_ip"))]
            interface.add_socket(smoltcp::socket::Dhcpv4Socket::new())
```

src/hardware/setup.rs
```diff
@@ -23,11 +23,20 @@ const NUM_TCP_SOCKETS: usize = 4;
...
+const NUM_DHCP_SOCKETS: usize = 0;
+#[cfg(not(feature = "static_ip"))]
+const NUM_DHCP_SOCKETS: usize = 1;
+
 pub struct NetStorage {
     pub ip_addrs: [smoltcp::wire::IpCidr; 1],
 
     // Note: There is an additional socket set item required for the DHCP socket.
-    pub sockets: [smoltcp::iface::SocketStorage<'static>; NUM_SOCKETS + 1],
+    pub sockets: [smoltcp::iface::SocketStorage<'static>;
```
ryan-summers, line 38: I'd just revert these changes. If we have 1 extra storage slot that we don't end up using, it's not a problem (minor memory overhead), but we keep the code a lot cleaner.

ryan-summers (changes requested):

src/hardware/setup.rs
```diff
@@ -23,11 +23,20 @@ const NUM_TCP_SOCKETS: usize = 4;
 const NUM_UDP_SOCKETS: usize = 1;
 const NUM_SOCKETS: usize = NUM_UDP_SOCKETS + NUM_TCP_SOCKETS;
 
+const IPADDR: [u8; 4] = [192, 168, 137, 17];
```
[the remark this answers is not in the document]
ryan-summers: As @jordens noted above, we could use some `env!(STATIC_IP).map(|addr| addr.parse()).unwrap_or(DEFAULT_IP)`

ryan-summers:
It occurs to me that we could actually do this entirely based on the presence of some input environment variable instead of using a cargo-feature. E.g. use some default `UNSPECIFIED` IP if nothing's provided on the `STATIC_IP` variable, otherwise use the provided IP.

src/hardware/setup.rs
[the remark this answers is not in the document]
jordens: Better yet, with that now you can remove the cargo feature and just act on the presence of the environment variable by (a) use that instead of UNSPECIFIED and (b) not adding the dhcp socket. That simplifies the code further and eases user configuration together with the broker IP.

svrotter:
Couldn't set it to `UNSPECIFIED` somehow, it always panics then. Maybe that's my incompetence but setting it to 0.0.0.0 and checking for `is_unspecified ()` worked

src/hardware/setup.rs
```diff
@@ -699,31 +689,25 @@ pub fn setup(
...
+            .parse()
+            .unwrap();
 
         // Note(unwrap): The hardware configuration function is only allowed to be called once.
         // Unwrapping is intended to panic if called again to prevent re-use of global memory.
         let store =
             cortex_m::singleton!(: NetStorage = NetStorage::default()).unwrap();
 
-        store.ip_addrs[0] = ip_addrs;
+        store.ip_addrs[0] = match ip_addrs.is_unspecified() {
```
ryan-summers, line 702: If ip_addrs is already unspecified, why do we need to construct it with a different netmask below? Can't both use the same 24 bit netmask without issue?

src/hardware/setup.rs
```diff
@@ -688,17 +688,26 @@ pub fn setup(
...
+
         // Note(unwrap): The hardware configuration function is only allowed to be called once.
         // Unwrapping is intended to panic if called again to prevent re-use of global memory.
         let store =
             cortex_m::singleton!(: NetStorage = NetStorage::default()).unwrap();
 
-        store.ip_addrs[0] = smoltcp::wire::IpCidr::new(
-            smoltcp::wire::IpAddress::Ipv4(
-                smoltcp::wire::Ipv4Address::UNSPECIFIED,
+        store.ip_addrs[0] = match ip_addrs.is_unspecified() {
```
ryan-summers, line 702:
```suggestion
        store.ip_addrs[0] = smoltcp::wire::IpCidr::new(option_env!("STATIC_IP").unwrap_or("0.0.0.0").parse().unwrap(), 24);
```

ryan-summers:
bors try

src/hardware/setup.rs
svrotter: Didn't follow the suggestion exactly but used 26 bit masks for both specified and unspecified IP cases. Also fixed an `UNSPECIFIED` mistake I made in the previous commit and tested everything on Stabilizer without issues

ryan-summers:
Looking good so far, only further changes:
1. Add an entry to the CHANGELOG
2. Either update the docs in `book/` to mention static IP configuration options or spawn an issue so we can add it later

jordens (approved):
LGTM

PR #512, opened 2022-02-11
release 0.6

jordens:
- ad9959: add metadata
- add ad9959 version

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "stabilizer"
-version = "0.5.0"
+version = "0.6.0"
 resolver = "2"
 authors = [
 	"Robert Jördens <rj@quartiq.de>",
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -2,7 +2,10 @@
 name = "stabilizer"
 version = "0.5.0"
```
ryan-summers, line 3: Need to update this version as well

ryan-summers (changes requested):

Cargo.toml
jordens: Yes. I'll be using `cargo-release` for this which will do the increment in the end just before committing that and publishing.
Next time i'll look into splitting the sequence.

jordens: In fact I'll try that workflow now.

ryan-summers (approved):

PR #532, opened 2022-05-05
hal: v0.12.1

jordens:


jordens:
bors try

jordens:
@nkrackow FYI

jordens:
bors try

jordens:
bors try

ryan-summers (approved):
Wow those IO changes are super clean. I imagine the compiler can actually figure out which AFx you need to convert into based on the accepted type signatures of the driver they the pins are being handed to (and only needs specification in certain cases).

That's super handy and eliminates the mental need to lookup which alternate function is needed. Nice change!

jordens:
I think for things like the AFEs we may actually want to use the type erased pins. Then both AFEs have the same type.

PR #541, opened 2022-05-16
Fixing batch size 1 calculations

### LINKED ISSUE #540: Batch size of 1 is broken
Any attempt to utilize a batch size of 1 will result in an internal panic. Below is a sample of the trace back with optimization significantly reduced to show more detail:
```
INFO - Starting
INFO - EUI48: 04-91-62-d2-a8-6f
INFO - setup() complete
WARN - Network link DOWN
panicked at 'assertion failed: `(left != right)`
  left: `0`,
 right: `0`', /rustc/7737e0b5c4103216d6fd8cf941b7ab9bdbaace7c/library/core/src/slice/mod.rs:906:9
────────────────────────────────────────────────────────────────────────────────
(HOST) WARN  no RAM region appears to contain the stack; cannot determine if this was a stack overflow
stack backtrace:
   0: HardFaultTrampoline
      <exception entry>
   1: lib::inline::__udf
        at ./asm/inline.rs:181:5
   2: __udf
        at ./asm/lib.rs:51:17
   3: cortex_m::asm::udf
        at C:\Users\rsummers\.cargo\registry\src\github.com-1ecc6299db9ec823\cortex-m-0.7.4\src\asm.rs:43:5
   4: rust_begin_unwind
        at src\hardware\mod.rs:90:5
   5: core::panicking::panic_fmt
        at /rustc/7737e0b5c4103216d6fd8cf941b7ab9bdbaace7c/library/core/src/panicking.rs:143:14
   6: core::panicking::assert_failed_inner
   7: core::panicking::assert_failed
        at /rustc/7737e0b5c4103216d6fd8cf941b7ab9bdbaace7c/library/core/src/panicking.rs:182:5
   8: core::slice::<impl [T]>::chunks_exact_mut
        at /rustc/7737e0b5c4103216d6fd8cf941b7ab9bdbaace7c/library/core/src/slice/mod.rs:906:9
   9: dual_iir::app::process::{{closure}}::{{closure}}::{{closure}}
        at src\bin\dual-iir.rs:361:34
  10: stabilizer::net::data_stream::StreamFrame::add_batch
        at src\net\data_stream.rs:172:9
  11: stabilizer::net::data_stream::FrameGenerator::add
```

This appears to be caused by integer division in some of the livestream const calculations that results in a calculated size of 1 * 2 / 4 = 0, which results in slices of zero size being used.

ryan-summers:
This PR fixes #540 by refactoring the stream frames to use `u8` as the underlying type. There was an integer division error when using a batch size of 1 when using `u32` container types, as a single sample cannot completely fill the `u32` container.

We can alternatively use a container type of `u16`, but I don't know if we gain anything by using a larger container, so I just opted for `u8` as it's byte-level representation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 * Minimum supported Rust version bumped to 1.59
 
 ### Fixed
+* Fixed panics when using a batch size of 1 ([#540](https://github.com/quartiq/stabilizer/issues/540))
 
 ## [v0.6.0] - 2022-02-11
 
--- a/src/bin/dual-iir.rs
+++ b/src/bin/dual-iir.rs
@@ -352,7 +352,7 @@ mod app {
 
                     // Stream the data.
                     const N: usize = BATCH_SIZE * core::mem::size_of::<i16>()
-                        / core::mem::size_of::<MaybeUninit<u32>>();
+                        / core::mem::size_of::<MaybeUninit<u8>>();
                     generator.add::<_, { N * 4 }>(|buf| {
                         for (data, buf) in adc_samples
                             .iter()
@@ -361,7 +361,7 @@ mod app {
                         {
                             let data = unsafe {
                                 core::slice::from_raw_parts(
-                                    data.as_ptr() as *const MaybeUninit<u32>,
+                                    data.as_ptr() as *const MaybeUninit<u8>,
                                     N,
                                 )
                             };
--- a/src/bin/lockin.rs
+++ b/src/bin/lockin.rs
@@ -422,7 +422,7 @@ mod app {
 
                 // Stream the data.
                 const N: usize = BATCH_SIZE * core::mem::size_of::<i16>()
-                    / core::mem::size_of::<MaybeUninit<u32>>();
+                    / core::mem::size_of::<MaybeUninit<u8>>();
                 generator.add::<_, { N * 4 }>(|buf| {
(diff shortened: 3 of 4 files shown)


ryan-summers:
bors try

jordens:
IIRC u8 was significantly slower than u32. We should check the assembly of those two approaches.

ryan-summers:
I just did testing with diffferent batch storage sizes (u8. u16, u32) and timed the amount of time to write to the buffers. I don't see a discernible difference in timing, even between batch size = 2 and batch size = 32. Almost all measurements indicate only about ~70-150ns is needed to transfer data. I took a stab at [Godbolt](https://godbolt.org/z/64ErnEhx5) and it looks like the compiler is smart enough to figure out the best storage instructions based on some of the const calculations, but I'm not 100% understanding some of the ASM operations listed there.

ryan-summers:
Scratch those measurements - I didn't have the ethernet cable connected, so the livestream was disabled and the data population was a no-op. Let me re-test

ryan-summers:
Testing was conducted with the stream target set to my local machine. MQTT was connected and the stream was directed:

These changes w/ batch size of 8, the `process()` function takes an average of 6.735 uS

On `master` with a batch size of 8, the `process()` function takes an average of 6.732 uS

I don't see a discernible difference in the `release` configuration here, so I'm comfortable merging as-is.

jordens:
Great. Thanks for checking.
bors r+

jordens (approved):

PR #543, opened 2022-05-23
fix phase increment

nkrackow:
The scaling factor was missing. Took me a bit to get that (1 << 31) is negative.


ryan-summers:
@nkrackow In order to get HITL to run, you can do `bors merge`, and bors will run HITL testing and merge if it is successful. I always forget this workflow myself on Stabilizer, so I just wanted to give you a heads up :)

PR #554, opened 2022-07-11
Nk/driver development

nkrackow:
Add driver bin and adapt setup (to be determined how driver is detected).
Implement LTC2320 driver with one qspi data lane. Tested working using Sampler.
No channel ordering or bitshuffling yet.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,9 +2,9 @@ name: Continuous Integration
 
 on:
   push:
-    branches: [master, staging, trying]
+
   pull_request:
-    branches: [master]
+    branches: [master, driver]
   schedule:
     # UTC
     - cron: '48 4 * * *'
--- a/src/bin/driver.rs
+++ b/src/bin/driver.rs
@@ -455,12 +455,12 @@ mod app {
         .unwrap();
     }
 
-    #[task(binds = TIM7, priority = 2, shared=[ltc2320])]
-    fn ltc2320_conv_done(mut c: ltc2320_conv_done::Context) {
-        c.shared
-            .ltc2320
-            .lock(|ltc| ltc.handle_conv_done_irq().unwrap()); // panic if last QSPI transfer not complete
-    }
+    // #[task(binds = TIM7, priority = 2, shared=[ltc2320])]
+    // fn ltc2320_conv_done(mut c: ltc2320_conv_done::Context) {
+    //     c.shared
+    //         .ltc2320
+    //         .lock(|ltc| ltc.handle_conv_done_irq().unwrap()); // panic if last QSPI transfer not complete
+    // }
 
     #[task(binds = QUADSPI, priority = 2, shared=[ltc2320, ltc2320_data])]
     fn ltc2320_transfer_done(c: ltc2320_transfer_done::Context) {
--- a/src/bin/driver.rs
+++ b/src/bin/driver.rs
@@ -446,9 +446,8 @@ mod app {
     #[task(priority = 2, shared=[ltc2320], local=[ltc2320_conversion_scheduled])]
     fn ltc2320_start_conversion(mut c: ltc2320_start_conversion::Context) {
         c.shared.ltc2320.lock(|ltc| ltc.start_conversion()).unwrap(); // panic if LTC2320 timing is not met
-        *c.local.ltc2320_conversion_scheduled =
-            *c.local.ltc2320_conversion_scheduled
-                + design_parameters::LTC2320_PERIOD.convert(); // update time at which the next conversion is scheduled
+        *c.local.ltc2320_conversion_scheduled +=
+            design_parameters::LTC2320_PERIOD.convert(); // update time at which the next conversion is scheduled
         ltc2320_start_conversion::Monotonic::spawn_at(
             *c.local.ltc2320_conversion_scheduled,
         )
--- a/src/bin/driver.rs
+++ b/src/bin/driver.rs
@@ -438,21 +438,21 @@ mod app {
         ethernet_link::Monotonic::spawn_after(1.secs()).unwrap();
     }
 
-    #[task(priority = 1, shared=[ltc2320])]
+    #[task(priority = 2, shared=[ltc2320])]
     fn ltc2320_start_conversion(mut c: ltc2320_start_conversion::Context) {
         // schedule next conversion for 1 Hz sample rate
-        ltc2320_start_conversion::Monotonic::spawn_after(1.secs()).unwrap();
+        ltc2320_start_conversion::Monotonic::spawn_after(500.micros()).unwrap();
         c.shared.ltc2320.lock(|ltc| ltc.start_conversion()).unwrap(); // panic if LTC2320 timer is already running
     }
 
-    #[task(binds = TIM7, priority = 1, shared=[ltc2320])]
+    #[task(binds = TIM7, priority = 2, shared=[ltc2320])]
     fn ltc2320_conv_done(mut c: ltc2320_conv_done::Context) {
         c.shared
             .ltc2320
             .lock(|ltc| ltc.handle_conv_done_irq().unwrap()); // panic if last QSPI transfer not complete
     }
 
-    #[task(binds = QUADSPI, priority = 1, shared=[ltc2320, ltc2320_data])]
+    #[task(binds = QUADSPI, priority = 2, shared=[ltc2320, ltc2320_data])]
     fn ltc2320_transfer_done(c: ltc2320_transfer_done::Context) {
         (c.shared.ltc2320, c.shared.ltc2320_data).lock(|ltc, data| {
             ltc.handle_transfer_done_irq(data);
--- a/src/bin/driver.rs
+++ b/src/bin/driver.rs
@@ -444,8 +444,7 @@ mod app {
 
     #[task(priority = 2, shared=[ltc2320], local=[ltc2320_conversion_scheduled])]
     fn ltc2320_start_conversion(mut c: ltc2320_start_conversion::Context) {
-        // schedule next conversion for 1 Hz sample rate
-        c.shared.ltc2320.lock(|ltc| ltc.start_conversion()).unwrap(); // panic if LTC2320 timer is already running
+        c.shared.ltc2320.lock(|ltc| ltc.start_conversion()).unwrap(); // panic if LTC2320 timing is not met
         *c.local.ltc2320_conversion_scheduled =
             *c.local.ltc2320_conversion_scheduled
                 + design_parameters::LTC2320_PERIOD.convert(); // update time at which the next conversion is scheduled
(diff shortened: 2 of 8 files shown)


jordens (commented):
Looks good. I'd still recommend leaning back and walking yourself through the code while reading it and asking yourself why you do/call things that way and not any other way.

src/hardware/driver/ltc2320.rs
```diff
@@ -0,0 +1,130 @@
...
+    pub fn new(
+        clocks: &rcc::CoreClocks,
+        qspi_rec: rcc::rec::Qspi,
+        qspi_peripheral: stm32::QUADSPI,
+        timer_rec: rcc::rec::Tim7,
+        timer_peripheral: stm32::TIM7,
+        mut pins: Ltc2320Pins,
+    ) -> Self {
+        let mut qspi =
+            qspi_peripheral.bank2(pins.spi, 1u32.MHz(), clocks, qspi_rec);
```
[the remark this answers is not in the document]
nkrackow: I just wanted to be relatively slow to look at scope traces. I'd make the decision for the SPI frequency once we have hardware. Since we don't use the return clock it's not so easy to say what the ideal frequency is, right?

src/bin/driver.rs
```diff
@@ -201,7 +182,7 @@ mod app {
         let clock = SystemTimer::new(|| monotonics::now().ticks() as u32);
 
         // Configure the microcontroller
-        let (stabilizer, _pounder) = hardware::setup::setup(
+        let (stabilizer, _pounder, driver) = hardware::setup::setup(
```
ryan-summers, line 185: Maybe `hardware::setup::setup()` should return (stabilizer, mezzanine), where `mezzanine` is the driver for the detected mezzanine card?

jordens: Right. A `enum { None, Pounder(Pounder), Driver(Driver) }`

src/bin/driver.rs
```diff
@@ -171,6 +152,8 @@ impl Default for Settings {
 
 #[rtic::app(device = stabilizer::hardware::hal::stm32, peripherals = true, dispatchers=[DCMI, JPEG, LTDC, SDMMC])]
 mod app {
+    use stabilizer::hardware::design_parameters;
```
jordens, line 155: Don't let vscode help you here. Move to top level.

src/bin/driver.rs
```diff
@@ -179,10 +162,11 @@ mod app {
     #[shared]
     struct Shared {
         network: NetworkUsers<Settings, Telemetry>,
-
         settings: Settings,
         telemetry: TelemetryBuffer,
         signal_generator: [SignalGenerator; 2],
+        ltc2320: driver::ltc2320::Ltc2320,
```
jordens, line 168: Please review the names in light of the existing ones.
remote_adc? Or header_adc?

src/bin/driver.rs
```diff
@@ -179,10 +162,11 @@ mod app {
...
+        ltc2320_data: [u16; 8],
```
jordens, line 169: Ditto

src/bin/driver.rs
```diff
@@ -226,6 +211,8 @@ mod app {
 
         let settings = Settings::default();
```
jordens, line 213: Add None to the enum. Then the unwrap and the double nesting in an option go away

src/bin/driver.rs
```diff
@@ -453,6 +443,26 @@ mod app {
         ethernet_link::Monotonic::spawn_after(1.secs()).unwrap();
     }
 
+    #[task(priority = 2, shared=[ltc2320], local=[ltc2320_conversion_scheduled])]
+    fn ltc2320_start_conversion(mut c: ltc2320_start_conversion::Context) {
+        c.shared.ltc2320.lock(|ltc| ltc.start_conversion()).unwrap(); // panic if LTC2320 timing is not met
+        *c.local.ltc2320_conversion_scheduled =
```
jordens, line 449: +=
But get the tooling to help you here

src/hardware/design_parameters.rs
```diff
@@ -17,6 +17,13 @@ pub const TIMER_PERIOD: f32 = 1. / (TIMER_FREQUENCY.to_Hz() as f32);
 /// The QSPI frequency for communicating with the pounder DDS.
 pub const POUNDER_QSPI_FREQUENCY: MegaHertz = MegaHertz::MHz(50);
 
+/// The QSPI frequency for LTC2320 readout.
+pub const DRIVER_QSPI_FREQUENCY: MegaHertz = MegaHertz::MHz(1); // Use slow 1 MHz for now. This will change.
```
jordens, line 21: If you don't need it anywhere else, i.e. if it's not shared, don't put it into design parameters.

src/hardware/driver/ltc2320.rs
```diff
@@ -0,0 +1,137 @@
...
+///! https://www.st.com/resource/en/errata_sheet/es0392-stm32h742xig-and-stm32h743xig-device-limitations-stmicroelectronics.pdf
+///!
+///! This driver is intended to be used in the following manner:
+///! 1. Trigger a new LTC2320 conversion with start_conversion(). This sets nCNV low and waits until TCONV has passed.
+///!    Then it pases the timer and starts the qspi readout.
+///! 2. Call handle_transfer_done_irq() in the QSPI ISR to retrieve the ADC data and set nCNV high again.
+///!
+///! Only works under the following condition:
+///! Conversions are not restarted faster than (T_readout + TCONV + TCNVH + readout/irq CPU overhead).
+///! You will always see an error
```
jordens, line 13: ?

src/hardware/driver/ltc2320.rs
```diff
@@ -0,0 +1,137 @@
...
+
+impl<E> From<E> for Error<E> {
+    fn from(err: E) -> Error<E> {
+        Error::QspiError(err)
+    }
+}
+
+pub struct Ltc2320Pins {
+    pub spi: (
+        gpiob::PB2<gpio::Alternate<9>>,
```
jordens, line 41: Is that really a different AF compared to  the other pins?

src/hardware/driver/ltc2320.rs
```diff
@@ -0,0 +1,137 @@
...
+        qspi_frequency: Hertz<u32>,
+        timer_rec: rcc::rec::Tim7,
+        timer_peripheral: stm32::TIM7,
+        timer_frequency: Hertz<u32>,
+        mut pins: Ltc2320Pins,
+    ) -> Self {
+        let mut qspi =
+            qspi_peripheral.bank2(pins.spi, qspi_frequency, clocks, qspi_rec);
+        qspi.configure_mode(QspiMode::OneBit).unwrap();
+        qspi.is_busy().unwrap(); // panic if qspi busy
```
jordens, line 72: Are you sure? I'd expect a Result<bool, _>

src/hardware/driver/ltc2320.rs
```diff
@@ -0,0 +1,137 @@
...
+                .admode()
+                .bits(1) // enable address phase
+                .adsize()
+                .bits(0b01) // set to 16 bit address to gain readback data alignment
+                .imode()
+                .bits(0) // disable instruction phase
+                .fmode()
+                .bits(0b01) // indirect read mode
+        });
+        qspi.inner_mut().cr.modify(
```
jordens, line 87: Can't you listen()?

src/hardware/driver/ltc2320.rs
```diff
@@ -0,0 +1,137 @@
...
+    /// Note that the CPU overhead for handling the irq leads to additional delay.
+    pub fn start_conversion(&mut self) -> Result<(), Error<QspiError>> {
+        self.cnv.set_low();
+        // check if the timer is running
+        let cr1 = self.timer.inner().cr1.read();
+        if cr1.cen().bit() == true {
+            return Err(Error::TimerBusyError);
+        }
+        self.timer.start(Ltc2320::TCONV.into_rate());
+        block!(self.timer.wait()).ok();
```
jordens, line 112: Why a timer if you block?
Just asmdelay

jordens (changes requested):
All in all looks good. No major problems.

src/hardware/driver/ltc2320.rs
nkrackow: The [asm-delay crate](https://docs.rs/asm-delay/0.9.0/asm_delay/struct.AsmDelay.html) only provides ms and us scale delays. I want to wait for only 450 ns. Should I just use cortex_m::asm::delay() with the right amount of cycles directly or rather wait for a us?

jordens: Either is fine.

src/hardware/driver/ltc2320.rs
nkrackow: Jes, just double checked it in the datasheet.

src/hardware/driver/ltc2320.rs
nkrackow: nope, returns an Err(Busy) if busy

jordens: It's a `Result<(), Busy>` which is a bit counter the usual API naming schemes.
Hence this is fine.

src/hardware/mod.rs
```diff
@@ -53,10 +57,32 @@ pub type NetworkManager = smoltcp_nal::shared::NetworkManager<
 pub type EthernetPhy = hal::ethernet::phy::LAN8742A<hal::ethernet::EthernetMAC>;
 
 /// System timer (RTIC Monotonic) tick frequency
-pub const MONOTONIC_FREQUENCY: u32 = 1_000;
+pub const MONOTONIC_FREQUENCY: u32 = 10_000;
```
jordens, line 60: Why do you need 10 kHz here?
If you change this then we have to double check the network stack timing as that expects millisecond ticks IIRC.

nkrackow: Ok. The problem is that this sets the highest possible Ltc2320 sample rate. Otherwise I'd have to eg. use a timer again instead of rtic scheduling.

jordens: Ok. Put (a) a comment why, (b) a TODO to clear this up and leave it like this for now. Later this may then require a timer.

.github/workflows/ci.yml
```diff
@@ -2,9 +2,9 @@ name: Continuous Integration
 
 on:
   push:
-    branches: [master, staging, trying]
```
jordens, line 5: Why does this need to be removed? I think the `pull_request` will trigger here.

jordens (approved):
Nice

.github/workflows/ci.yml
jordens: @nkrackow Forgot about this when approving. Could you comment and address in some later PR?

nkrackow: I wasn't sure how I get it to run CI on pushes to any branch. I thought I had it like this for thermostat and it ran on pushes to any branch but I did not check this.

PR #557, opened 2022-07-12
rj/pounder rework

jordens:
- stream: increase payload size, add fls format id
- pounder: need one io-update per batch
- pounder: use qspi DR write from PAC
- pounder: refactor
- pounder: rework io-update hrtimer


src/hardware/pounder/hrtimer.rs
```diff
@@ -54,37 +54,32 @@ impl HighResTimerE {
...
             3
         } else {
             panic!("Unattainable timing parameters!");
         };
 
-        let divider = 1 << (setting - 1);
-
         // The period register must be greater than or equal to 3 cycles.
-        let period = (source_cycles / divider as u32) as u16;
+        let period = (end / (1 << (div - 1)) as u32) as u16;
```
ryan-summers, line 82: Hmm... This indeed looks like it didn't work right for div = 3. Looks like we were unintentionally using a slightly shorter period in some cases. Nice fix!

src/net/data_stream.rs
```diff
@@ -46,7 +46,8 @@ const HEADER_SIZE: usize = 8;
 const FRAME_COUNT: usize = 4;
 
 // The size of each livestream frame in bytes.
-const FRAME_SIZE: usize = 1024 + HEADER_SIZE;
+// Ensure this is below the network MTU (e.g. 1500 - ip and udp header)
+const FRAME_SIZE: usize = 4 * 8 * 2 * 22 + HEADER_SIZE;
```
ryan-summers, line 50: I'm not following what these magic numbers (4, 8, 2, 22) are supposed to be. Can we either document and/or break them out into their own configurations?

Also, there's a downside of using multiplication here in that it's hard to obviously see if the MTU size is violated. Perhaps we should add a static, compile-time assert on the frame size.

ryan-summers (approved):
Were most of the changes to the `ad9959` configuration to help improve calculation speeds? Everything looked acceptable to me, just curious about the purpose.

There's some (very minor) nits below, but feel free to ignore if you disagree with any of them.

src/hardware/pounder/hrtimer.rs
jordens: What do you mean. The code should be equivalent. This wasn't intended to be a fix ;)

src/net/data_stream.rs
jordens: Changed to count down from MTU size.

src/hardware/pounder/hrtimer.rs
ryan-summers: Ah, sorry, I had completely missed the deletion of the `let divider = 1 << (setting - 1)` above and thought we were using `divider = setting` in the previous version. Indeed, there is no actual change here.

jordens:
bors r=ryan-summers

PR #558, opened 2022-07-13
pounder: cpu adc settings

jordens:
* the adcs are already calibrated by the hal
* make 16 bit explicit
* also opt for a longer acquisition time


ryan-summers:
Indeed, the ADCs are calibrated by default. Longer sampling time seems reasonable to me as well, as this is only currently used for Pounder RF power measurements.

bors r+

PR #560, opened 2022-07-14
pounder: invert feature: pounder_v1_1 -> not(pounder_v1_0)

jordens:
Rationale: later versions are expected to maintain compatibility with
not(pounder_v1_0), else will have to implement their own feature flags.

### DIFF
--- a/.github/bors.toml
+++ b/.github/bors.toml
@@ -5,9 +5,7 @@ timeout_sec = 1200
 status = [
   "style",
   "compile (stable)",
-  "compile (stable, pounder_v1_1)",
   "compile (1.59.0)",
-  "compile (1.59.0, pounder_v1_1)",
   "doc",
   "HITL Run Status"
 ]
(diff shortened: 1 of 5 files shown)


.github/bors.toml
```diff
@@ -5,9 +5,7 @@ timeout_sec = 1200
 status = [
   "style",
   "compile (stable)",
-  "compile (stable, pounder_v1_1)",
   "compile (1.59.0)",
-  "compile (1.59.0, pounder_v1_1)",
```
ryan-summers, line 10: Do we want CI to also run to check for Pounder v1.0 support as well?

ryan-summers: Sounds good. This was my thought as well, but didn't know what the plans might be fore v1.0 - are any of them still running in the wild?

ryan-summers (approved):

.github/bors.toml
jordens: AFAIK a handful and few are used in practice.

PR #565, opened 2022-07-24
rj/pounder vec

jordens:
- pounder: use Vec for profile builder
- ad9959/ProfileSerializer: tune for uninit


jordens:
In fact this remains significantly slower than the previous implementation. Keeping the previous one and closing in favor of #566

PR #566, opened 2022-07-24
pounder: tune initialization

jordens:
https://github.com/japaric/heapless/issues/305

### DIFF
--- a/ad9959/src/lib.rs
+++ b/ad9959/src/lib.rs
@@ -518,7 +518,8 @@ pub struct ProfileSerializer {
     // heapless::Vec<u8, 32>, especially its extend_from_slice() is slow
     data: [u8; 32],
     index: usize,
-    mode: Mode,
+    // make mode u32 to work around https://github.com/japaric/heapless/issues/305
+    mode: u32,
 }
 
 impl ProfileSerializer {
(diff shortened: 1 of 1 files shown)


ryan-summers (commented):
While these changes seem sensible, I'm not really following exactly what is going on in the linked issue. Can you summarize what you're seeing and why this change makes a difference?

It seems like the issue is referring to MaybeUninit::uninit() still being initialized using memset/memclr, but we don't have any MaybeUninit::uninit() calls around here, so I'm confused.

jordens:
Yes. It's convoluted. For both a `Vec` (which uses `[MaybeUninit<u8>; capacity]`) and the current "manual" implementation (`[u8; capacity]`array) the compiler explicitly clears the entire array on stack via `memset()` but figures out that it doesn't need to if `mode` is a `u32`. Reasons why exactly are unknown to me.
The usage of `Vec` in #565 was abandoned because the `Vec<u8, _>::extend_from_slice()` is significantly slower than our `[u8; _]::copy_from_slice()`. The latter may actually be a possible valuable performance improvement in `heapless` (either unsafe and then `copy_from_slice()` or use the yet-to-be-stabilized array-of-MaybeUninit methods).

jordens:
The difference here is actually very significant (in the ~300 cycle range) for several reasons:
* the `memclr` is in the hot path of process()
* it doesn't get inlined
* it is in flash, not in ITCM
* needs a indirect jump with another wrapper generated

ryan-summers:
LGTM - thanks for the background info. As discussed, if we don't hear back on the heapless issue, we should consult rustc - it seems like MaybeUninit should be updated to guarantee no-initiatilization (or a variant should be added that does this).

bors r+

PR #570, opened 2022-07-25
Nk/driver development adcinternal

nkrackow:
This small PR implements a dummy for the MCU internal ADC module for Driver. Only small changes should be necessary once we have HW.

### DIFF
--- a/src/hardware/driver/adc_internal.rs
+++ b/src/hardware/driver/adc_internal.rs
@@ -32,9 +32,11 @@ pub struct AdcInternal {
 
 impl AdcInternal {
     // ADC sampling time (aka time to charge the sampling capacitor).
-    // For the RC circuit on the Stabilizer CPU ADC hearder, sampling times of more than
+    // For the RC circuit on the Stabilizer CPU ADC header, sampling times of more than
     // 160 ns were not observed to have an effect on the sampled values.
-    // 64 cycles with a 50 MHz ADC clock gives 1.28 us, which should be on the safe side.
+    // According to the datasheet the slow ADC channels, of which we use some,
+    // need at least 172 ns sampling time for 16 bits accuracy.
+    // 64 cycles with a 50 MHz ADC clock gives 1.28 us.
     const STABILIZER_T_SAMP: adc::AdcSampleTime = adc::AdcSampleTime::T_64;
     pub fn new(
         delay: &mut impl DelayUs<u8>,
--- a/src/hardware/driver/adc_internal.rs
+++ b/src/hardware/driver/adc_internal.rs
@@ -6,8 +6,8 @@ use super::super::hal::{
     gpio::{gpiof::*, Analog},
     hal::blocking::delay::DelayUs,
     prelude::*,
-    rcc::{rec, CoreClocks},
-    stm32::{ADC1, ADC2, ADC3},
+    rcc::{rec, CoreClocks, ResetEnable},
+    stm32::{ADC1, ADC12_COMMON, ADC2, ADC3, ADC3_COMMON},
 };
 
 const V_REF: f32 = 2.048; // ADC reference voltage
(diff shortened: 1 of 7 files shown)


nkrackow:
Works with analog voltages injected into the Stabilizer header pins. The values read just a bit low (3.5 % error).

nkrackow:
I opened an issue to track this https://github.com/quartiq/stabilizer/issues/571

src/hardware/driver/adc_internal.rs
```diff
@@ -0,0 +1,84 @@
...
+        adc_rcc: (rec::Adc12, rec::Adc3),
+        adc: (ADC1, ADC2, ADC3),
+        pins: AdcInternalPins,
+    ) -> Self {
+        // Setup ADCs
+        let (adc1, _adc2) = adc::adc12(adc.0, adc.1, delay, adc_rcc.0, clocks);
+        let adc3 = adc::Adc::adc3(adc.2, delay, adc_rcc.1, clocks);
+
+        let mut adc1 = adc1.enable();
+        adc1.set_resolution(adc::Resolution::SIXTEENBIT);
```
jordens, line 46: Better explicitly set a sampling time for both ADCs now.
Even if #571 is unresolved.

jordens (changes requested):
Just one minor thing.

src/hardware/driver/adc_internal.rs
nkrackow: I set it to a reasonable sampling time of 64 cycles now. With a 100 MHz clock I still eyeballed a difference between 8 and 16 cycles, so 64 cycles at 50 MHz should be on the safe side.

src/hardware/driver/adc_internal.rs
```diff
@@ -44,9 +44,13 @@ impl AdcInternal {
 
         let mut adc1 = adc1.enable();
         adc1.set_resolution(adc::Resolution::SIXTEENBIT);
+        // use reasonable sampling time that gives accurate results with headroom bat is also not too slow
```
jordens, line 47:
```suggestion
        // use reasonable sampling time that gives accurate results with headroom but is also not too slow
```

I'd add the actual calculation.
Also please start a branch where you port/cherry-pick changes like this one (which applies to Pounder CPU ADC channels as well) to master. Or at least note (file issue) that this should be done.

src/hardware/setup.rs
```diff
@@ -265,6 +268,8 @@ pub fn setup(
         .per_ck(design_parameters::TIMER_FREQUENCY.convert())
         .pll2_p_ck(100.MHz())
         .pll2_q_ck(100.MHz())
+        .pll3_p_ck(50.MHz())
```
jordens, line 271: Why two outputs?

jordens (changes requested):

src/hardware/setup.rs
nkrackow: I have to initialize the p output in order for the r output to work. Otherwise the HAL panics.

jordens: Weird. But ok.
But do explain that.
And also add pointers/explanation of the prescaler issue in the hal (we should ultimagely use standard per_ck with a /2 additional divider, right?).

src/hardware/driver/adc_internal.rs
```diff
@@ -31,6 +31,11 @@ pub struct AdcInternal {
 }
 
 impl AdcInternal {
+    // ADC sampling time (aka time to charge the sampling capacitor).
+    // For the RC circuit on the Stabilizer CPU ADC hearder, sampling times of more than
+    // 160 ns were not observed to have an effect on the sampled values.
```
jordens, line 36: Ok. But it's always strictly better to derive things (like the minimum sampling time) instead of eyeballing them.

src/hardware/setup.rs
nkrackow: Yes I'm still trying to figure out why the prescaler register is always reset (even if I delete the resets/restarts in the HAL)...

jordens (approved):
Ok.

PR #572, opened 2022-07-26
rj/pounder aux adc

jordens:
- pounder: rename cpu adc channels, add aux adc inputs
- pounder: add aux adc api, clippy lints

* [x] tested on fls


ryan-summers (approved):
The changes look good to me. We'll likely build on this for temperature sensor support as well.

bors r+

PR #573, opened 2022-07-27
Feature/shared adc channels

ryan-summers:
This PR presents a proof-of-concept for using an ADC with many different channels that are needed by different drivers.

The implementation mirrors that of `shared-bus`. 

This builds on #572

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -56,7 +56,7 @@ fugit = "0.3"
 rtt-logger = "0.2"
 systick-monotonic = "1.0"
 mono-clock = "0.1"
-spinning = { version = "0.1", default-features = false }
+spin = { version = "0.9", default-features = false, features = ["spin_mutex"]}
 
 [dependencies.stm32h7xx-hal]
 features = ["stm32h743v", "rt", "ethernet", "xspi"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -56,6 +56,7 @@ fugit = "0.3"
 rtt-logger = "0.2"
 systick-monotonic = "1.0"
 mono-clock = "0.1"
+spinning = { version = "0.1", default-features = false }
 
 [dependencies.stm32h7xx-hal]
 features = ["stm32h743v", "rt", "ethernet", "xspi"]
--- a/src/hardware/setup.rs
+++ b/src/hardware/setup.rs
@@ -801,10 +801,10 @@ pub fn setup(
             )
         };
 
-        let pwr0 = adc1.create_channel(gpiof.pf11.into_analog()).unwrap();
-        let pwr1 = adc2.create_channel(gpiof.pf14.into_analog()).unwrap();
-        let aux_adc0 = adc3.create_channel(gpiof.pf3.into_analog()).unwrap();
-        let aux_adc1 = adc3.create_channel(gpiof.pf4.into_analog()).unwrap();
+        let pwr0 = adc1.create_channel(gpiof.pf11.into_analog());
+        let pwr1 = adc2.create_channel(gpiof.pf14.into_analog());
+        let aux_adc0 = adc3.create_channel(gpiof.pf3.into_analog());
+        let aux_adc1 = adc3.create_channel(gpiof.pf4.into_analog());
 
         let pounder_devices = pounder::PounderDevices::new(
             io_expander,
--- a/src/hardware/setup.rs
+++ b/src/hardware/setup.rs
@@ -14,8 +14,9 @@ use smoltcp_nal::smoltcp;
 
 use super::{
     adc, afe, dac, design_parameters, eeprom, input_stamper::InputStamper,
-    pounder, pounder::dds_output::DdsOutput, shared_adc, timers, DigitalInput0,
-    DigitalInput1, EthernetPhy, NetworkStack, SystemTimer, Systick, AFE0, AFE1,
+    pounder, pounder::dds_output::DdsOutput, shared_adc::SharedAdc, timers,
+    DigitalInput0, DigitalInput1, EthernetPhy, NetworkStack, SystemTimer,
+    Systick, AFE0, AFE1,
 };
 
 const NUM_TCP_SOCKETS: usize = 4;
@@ -745,12 +746,21 @@ pub fn setup(
 
         hal::adc::Temperature::new().enable(&adc3);
 
-        (adc1.enable(), adc2.enable(), adc3.enable())
-    };
+        let adc1 = adc1.enable();
+        let adc2 = adc2.enable();
+        let adc3 = adc3.enable();
 
-    let adc1 = shared_adc::new_shared_adc!(hal::stm32::ADC1 = adc1).unwrap();
-    let adc2 = shared_adc::new_shared_adc!(hal::stm32::ADC2 = adc2).unwrap();
-    let adc3 = shared_adc::new_shared_adc!(hal::stm32::ADC3 = adc3).unwrap();
+        (
+            // The ADCs must live as global, mutable singletons so that we can hand out references
+            // to the internal ADC. If they were instead to live within e.g. StabilizerDevices,
+            // they would not yet live in 'static memory, which means that we could not hand out
+            // references during initialization, since those references would be invalidated when
+            // we move StabilizerDevices into the late RTIC resources.
+            cortex_m::singleton!(: SharedAdc<hal::stm32::ADC1> = SharedAdc::new(adc1.slope() as f32, adc1)).unwrap(),
+            cortex_m::singleton!(: SharedAdc<hal::stm32::ADC2> = SharedAdc::new(adc2.slope() as f32, adc2)).unwrap(),
+            cortex_m::singleton!(: SharedAdc<hal::stm32::ADC3> = SharedAdc::new(adc3.slope() as f32, adc3)).unwrap(),
+        )
+    };
 
     // Measure the Pounder PGOOD output to detect if pounder is present on Stabilizer.
     let pounder_pgood = gpiob.pb13.into_pull_down_input();
--- a/src/hardware/setup.rs
+++ b/src/hardware/setup.rs
@@ -14,7 +14,7 @@ use smoltcp_nal::smoltcp;
 
 use super::{
     adc, afe, dac, design_parameters, eeprom, input_stamper::InputStamper,
-    pounder, pounder::dds_output::DdsOutput, timers, DigitalInput0,
+    pounder, pounder::dds_output::DdsOutput, shared_adc, timers, DigitalInput0,
     DigitalInput1, EthernetPhy, NetworkStack, SystemTimer, Systick, AFE0, AFE1,
 };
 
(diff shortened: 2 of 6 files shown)


src/hardware/shared_adc.rs
```diff
@@ -0,0 +1,80 @@
...
+
+impl<'a, Adc, PIN> AdcChannel<'a, Adc, PIN>
+where
+    PIN: Channel<Adc, ID = u8>,
+    hal::adc::Adc<Adc, hal::adc::Enabled>: OneShot<Adc, u32, PIN>,
+    <hal::adc::Adc<Adc, hal::adc::Enabled> as OneShot<Adc, u32, PIN>>::Error:
+        core::fmt::Debug,
+{
+    pub fn read(&mut self) -> f32 {
+        cortex_m::interrupt::free(|cs| {
```
ryan-summers, line 25: The current synchronization mechanism is a critical section, which is likely undesriable.

Currently, the only user of ADC channels is Pounder, which uses them for:
1. Reading input power measurements
2. Reading auxiliary ADC inputs

In the future, the ADC will also be used for:
1. Measuring internal CPU temperature

We need to determine a proper synchronization mechanism for this.

@jordens When are the auxiliary pounder ADC inputs supposed to be read?

ryan-summers: It looks like the input power measurements and the CPU temperature will likely all be measured in the same `telemetry` task, so we can likely use an atomic-check "mutex" that just double-checks that we haven't pre-empted ourselves (similar to shared-bus-rtic's concepts).

src/hardware/shared_adc.rs
```diff
@@ -0,0 +1,117 @@
...
+#[derive(Debug, Copy, Clone)]
+pub enum AdcError {
+    /// Indicates that the ADC channel has already been allocated.
+    Allocated,
+}
+
+/// A single channel on an ADC peripheral.
+pub struct AdcChannel<'a, Adc, PIN> {
+    pin: PIN,
+    slope: f32,
```
jordens, line 29: also add offset such that: `read() -> raw_voltage * slope + offset`

src/hardware/shared_adc.rs
jordens: Yes. Mutex (advantage no blocking of higher prio tasks, disadvantage possible deadlock) would be better than CS (blocking of higher prio tasks) here.
But alternatively we could just let RTIC handle it as an ADC resource. For that move it into `StabilizerDevices` (IIRC) and move all three ADCs into it and require that the ADC resource be `lock()` ed when a tasks wants to sample a (e.g. local resource) channel. API would be `ADCsResource.read(&AdcChannel)`. Since ADC1/2/3 are used all over there doesn't seem a natural need or way to split them.

jordens (changes requested):

ryan-summers:
The "AtomicCheckMutex" used in shared-bus is essentially "Check & Panic" if the ADC is currently in use, and it's what I was thinking about as well. We do this for the various networking clients.

The main issue is that we can't make the compile-time checks as clean as with networking, since we'd need a struct like `AdcUsers` - however, right now, there's the stabilizer app **and** the PounderDevices that will eventually need them, and putting them into a single structure would be... odd

src/hardware/shared_adc.rs
ryan-summers: The main purpose of this design is such that the ADC1/ADC2/ADC3 peripherals do **not** need to be passed around because that's painful from an end-user perspective.

ryan-summers:
@jordens What's the current use-case for measuring the AUX ADCs on Pounder? One possible API that we can go with is "fail-if-in-use". This would essentially cause `read()` to return an `Err` variant if the ADC was in use by a lower priority task. This would be fine for the current telemetry-based API, but if the AUX ADCs are being used somewhere else where fallibility isn't acceptable, that design wouldn't be acceptable.

src/hardware/shared_adc.rs
jordens: Ack

jordens:
Current use case is, as you say, only measuring a few things (rf power, aux adc voltages, soon cpu temperature) in telemetry. Driver will likely use some ADC channels outside telemetry in higher prio tasks as well.
Let's do it the way you propose with a fail-if-in-use. I think that then wants `RefCell<Mutex>` and `try_lock()`, right?

src/hardware/shared_adc.rs
ryan-summers: The `read()` function is really only intended to provide a normalized reading (e.g. 0-1 ADC scale). The slope here is the `adcX.slope()` result, which indicates the upper-limit of measurement.

src/hardware/shared_adc.rs
```diff
@@ -12,24 +12,64 @@
...
+struct AdcMutex<Adc> {
+    in_use: core::sync::atomic::AtomicBool,
+    adc: core::cell::RefCell<hal::adc::Adc<Adc, hal::adc::Enabled>>,
+}
+
+impl<Adc> AdcMutex<Adc> {
+    pub fn new(adc: hal::adc::Adc<Adc, hal::adc::Enabled>) -> Self {
+        Self {
+            adc: core::cell::RefCell::new(adc),
+            in_use: core::sync::atomic::AtomicBool::new(false),
```
jordens, line 30: Isn't there a generic mutex no_std crate that does this? If not, shouldn't there be?
Shouldn't this be an `impl` of https://rahix.github.io/embedded-hal/embedded_hal/mutex/trait.RwMutex.html ?

ryan-summers: Maybe there's other structurs in Spinning that do this. Give me a moment.

src/hardware/shared_adc.rs
jordens: Ack.

src/hardware/shared_adc.rs
```diff
@@ -0,0 +1,113 @@
+/// Shared Internal ADC Support
```
ryan-summers, line 1: As a general comment, I think this whole module could be broken out into a separate crate for other users.

jordens: Like shared-bus? Or a specific shared-adc?

src/hardware/setup.rs
```diff
@@ -722,6 +722,36 @@ pub fn setup(
...
+        adc3.set_resolution(hal::adc::Resolution::SIXTEENBIT);
+
+        hal::adc::Temperature::new().enable(&adc3);
+
+        (adc1.enable(), adc2.enable(), adc3.enable())
+    };
+
+    let adc1 = shared_adc::new_shared_adc!(hal::stm32::ADC1 = adc1).unwrap();
+    let adc2 = shared_adc::new_shared_adc!(hal::stm32::ADC2 = adc2).unwrap();
+    let adc3 = shared_adc::new_shared_adc!(hal::stm32::ADC3 = adc3).unwrap();
```
jordens, line 753:
```suggestion
        (
            shared_adc::new_shared_adc!(hal::stm32::ADC1 = adc1.enable()).unwrap(),
            shared_adc::new_shared_adc!(hal::stm32::ADC2 = adc2.enable()).unwrap(),
            shared_adc::new_shared_adc!(hal::stm32::ADC3 = adc3.enable()).unwrap(),
        )
    };
```

src/hardware/shared_adc.rs
```diff
@@ -0,0 +1,113 @@
...
+    <hal::adc::Adc<Adc, hal::adc::Enabled> as OneShot<Adc, u32, PIN>>::Error:
+        core::fmt::Debug,
+{
+    /// Read the ADC channel.
+    ///
+    /// # Returns
+    /// The normalized ADC measurement as a ratio of full-scale.
+    pub fn read(&mut self) -> Result<f32, AdcError> {
+        let mut adc = self.mutex.try_lock().ok_or(AdcError::InUse)?;
+        Ok(adc.read(&mut self.pin).unwrap() as f32 / self.slope)
```
jordens, line 50: Why not just

```suggestion
        Ok(adc.read(&mut self.pin).unwrap() as f32 / adc.slope() as f32)
```

And forget about `self.slope`? `adc.slope() as f32` is fast.

src/hardware/shared_adc.rs
```diff
@@ -0,0 +1,113 @@
...
+    }
+}
+
+/// An ADC peripheral that can provide ownership of individual channels for sharing between
+/// drivers.
+pub struct SharedAdc<Adc> {
+    mutex: spinning::Mutex<hal::adc::Adc<Adc, hal::adc::Enabled>>,
+    allocated_channels: core::cell::RefCell<[bool; 20]>,
```
jordens, line 58: Is that needed? The `hal` will presumably ensure that `adc::Channel` is unique (it's the pin), and since that's owned by `AdcChannel`.

src/hardware/shared_adc.rs
```diff
@@ -0,0 +1,113 @@
...
+    ($adc_type:ty = $adc:expr) => {{
+        let m: Option<&'static mut _> = cortex_m::singleton!(
+            : $crate::hardware::shared_adc::SharedAdc<$adc_type> = $crate::hardware::shared_adc::SharedAdc::new($adc.slope() as f32, $adc)
+        );
+
+        m
+    }};
+}
+
+/// Construct a shared ADC driver into a global, mutable singleton.
```
jordens, line 112: Could you comment why that's needed and why it can't just live in `StabilizerDevices`? Just convenience to be able to created channels from the global `SharedAdcs`?

Cargo.toml
```diff
@@ -56,6 +56,7 @@ fugit = "0.3"
 rtt-logger = "0.2"
 systick-monotonic = "1.0"
 mono-clock = "0.1"
+spinning = { version = "0.1", default-features = false }
```
jordens, line 59: I wonder whether the more active `spin` might be a safer bet here. `spinning` seems a bit dead and claims to distinguish itself from `spin` by features that `spin` certainly has (`lock_api` etc). Certainly compiles fine with `spin` instead.

jordens (changes requested):
Nice

src/hardware/shared_adc.rs
```diff
@@ -0,0 +1,113 @@
...
+
+        Ok(AdcChannel {
+            pin,
+            slope: self.slope,
+            mutex: &self.mutex,
+        })
+    }
+}
+
+macro_rules! new_shared_adc {
```
jordens, line 102: And can't that be just inlined? Doesn't look like a lot of code.

src/hardware/shared_adc.rs
jordens: Ah. That's only implemented in a macro on each `Adc<$ADC, _>` in the hal. Unfortunate.

src/hardware/shared_adc.rs
ryan-summers: I think we could make a `shared-adc` crate, although I don't know if there's a ton of interest or not.

src/hardware/shared_adc.rs
ryan-summers: Yeah, its hidden behind a trait implementation and is _exceptionally_ not obvious. Took me a while to debug why it wasn't working in the first place, since I had it like this originally.

src/hardware/shared_adc.rs
ryan-summers: There's some edge cases for internal channels where this isn't necessarily the case (although they are few and limited). One example is the CPU temperature sensor. You just instantiate a `let temp = hal::adc::Temperature::new()` to get an object that impls the adc::Channel.

That being said, I'm not opposed to just removing this check and making it up to the user to not do dumb things. For the vast majority of pin-related channels, as you note, this won't be an issue because of ownership.

src/hardware/shared_adc.rs
ryan-summers: This is the same principal behind shared-bus - to be able to hand out references to the ADC, it must live in `'static` memory. `StabilizerDevices` is a `LateResource`, which means it's run-time initialized and then **moved** into `'static` memory. We are trying to give out references at initialization-time, which is before that move occurs, hence the need for a `'static mut` singleton. Will update docs.

Cargo.toml
ryan-summers: Didn't even see `spin`. Happy to move to a more maintained crate!

src/hardware/shared_adc.rs
ryan-summers: Yeah, there's a fair argument for making the code more understandable. Let me take a look.

src/hardware/shared_adc.rs
jordens: Ah. But even then, if multiple `adc::Temperature::new()` can be created, they will happily coexist, right?

src/hardware/shared_adc.rs
jordens: Ack!

src/hardware/shared_adc.rs
ryan-summers: That's an interesting point. Indeed, it seems like there's no reason that an AdcChannel shouldn't be copyable if we're avoiding potential pre-emption of the resource.

ryan-summers: I think as long as you exclusively consume channels, you'd be fine. If we provided a `free()` mechanism that returned the pins, this obviously wouldn't work. In any case, that's out-of-scope here, so I'll go ahead and remove the checks. Should clean up the code a bit.

src/hardware/shared_adc.rs
jordens: Ok.

jordens (approved):

Issue #571, opened 2022-07-25
Stabilizer MCU internal ADCs show an offset error.


nkrackow:
The MCU internal ADCs show a DC offset error. Eg. connecting 1 V to a pin gives a reading equivalent to ~0.96 V. The voltage over the last capacitor right before the MCU pin reads 0.990 V on a good multimeter.
I checked the ADC reference voltage and it looks very stable at 2.046 V (close to nominal 2.048 V).
We've seen something similar on Thermostat-EEM that has the same MCU. 
https://github.com/quartiq/thermostat-eem/issues/21 https://github.com/quartiq/thermostat-eem/issues/17 https://github.com/quartiq/thermostat-eem/issues/16

nkrackow:
Some things I currently don't understand:
- Setting the sampling time lower seems to make the values more accurate. With 2.5 ADC cycles sampling time the values are bang on. With 8 cycles the values are about 1 % off and for >16 cycles they are ~ 2-3 % off.
- The default sampling time should be 1.5 ADC cycles since this is the reset vaue and the respective register is not written in the adc startup of the hal driver. But if I don't set another sampling time the values read low as observed with longer sampling times.

nkrackow:
More observations:
- The sampling time has a strong effect on linearity. A short sampling time leads to a massive error at both ends of the dynamic range.
- With a long sampling time there seems to be little non-linearity but a constant offset of about 0.03 V.

nkrackow:
Pin leakage current should be less than 1 uA. 2K external resistance cannot account for the offset and I also measured the voltage on the capacitor right on the pin.

nkrackow:
I've now thoroughly checked that [25.4.8 Calibration](https://www.st.com/resource/en/reference_manual/dm00314099-stm32h742-stm32h743-753-and-stm32h750-value-line-advanced-arm-based-32-bit-mcus-stmicroelectronics.pdf#page=931) is followed. The only thing that is not explicitly checked is the that the LDO is ready, which can only be checked on some device revisions apparently. I inserted an extra second of delay right before starting the calibration and it makes no difference so I think the LDO is not the problem. 
Therefore I can't find a fault in the HAL's calibration routine.

nkrackow:
I should also mention that not performing the calibration leads to an offset of about the same magnitude but in the positive direction.

nkrackow:
I seem to have found the culprit. A slower (10 MHz) ADC clock leads to very accurate samples on all channels with a long sampling time.
[...]

PR #575, opened 2022-07-28
Temperature sensor support

### LINKED ISSUE #568: temperature sensor support (cpu on stabilizer and lm75 on pounder)
* The CPU has an internal temperature sensor connected to ADC3. Implement that and add it to telemetry in `dual-iir`/`lockin`.
* Pounder has an LM75 connected via I2C. Implement that and add it to the Pounder API (to be added to application telemetry)

ryan-summers:
This PR fixes #568 by adding temperature sensor support for Pounder and Stabilizer's CPU.

The LM75 on Pounder is exposed as a public member variable, so end-user applications can choose where to use it.

The Stabilizer CPU temperature sensor is exposed as a member of `StabilizerDevices` and can be used to measure the CPU junction temperature.

### Testing
I installed Pounder on top of Stabilizer and modified `dual-iir` to report both stabilizer CPU and pounder temperatures in telemetry. Temperatures were measured over time using MQTT explorer:
image


src/bin/dual-iir.rs
```diff
@@ -194,6 +194,7 @@ mod app {
         dacs: (Dac0Output, Dac1Output),
         iir_state: [[iir::Vec5<f32>; IIR_CASCADE_LENGTH]; 2],
         generator: FrameGenerator,
+        temp_sensor: stabilizer::hardware::temp_sensor::CpuTempSensor,
```
jordens, line 197:
```suggestion
        cpu_temp: stabilizer::hardware::temp_sensor::CpuTempSensor,
```

as it's not the only temperature sensor on Stabilizer.

src/bin/dual-iir.rs
```diff
@@ -252,6 +253,7 @@ mod app {
             dacs: stabilizer.dacs,
             iir_state: [[[0.; 5]; IIR_CASCADE_LENGTH]; 2],
             generator,
+            temp_sensor: stabilizer.temperature_sensor,
```
jordens, line 256:
```suggestion
            cpu_temp: stabilizer.cpu_temp,
```

src/bin/dual-iir.rs
```diff
@@ -427,7 +429,7 @@ mod app {
         c.shared.network.lock(|net| net.direct_stream(target));
     }
 
-    #[task(priority = 1, shared=[network, settings, telemetry])]
+    #[task(priority = 1, shared=[network, settings, telemetry], local=[temp_sensor])]
```
jordens, line 432:
```suggestion
    #[task(priority = 1, shared=[network, settings, telemetry], local=[cpu_temp])]
```

src/bin/dual-iir.rs
```diff
@@ -438,8 +440,11 @@ mod app {
             .lock(|settings| (settings.afe, settings.telemetry_period));
 
         c.shared.network.lock(|net| {
-            net.telemetry
-                .publish(&telemetry.finalize(gains[0], gains[1]))
+            net.telemetry.publish(&telemetry.finalize(
+                gains[0],
+                gains[1],
+                c.local.temp_sensor.get_temperature().unwrap(),
```
jordens, line 446:
```suggestion
                c.local.cpu_temp.get_temperature().unwrap(),
```

etc

src/hardware/mod.rs
```diff
@@ -11,6 +11,7 @@ pub mod pounder;
 pub mod setup;
 pub mod shared_adc;
 pub mod signal_generator;
+pub mod temp_sensor;
```
jordens, line 14:
```suggestion
pub mod cpu_temp;
```

jordens (changes requested):
Just a small rename.
I do wonder why the measured CPU temperature is fluctuating that much.
https://github.com/quartiq/stabilizer/issues/571 ist certainly a factor but usually that's a constant error.

ryan-summers:
> Just a small rename. I do wonder why the measured CPU temperature is fluctuating that much. #571 ist certainly a factor but usually that's a constant error.

Is it usually constant? I imagine running at above the rated max clock rate would definitely result in fluctuation. Let me also take a second look at the required settling time specified in the datasheet.

jordens:
Few things are constant, but I'd be surprised if our CPU temperature changes by more than 0.1 K in a second when running.
Let's see how Norman's ADC changes pan out. They will lower the ADC clock (but also lower the sampling time). We may have to set up a different sampling time (entire ADC configuration) for the temperature sensor.

PR #578, opened 2022-07-28
use refactored mcp230xx

jordens:
I gave up trying to fix an issue where GPIO pin setting failed and refactored/rewrote the mcp23017 crate. The previous crate tried to copy the arduino library too much. Now it uses proper rust enums for pins and states, crates for bit manipulation etc. I ripped out the interrupt code which we don't use and which potentially was never tested. Overall this leads to a much more compact and maintainable crate and also here in stabilizer the API feels much better. 
It also supports the mcp23008 of @nkrackow.

* [x] `mcp230xx` [released](https://crates.io/crates/mcp230xx)
* [x] tested on fls
* [x] bump MSRV as `mcp230xx` uses the Default derive for enums.

### DIFF
--- a/src/hardware/pounder/mod.rs
+++ b/src/hardware/pounder/mod.rs
@@ -1,3 +1,5 @@
+use self::attenuators::AttenuatorInterface;
+
 use super::hal;
 use crate::hardware::shared_adc::AdcChannel;
 use embedded_hal::blocking::spi::Transfer;
@@ -379,11 +381,11 @@ impl PounderDevices {
         // selected and enabled, attenuators out of reset. Note that testing indicates the
         // output state needs to be set first to properly update the output registers.
         for (pin, level) in [
-            (GpioPin::AttLe0, mcp230xx::Level::High),
-            (GpioPin::AttLe1, mcp230xx::Level::High),
-            (GpioPin::AttLe2, mcp230xx::Level::High),
-            (GpioPin::AttLe3, mcp230xx::Level::High),
-            (GpioPin::AttRstN, mcp230xx::Level::High),
+            (GpioPin::AttLe1, mcp230xx::Level::Low),
+            (GpioPin::AttLe2, mcp230xx::Level::Low),
+            (GpioPin::AttLe3, mcp230xx::Level::Low),
+            (GpioPin::AttLe0, mcp230xx::Level::Low),
+            (GpioPin::AttRstN, mcp230xx::Level::Low),
             (GpioPin::ExtClkSel, mcp230xx::Level::Low),
             (GpioPin::OscEnN, mcp230xx::Level::Low),
             (GpioPin::Led4Green, mcp230xx::Level::Low),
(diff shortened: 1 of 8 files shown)


ryan-summers:
I remember having similar difficulties with the mcp23017 driver, and I do not believe it's very actively maintained. That being said, we should either:
1. Host the library as a separate, refactored crate
2. PR the changes into the existing crate

It'd be best for maintainability to not work off a branch or something similar.

nkrackow:
Maybe we should just have our own mcp230xx crate with these changes and small adaptations to support both chips then?

nkrackow (approved):

src/hardware/pounder/mod.rs
```diff
@@ -350,19 +381,33 @@ impl PounderDevices {
...
-            .mcp23017
-            .write_gpio(mcp23017::Port::GPIOA, 0x00)
-            .map_err(|_| Error::I2c)?;
-        devices
-            .mcp23017
-            .write_gpio(mcp23017::Port::GPIOB, 0x2F)
-            .map_err(|_| Error::I2c)?;
-
+        for pin in [
+            GpioPin::AttLe1,
```
ryan-summers, line 385: This seems like an easy place to forget to add one of the `GpioPin` members. Why not use https://crates.io/crates/iter-enum?

src/hardware/pounder/mod.rs
```diff
@@ -383,40 +428,42 @@ impl PounderDevices {
...
     /// Args:
     /// * `channel` - The attenuator channel to latch.
     fn latch_attenuator(&mut self, channel: Channel) -> Result<(), Error> {
-        let pin =
-            GpioPin::try_from(GpioPin::AttLe0 as u8 + channel as u8).unwrap();
-        // Active low
-        self.set_gpio_pin(pin, false)?;
-        self.set_gpio_pin(pin, true)
+        // Rising edge sensitive
+        self.set_gpio_pin(channel.into(), mcp230xx::Level::High)?;
```
ryan-summers, line 465: This looks like it inverted the low -> high logic that we had previously to generate a rising edge. This wouldn't generate an edge if the IO were already high.

ryan-summers (changes requested):

src/hardware/pounder/mod.rs
jordens: That's old/dead I think.
https://crates.io/crates/enum-iterator looks good.

src/hardware/pounder/mod.rs
jordens: yes. It's inverted because that's what the datasheet uses.
The second comment also applied previously. But it's not a problem since the state is know to be correct.

ryan-summers: Indeed, it appears that the device is infact rising edge sensitive.

While this code may be correct now, I feel that it would be easy to break in the future (e.g. by initializing IOs high instead of low). We can mitigate that and _guarantee_ that a rising edge is generated regardless of initial pin state by first setting low, then high.

jordens: Right. That's also why I changed it to this sequence a while ago.
https://github.com/quartiq/stabilizer/commit/fc78b087eeba5343d25dd90f729e4bf1a5180dd9
I'll try again with that robust sequence and see.

The reverted sequence here is the result of trying to hunt down the mcp23017 issues (which I still haven't fully done)...

ryan-summers (approved):

PR #582, opened 2022-08-02
Nk/driver development i2c

nkrackow:
This PR, once in a sensible state, will provide the I2C capabilities for Driver. 

I2C devices:
 - LM75
 - MCP23008 
 - eeprom currently just for checking before Driver is initialized

Using shared-bus-rtic for shared I2C bus and spin::Mutex for shared MCP23008 for the relays.


src/hardware/driver/shared_mcp23008.rs
```diff
@@ -0,0 +1,131 @@
+use core::fmt::Debug;
+
+///! Driver relays driver
```
jordens, line 3: Spurious file?

src/hardware/setup.rs
```diff
@@ -992,9 +1009,33 @@ pub fn setup(
...
+        let lm75 =
+            lm75::Lm75::new(i2c_manager.acquire(), lm75::Address::default());
+
+        let mcp23008 = mcp23017::MCP23017::new_default(
+            i2c_manager.acquire(),
+            mcp23017::Variant::MCP23008,
+        )
+        .unwrap();
+
+        let mcp_mutex =
```
jordens, line 1025: I don't think this needs to be a static singleton. Just make it, then move it into I2cDevices. You should still be able to create references to it when making `Relay::new()`.

jordens (changes requested):

src/hardware/driver/shared_mcp23008.rs
nkrackow: yep..

src/hardware/driver/relay.rs
```diff
@@ -0,0 +1,185 @@
...
+    pub fn handle_relay_done(&mut self) {
+        self.process_event(sm::Events::RelayDone).unwrap();
+    }
+}
+
+/// A SharedMcp can provide ownership of one of two sets of relays on Driver.
+/// Each set consists of two relays that control the state of a Driver output channel.
+/// The relays are controlled by toggeling pins on an MCP23008 I2C IO expander on the Driver board.
+/// Both sets use the same MCP23008 chip, hence the SharedMcp.
+pub struct SharedMcp<I2C> {
```
nkrackow, line 157: I now have a shared mcp that hands out relays.

src/hardware/driver/relay.rs
```diff
@@ -0,0 +1,185 @@
...
+            RelayPin::HP_K1_EN => Self::Gp5,
+            RelayPin::HP_K0_D => Self::Gp6,
+            RelayPin::HP_K0_CP => Self::Gp7,
+        }
+    }
+}
+
+pub struct Relay<'a, I2C: WriteRead + Write> {
+    mutex: &'a spin::Mutex<MCP23008<I2C>>,
+    k1_en_n: RelayPin,
```
nkrackow, line 54: And the Relay contains the four pins that can be one or the other set.

nkrackow:
I accidentally reviewed my own PR..
What do you think about the changes I mentioned in the comments? @jordens

src/hardware/setup.rs
nkrackow: Even after some discussion with @ryan-summers I couldn't figure out how to get by without a singleton.

jordens: Ack. Certainly fine as it is now.

jordens:
Looks good.

src/bin/driver.rs
```diff
@@ -482,6 +493,17 @@ mod app {
         });
     }
 
+    // This only processes the Low noise channel relay now.
+    // Todo: figure out how to handle both channels in one function.
+    #[task(priority = 1, shared=[driver_relay_state])]
+    fn handle_relay(mut c: handle_relay::Context) {
```
nkrackow, line 499: I just have to solve this problem still. I would like to only have one `relay_done` rtic task handling both the LN and HP relays.

nkrackow: nevermind, that was easy with just a function parameter

nkrackow:
This is currently untested. I would like to test next week:
- if I can toggle IOs on the MCP23017 on Pounder
- the basic relay state machine logic, either with Pounder or just with debug prints
- the LM75 using Pounder

src/hardware/driver/relay.rs
```diff
@@ -0,0 +1,256 @@
...
+/// [Relay] contains a mutex of a shared MCP23008, an I2C io expander chip with 8 GPIOs.
+/// 4 of those GPIOs are used to controll one set of relays (aka one [Relay]):
+///     - `k1_en_n` and `k1_en` differentially control the state of K1
+///     - `k0_d` is that data input of a flipflop which controls the state of K0
+///     - `k0_cp` is the clock input if the flipflop. Note that there is additional circuitry
+///        which ensures this clock input cannot be driven if K1 is not in the correct state.
+///        See Driver schematic for exact details.
+/// A `delay` is used to generate a rising edge for the flipflip with enaugh timing margins.
+pub struct Relay<'a, I2C: WriteRead + Write> {
+    mutex: &'a spin::Mutex<Mcp230xx<I2C, Mcp23008>>,
```
jordens, line 69: You should (but I've been wrong before) be able to get rid of the lifetime parameter. Since all you support is static anyways.

```suggestion
pub struct Relay<I2C: WriteRead + Write> {
    mutex: &'static spin::Mutex<Mcp230xx<I2C, Mcp23008>>,
```

And give it a better name than `mutex`. E.g. `gpio`, because the inner value is the relevant one, not the mutex.

src/hardware/driver/relay.rs
```diff
@@ -0,0 +1,256 @@
...
+    k0_cp: RelayPin,
+}
+
+impl<'a, I2C, E> Relay<'a, I2C>
+where
+    I2C: WriteRead<Error = E> + Write<Error = E>,
+    E: Debug,
+{
+    const K0_DELAY: fugit::MillisDuration<u64> =
+        fugit::MillisDurationU64::millis(10);
```
jordens, line 83: You want to check this against the relay datasheet already now. Saves you time later.

src/hardware/eeprom.rs
```diff
@@ -17,15 +24,15 @@ where
     // accomodate this, we repeat the I2C read for a set number of attempts with a fixed delay
     // between them. Then, we wait for the bus to stabilize by waiting until the MAC address
     // read-out is identical for two consecutive reads.
-    for _ in 0..40 {
+    for _ in 0..trys {
```
jordens, line 27: I'd stick with 40 for everything here (and not pass `trys`, but keep the return type change).
Or is there a point in reducing it to 5 for driver?

src/bin/driver.rs
```diff
@@ -482,6 +491,17 @@ mod app {
         });
     }
 
+    // Task for waiting the relay transition times.
+    #[task(priority = 1, shared=[driver_relay_state])]
+    fn handle_relay(mut c: handle_relay::Context, channel: driver::Channel) {
```
jordens, line 496: Nice.
But also prototype/mock the code that would call `enable()`/`disable()`.

src/hardware/driver/relay.rs
```diff
@@ -0,0 +1,256 @@
...
+            RelayPin::HP_K1_EN => Self::P5,
+            RelayPin::HP_K0_D => Self::P6,
+            RelayPin::HP_K0_CP => Self::P7,
+        }
+    }
+}
+
+// small helper to lock the mutex
+// maybe todo: Pass out an error if in use. Not sure how to get that out of the state machine though..
+fn get_mcp<I2C>(
```
jordens, line 52: It's a one-liner. Why not inline this?

src/hardware/driver/relay.rs
```diff
@@ -0,0 +1,256 @@
...
+    E: Debug,
+{
+    // set K0 to upper position (note that "upper" and "lower" refer to the schematic)
+    fn engage_k0(&mut self) {
+        let mut mcp = get_mcp(self.mutex);
+        // set flipflop data pin
+        mcp.set_gpio(self.k0_d.into(), Level::High).unwrap();
+        // set flipflop clock input low to prepare rising edge
+        mcp.set_gpio(self.k0_cp.into(), Level::Low).unwrap();
+        self.delay.delay_us(100u32);
```
jordens, line 160: 100 µs plus I2C transaction time just to clock a flipflop? I think this can be removed. Alternatively commenty why 100 µs.

src/hardware/driver/relay.rs
```diff
@@ -0,0 +1,256 @@
...
+}
+
+impl<'a, I2C, E> sm::StateMachine<Relay<'a, I2C>>
+where
+    I2C: WriteRead<Error = E> + Write<Error = E>,
+    E: Debug,
+{
+    /// Start relay enabling sequence. Returns the relay delay we need to wait for.
+    pub fn enable(&mut self) -> fugit::MillisDuration<u64> {
+        self.process_event(sm::Events::Enable).unwrap();
```
jordens, line 197: Note that this will panic if you call `enable()` or `disable()` (triggered by the user through MQTT at any time) while in one of the `*Wait*` states or in the wrong one of `Disable/Enable`. You should handle the first cases in code and the last one in the state machine (identity transitions `Disabled + Disable = Disabled` etc).

In general go through your code and review every unwrap for possible cases where the user might trigger them.

src/hardware/driver/relay.rs
```diff
@@ -0,0 +1,256 @@
...
+            k0_cp,
+        }
+    }
+}
+
+pub mod sm {
+    use super::*;
+    statemachine! {
+        transitions: {
+            *Disabled + Enable / engage_k0 = EnableWaitK0,
```
jordens, line 138: Comment that the hardware guarantees that initially you are in `Disabled`.

Also review whether there is a way for the hardware to go out of sync (`TEMP`?) with your internal state machine state.

src/hardware/driver/relay.rs
```diff
@@ -0,0 +1,256 @@
...
+                RelayPin::HP_K0_D,
+                RelayPin::HP_K0_CP,
+            )
+        };
+        let mut mcp = get_mcp(mutex);
+        // set GPIOs to default position
+        mcp.set_gpio(k1_en.into(), Level::Low).unwrap();
+        mcp.set_gpio(k1_en_n.into(), Level::High).unwrap();
+        mcp.set_gpio(k0_d.into(), Level::Low).unwrap();
+        mcp.set_gpio(k0_cp.into(), Level::Low).unwrap();
```
jordens, line 118: Review that this sequence does not accidentally cause any unwanted transitions in the hardware.

jordens (changes requested):
Just a couple minor nits. Looks fine otherwise!

src/hardware/eeprom.rs
nkrackow: It tries for a whopping 4 seconds if Driver is not present then.. It should really work on the first try for Driver.

jordens: 40 i2c transaction attempts of one byte don't take 4 s at 100 kHz. Something else is wrong.

jordens: Ah. I didn't see the 100 ms wait. Remove that then.

And if it should work the first time in your case, just use the respective function content. It's a single line and you don't need to use the repeated attempt logic.

nkrackow: It says in the comments that it can take seconds on Stabilizer for the eeprom power to come good. 
I think I'll just leave this file unchanged then and inline one eeprom read.

jordens: > It says in the comments that it can take seconds on Stabilizer for the eeprom power to come good.

But what's the point in waiting instead of continuing to try reading? It can't be synchronization and it can't be efficient use of resources or speed.

> I think I'll just leave this file unchanged then and inline one eeprom read.

Exactly. Less code.

src/hardware/setup.rs
```diff
@@ -748,12 +748,12 @@ pub fn setup(
...
         )
     };
-
-    let driver_mac_addr = eeprom::read_eui48(
-        &mut i2c1, &mut delay, 5, // try a few times
-    );
+    let mut buffer = [0u8; 6];
+    let driver_mac_addr =
+        i2c1.write_read(EEPROM_I2C_ADDR, &[EEPROM_MAC_POINTER], &mut buffer);
     log::info!("Driver EUI48: {:?}", driver_mac_addr);
```
jordens, line 754: Print `buffer`, not what `write_read()` returns.

src/hardware/driver/relay.rs
```diff
@@ -0,0 +1,242 @@
...
+    ///
+    /// # Args
+    /// * `ccdr`    - core clocks to construct a `delay` for flipflop control
+    /// * `ch`      - The Driver channel the relay pins are used for.
+    ///
+    /// # Returns
+    /// An instantiated [Relay] whose ownership can be transferred to other drivers.
+    pub fn obtain_relay(&'static self, ch: Channel) -> Relay<I2C> {
+        Relay::new(&self.mutex, ch)
+    }
```
jordens, line 241: There is no protection against multiple instantiation of the same relay, right? They would have independent state machines and would not match reality.

Ryan was tracking ADC channel allocation status in https://github.com/quartiq/stabilizer/blob/d6df836b8be45496d7c2dca6289ce3b05f1c5346/src/hardware/shared_adc.rs and making sure they aren't allocated multiple times. But with ADC channels there is actually no problem allocating them twice (they don't have internal state). Hence it was removed there. Your case explicitly has a lot of internal state (the machine).

src/hardware/driver/relay.rs
```diff
@@ -0,0 +1,242 @@
...
+            _ => None, // done, no delay needed
+        }
+    }
+}
+
+/// A SharedMcp can provide ownership of one of two sets of relays on Driver.
+/// Each set consists of two relays that control the state of a Driver output channel.
+/// The relays are controlled by toggeling pins on an MCP23008 I2C IO expander on the Driver board.
+/// Both sets use the same MCP23008 chip, hence the SharedMcp.
+pub struct SharedMcp<I2C1> {
```
jordens, line 212: I would remove this struct and inline the `Relay::new(&mutex, ch)` call.
It really doesn't seem to provide any functionality or abstraction.

src/hardware/eeprom.rs
```diff
@@ -1,10 +1,6 @@
 use embedded_hal::blocking::{delay::DelayMs, i2c::WriteRead};
 
-// The EEPROM is a variant without address bits, so the 3 LSB of this word are "dont-cares".
-const I2C_ADDR: u8 = 0x50;
```
jordens, line 4: make them `pub` here and use `hardware::eeprom::I2C_ADDR` etc in your code.
No need to move them into `hardware`.

jordens (changes requested):

src/hardware/driver/relay.rs
nkrackow: oops, this was more code at some point..

src/hardware/driver/relay.rs
nkrackow: It shouldn't afaict. But I am considering toggeling the flipflop once to ensure the latched state is not some glitch.

jordens (approved):

PR #584, opened 2022-08-03
add adc fix from driver

nkrackow:
Very small PR to fix the internal ADCs by setting the clock prescalers and then recalibrating. Code from the `driver` branch.

### DIFF
--- a/src/hardware/setup.rs
+++ b/src/hardware/setup.rs
@@ -724,6 +724,7 @@ pub fn setup(
     fp_led_2.set_low();
     fp_led_3.set_low();
 
+    // Use default PLL2p clock input with 1/2 prescaler for 50 MHz ADC clock.
     let (adc1, adc2, adc3) = {
         let (mut adc1, mut adc2) = hal::adc::adc12(
             device.ADC1,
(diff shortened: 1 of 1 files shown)


ryan-summers:
This doesn't seem to have fixed CPU temperature measurement appearing to be erratic:
image

nkrackow:
Hmm, let me quickly check about the other channels.

nkrackow:
Ok, I've confirmed that exactly these changes lead to the external inputs being very accurate. Without the changes we have the same ~3% error observed before.
I tested it by just printing adc values in an infinite loop after initialization but using exactly the same code otherwise.

nkrackow:
That means that the CPU temp readings have an additional, probably unrelated issue.

ryan-summers:
Hmm, then perhaps the ADC clock is not the only issue with CPU measurement stability. I'll open up an issue.

jordens (approved):

jordens:
Ack. But then we should land this.
bors r+

src/hardware/setup.rs
```diff
@@ -724,6 +724,7 @@ pub fn setup(
     fp_led_2.set_low();
     fp_led_3.set_low();
 
+    // Use default PLL2p clock input with 1/2 prescaler for 50 MHz ADC clock.
```
richardeoin, line 727: I think that PLL2p = 100MHz and PRESC = div 2 actually gives 25MHz a ADC clock on current parts. There's an extra /2 at the multiplexer output in RM0433 Figure 136.

image

I assume this ADC clock after the /2 is the clock denoted f_ADC in the datasheet image

nkrackow: Thanks. I would think you're right. They just drop the "ker" later and call this f_ADC. 
The only thing is that we consistently observed that the ADC reading are off using a 100 MHz clock input without an additional divider (aka this should make f_ADC = 50 MHz). This should be allowed with BOOST = 11 according to the datasheet. This was independent of the sample time or any other ADC variables AFAICT.

nkrackow: I find RM0433 kinda confusing regarding the ADC clocking...
Screenshot from 2022-08-15 10-13-44
How am I supposed to understand this? I assume  they mean the F_adc_ker_ck_input. Or is there another /2 divider behind F_adc_ker_ck?

jordens: 50 MHz is incompatible with the 80 MHz value for adc_ker_ck_input in RM0433, which is definitely before CCR:PREC and thus before the div2 no matter how one looks at it (and no matter whether the 50 MHz are before or after the div2.

image

Since the 80 MHz adc_ker_ck_input is the tighter constraint and matches the dramatic drop of between 80 MHz and 100 MHz we have seen, I'd go for that and enforce max 40 MHz adc_ker_ck.
Note the requirement for 50% duty and even PLL divider which might be a bit hairier to enforce. (Also note the typo VOS2/VOS3 in the column header and that I find it unlikely that VOS2 doesn't have this requirement).

PR #587, opened 2022-08-06
Python tooling refactorings

jordens:
- stream: increase OS level buffer size
- move get_local_ip to streaming
- move telemetry client implementation into python module (giving it ownership of its queue)
- add docs and update hitl tests


jordens:
bors try

jordens:
bors try

ryan-summers (approved):
bors r+

PR #590, opened 2022-08-09
Adding custom ASM delay implementation

### LINKED ISSUE #524: AsmDelay delays are likely 1/4th the time they should be.
https://github.com/quartiq/stabilizer/blob/96732cbea115efaa2012b69fc23fa18f95a56298/src/hardware/setup.rs#L277-L279

As discovered here https://github.com/quartiq/thermostat-eem/issues/27, the delays are not the expected time on the STM32H7. Since Stabilizer has the same CPU it should be affected too.

ryan-summers:
This PR fixes #524 by using a custom delay implementation that internally scales frequency x2 to account for the dual issue instruction pipeline in the M7.

Measurements were made on an oscilloscope by asserting the RX GPIO line for 200us. Pulse width measurement indicated 200.0uS was measured.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Fixed
 
 * Fixed panics when using a batch size of 1 ([#540](https://github.com/quartiq/stabilizer/issues/540))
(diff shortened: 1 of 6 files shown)


jordens (approved):
Ack.

But we should try to get away from `DelayUs`/`DelayMs` at some point in the future. The API is an invitation for error: `Into<u32>` strips units, granularity, and context as `delay_us(something.into())` looks just as good (but is wrong) if `something.into::<u32>()` are actually ms.

PR #598, opened 2022-08-16
use handled_update to return path

nkrackow:
Small PR to get information about which settings path was published on during polling. Unused in `dual-iir` and `lockin`.
Enables interlock functionality for Driver.

### DIFF
--- a/src/net/mod.rs
+++ b/src/net/mod.rs
@@ -37,12 +37,12 @@ pub enum UpdateState {
     Updated,
 }
 
-#[derive(Copy, Clone, PartialEq)]
 pub enum NetworkState {
-    SettingsChanged,
+    SettingsChanged(String<64>),
     Updated,
     NoChange,
 }
+
 /// A structure of Stabilizer's default network users.
 pub struct NetworkUsers<S: Default + Miniconf + Clone, T: Serialize> {
     pub miniconf: miniconf::MqttClient<S, NetworkReference, SystemTimer, 512>,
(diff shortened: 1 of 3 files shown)


jordens (approved):

src/net/mod.rs
```diff
@@ -159,8 +160,13 @@ where
             UpdateState::Updated => NetworkState::Updated,
         };
 
-        match self.miniconf.update() {
-            Ok(true) => NetworkState::SettingsChanged,
+        let mut settings_path = String::new();
+        match self.miniconf.handled_update(|path, old, new| {
+            settings_path.push_str(path).unwrap();
+            *old = new.clone();
```
ryan-summers, line 166: There's no need to perform this clone - miniconf does it internally. The only reason `old` and `new` are provided is so the user can detect what the value transition was on the given path.

ryan-summers (changes requested):

src/net/mod.rs
nkrackow: Ok, what's the reason for it being performed here?
https://github.com/quartiq/miniconf/blob/98438ab27794fbe0ad9b43a556b87241ad4097e8/src/mqtt_client/mod.rs#L433

ryan-summers: Ah you're correct - I'm mistaken here, this is required currently.

I don't see this mentioned in the `miniconf` docs anywhere, so I'll open an issue to clarify this. Thanks for pointing this out to me!

ryan-summers: I've opened https://github.com/quartiq/miniconf/issues/93 to clarify this. I think we could easily incorporate this into miniconf, but as-is, this implementation is correct.

ryan-summers (approved):
bors try

src/net/mod.rs
nkrackow: Ha, I'm happy that I understood what this closure does here, I was already disappointed.. ;)

PR #603, opened 2022-08-24
add limits calculation and helper script to plot the frequency response

nkrackow:
Small PR to add gain limits to the IIR coefficient calculations. I've also added a small helper to quickly plot the frequency response of the filter. I found it hard without that to just guess the parameters for the ballpark response that I want because I need to extrapolate the gains from 1 Hz.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -43,8 +43,6 @@ jobs:
             python -m pip install --upgrade pip
             pip install -e py
             pip install pylint
-            pip install numpy
-            pip install scipy
       - name: Run Pylint
         run: |
           pylint py/stabilizer
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -43,6 +43,8 @@ jobs:
             python -m pip install --upgrade pip
             pip install -e py
             pip install pylint
+            pip install numpy
+            pip install scipy
       - name: Run Pylint
         run: |
           pylint py/stabilizer
--- a/py/stabilizer/iir_coefficients.py
+++ b/py/stabilizer/iir_coefficients.py
@@ -112,14 +112,22 @@ def get_filters():
                       arguments=[
                           add_argument("--Kii", default=0, type=float,
                                        help="Double Integrator (I^2) gain"),
+                          add_argument("--Kii_limit", default=float('inf'), type=float,
+                                       help="Integral gain limit"),
                           add_argument("--Ki", default=0, type=float,
                                        help="Integrator (I) gain"),
+                          add_argument("--Ki_limit", default=float('inf'), type=float,
+                                       help="Integral gain limit"),
                           add_argument("--Kp", default=0, type=float,
                                        help="Proportional (P) gain"),
                           add_argument("--Kd", default=0, type=float,
                                        help="Derivative (D) gain"),
+                          add_argument("--Kd_limit", default=float('inf'), type=float,
+                                       help="Derivative gain limit"),
                           add_argument("--Kdd", default=0, type=float,
                                        help="Double Derivative (D^2) gain"),
+                          add_argument("--Kdd_limit", default=float('inf'), type=float,
+                                       help="Derivative gain limit"),
                       ],
                       coefficients=pid_coefficients),
     }
@@ -179,12 +187,12 @@ def pid_coefficients(args):
 
     # Determine filter order
     if args.Kii != 0:
-        assert (args.Kdd, args.Kd) == (0, 0), \
-                "IIR filters I^2 and D or D^2 gain are unsupported"
+        assert (args.Kdd, args.Kd, args.Kdd_limit, args.Kd_limit) == (0, 0, float('inf'), float('inf')), \
+            "IIR filters I^2 and D or D^2 gain/limit are unsupported"
         order = 2
     elif args.Ki != 0:
-        assert args.Kdd == 0, \
-                "IIR filters with I and D^2 gain are unsupported"
+        assert (args.Kdd, args.Kdd_limit) == (0, float('inf')), \
+            "IIR filters with I and D^2 gain/limit are unsupported"
         order = 1
     else:
         order = 0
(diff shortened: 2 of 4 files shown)


py/stabilizer/iir_coefficients.py
```diff
@@ -112,14 +112,22 @@ def get_filters():
                       arguments=[
                           add_argument("--Kii", default=0, type=float,
                                        help="Double Integrator (I^2) gain"),
+                          add_argument("--Kii_limit", default=float('inf'), type=float,
```
jordens, line 115:
```suggestion
from math import inf
...
                          add_argument("--Kii_limit", default=inf, type=float,
```

etc

py/stabilizer/iir_coefficients.py
```diff
@@ -196,12 +204,16 @@ def pid_coefficients(args):
     ]
 
     gains = [args.Kii, args.Ki, args.Kp, args.Kd, args.Kdd]
+    limits = [args.Kii/args.Kii_limit, args.Ki/args.Ki_limit,
+              1, args.Kd / args.Kd_limit, args.Kdd / args.Kdd_limit]
```
jordens, line 208:
```suggestion
              1, args.Kd/args.Kd_limit, args.Kdd/args.Kdd_limit]
```

py/stabilizer/plot_iir_frequency_response.py
```diff
@@ -0,0 +1,70 @@
...
+    # The feed-forward gain of the IIR filter is the summation
+    # of the "b" components of the filter.
+    forward_gain = sum(coefficients[:3])
+    if forward_gain == 0 and args.x_offset != 0:
+        print("Filter has no DC gain but x_offset is non-zero")
+
+    f = np.logspace(-7, np.log10(0.5 / args.sample_period), 1024, endpoint=False)
+    f, h = signal.freqz(
+        coefficients[:3],
+        [1] + [-c for c in coefficients[3:]],
```
jordens, line 54:
```suggestion
        np.r_[1, -coefficients[3:]],
```

py/stabilizer/plot_iir_frequency_response.py
```diff
@@ -0,0 +1,70 @@
...
+    subparsers = parser.add_subparsers(
+        help="Filter-specific design parameters", dest="filter_type", required=True
+    )
+
+    filters = get_filters()
+
+    for (filter_name, filt) in filters.items():
+        subparser = subparsers.add_parser(filter_name, help=filt.help)
+        for arg in filt.arguments:
+            subparser.add_argument(*arg.positionals, **arg.keywords)
```
jordens, line 36:
```suggestion
    for name, filter in get_filters().items():
        subparser = subparsers.add_parser(name, help=filter.help)
        for arg in filter.arguments:
            subparser.add_argument(*arg.positionals, **arg.keywords)
```

py/stabilizer/plot_iir_frequency_response.py
```diff
@@ -0,0 +1,70 @@
...
+
+    print(coefficients)
+
+    # The feed-forward gain of the IIR filter is the summation
+    # of the "b" components of the filter.
+    forward_gain = sum(coefficients[:3])
+    if forward_gain == 0 and args.x_offset != 0:
+        print("Filter has no DC gain but x_offset is non-zero")
+
+    f = np.logspace(-7, np.log10(0.5 / args.sample_period), 1024, endpoint=False)
```
jordens, line 51:
```suggestion
    f = np.logspace(-6.5, 0, 1024, endpoint=False)*(.5/args.sample_period)
```

py/stabilizer/plot_iir_frequency_response.py
```diff
@@ -0,0 +1,70 @@
...
+    f, h = signal.freqz(
+        coefficients[:3],
+        [1] + [-c for c in coefficients[3:]],
+        worN=f,
+        fs=1 / args.sample_period,
+    )
+    fig, ax = plt.subplots()
+    ax.margins(0, 0.1)
```
jordens, line 59:
```suggestion
```

Rely on default style as much as possible.

jordens (changes requested):
nice, just a few minor things.

.github/workflows/ci.yml
```diff
@@ -43,6 +43,8 @@ jobs:
             python -m pip install --upgrade pip
             pip install -e py
             pip install pylint
+            pip install numpy
+            pip install scipy
```
ryan-summers, line 47: These requirements should be added to stabilizer's `setup.py` file. Otherwise, we have a broken package.

py/stabilizer/plot_iir_frequency_response.py
```diff
@@ -57,7 +57,6 @@ def _main():
         fs=1 / args.sample_period,
     )
     _fig, ax = plt.subplots()
-    ax.margins(0, 0.1)
     ax.plot(f, 20 * np.log10(abs(h)))
```
jordens, line 60:
```suggestion
    ax.plot(f, 20 * np.log10(np.absolute(h)))
```

py/stabilizer/plot_iir_frequency_response.py
nkrackow: I cant just put the `-` on the list, right?

py/stabilizer/plot_iir_frequency_response.py
nkrackow: Actually I'd leave a bit more than 6.5 decades here. Eg. when using I gain limit that might taper off lower than you can see.

py/stabilizer/plot_iir_frequency_response.py
nkrackow: Ok, but I honestly like it much more without the useless space without a curve to the left and right of every plot.

py/stabilizer/plot_iir_frequency_response.py
nkrackow: I put `np.r_[1, [-c for c in coefficients[3:]]],` now

py/stabilizer/plot_iir_frequency_response.py
jordens: Sure. But before the lower limit didn't depend on the sample_period at all.

nkrackow: True. I put in 8 decades now.

jordens (approved):

PR #605, opened 2022-08-27
Add lvds[4-7] gpio to StabilizerDevices

pmldrmota:
We are using a stabilizer with a DIO-BNC card connected. This change facilitates access to the GPIO pins on the EEM connector without breaking existing applications.

The port assignment is 

LVDS | PortD | Direction
-- | -- | --
4 | 1 | Input
5 | 2 | Input
6 | 3 | Output
7 | 4 | Output

### DIFF
--- a/src/hardware/mod.rs
+++ b/src/hardware/mod.rs
@@ -35,6 +35,18 @@ pub type DigitalInput0 = hal::gpio::gpiog::PG9<hal::gpio::Input>;
 // Type alias for digital input 1 (DI1).
 pub type DigitalInput1 = hal::gpio::gpioc::PC15<hal::gpio::Input>;
 
+// Type alias for LVDS4 (digital input).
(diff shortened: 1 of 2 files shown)


jordens:
How does this conflict with pounder?

pmldrmota:
There is no conflict. Either Pounder is present, i.e. `Some(pounder)` and the EEM is `None`, or Pounder is `None` and EEM is `Some(eem)`.

jordens:
If there is no conflict, then why is it an `Option`?

pmldrmota:
I'm not using Pounder, so I'm not sure, but I thought Pounder occupies the EEM connector. If that's true, then the two are exclusive. We wouldn't want to mess with the EEM if Pounder is connected in that case.

jordens:
It doesn't.

pmldrmota:
Yes, sorry, I didn't double-check that.

jordens:
Thanks

Bors r+

pmldrmota:
(fixed style)

PR #609, opened 2022-08-30
Nk/driver development interlock

nkrackow:
This PR adds the MQTT interlock to Driver. 
I think I tested all interlock functionality/conditions and the behavior is always as I would expect.
I just find it a little confusing now since ignoring errors from the Handle (which happens in driver.rs) is part of the logic (eg. after an interlock has tripped). Maybe leaving everything in a single match in `settings_update()` would be clearer?

Also I currently can't think of anything that should happen on a clear true->false transition. Should I still put this option in?


src/hardware/driver/interlock.rs
```diff
@@ -0,0 +1,97 @@
...
+    path: String<64>,
+    handle_is_some: bool,
+    new: Interlock,
+    old: Interlock,
+) -> Option<Handle> {
+    let path = path.as_str();
+    if path == "interlock/interlock"
+        || path == "interlock/armed"
+        || path == "interlock/clear"
+    {
```
jordens, line 77: This I would do in the task, not here. It required knowledge of the complete settings structure and only serves to prevent doing anything here.

src/hardware/driver/interlock.rs
```diff
@@ -0,0 +1,97 @@
...
+        let cleared = !old.clear && new.clear;
+        match (handle_is_some, new.armed, cleared, new.interlock) {
```
jordens, line 79: My suspicion is that one or two (layers of) `if`s (with `return` if you want) would be sufficient to cover this in a more readable way.

src/hardware/driver/interlock.rs
```diff
@@ -0,0 +1,97 @@
...
+    Cancel,
+    Idle,
+}
+
+pub fn handle(
+    path: String<64>,
+    handle_is_some: bool,
+    new: Interlock,
+    old: Interlock,
+) -> Option<Handle> {
```
jordens, line 72: This is certainly not the handle. And you don't need the option (with the comment in L77)

```suggestion
) -> Action {
```

src/hardware/driver/interlock.rs
```diff
@@ -0,0 +1,97 @@
...
+}
+
+pub enum Handle {
+    Spawn(fugit::Duration<u64, 1, 10000_u32>),
+    Reschedule(fugit::Duration<u64, 1, 10000_u32>),
+    Cancel,
+    Idle,
+}
+
+pub fn handle(
```
jordens, line 67: Make it a method (of the new settings).

src/bin/driver.rs
```diff
@@ -398,41 +413,70 @@ mod app {
...
+                    .unwrap()
+                    .reschedule_after(millis)
+                    .ok(), // return `None` if rescheduled too late aka interlock already tripped
+                interlock::Handle::Cancel => {
+                    let _ = c.local.interlock_handle.take().unwrap().cancel(); // ignore error if cancelled too late
+                    None
+                }
+                interlock::Handle::Idle => None,
+            };
+        } else {
```
jordens, line 454: Always do this regardless.

src/bin/driver.rs
```diff
@@ -398,41 +413,70 @@ mod app {
...
+                interlock::Handle::Spawn(millis) => {
+                    Some(trip_interlock::spawn_after(millis).unwrap())
+                }
+                interlock::Handle::Reschedule(millis) => c
+                    .local
+                    .interlock_handle
+                    .take()
+                    .unwrap()
+                    .reschedule_after(millis)
+                    .ok(), // return `None` if rescheduled too late aka interlock already tripped
```
jordens, line 447: `log` those!

src/bin/driver.rs
```diff
@@ -264,7 +279,7 @@ mod app {
         local.dacs.1.start();
 
         // Spawn a settings update for default settings.
-        settings_update::spawn().unwrap();
+        settings_update::spawn(String::from("")).unwrap();
```
jordens, line 282: This is where you'd use an `Option<String<64>>` don't encode that info in magic (empty) strings.

jordens (changes requested):

nkrackow:
One detail still in the logic now is that I have to re-clear the interlock to re-arm it after I disarmed it. So if I set `armed=false`, I cannot re-arm by setting `armed=true`, but I have to clear the interrupt (even though that didn't trigger. 
I find that OK and if I wanted to be able to directly re-arm I would have to memorize what happened last.

jordens:
> One detail still in the logic now is that I have to re-clear the interlock to re-arm it after I disarmed it. So if I set `armed=false`, I cannot re-arm by setting `armed=true`, but I have to clear the interrupt (even though that didn't trigger). I find that OK and if I wanted to be able to directly re-arm I would have to memorize what happened last.

Maybe then you can merge `armed` and `clear` so that a transition `false` -> `true` on `armed` is required to clear and will implicitly do so.

jordens (changes requested):
A minor tweak. But if you feel you've stewed long enough over this just merge it and refactor in the future. Perfection not required now.

nkrackow:
> > One detail still in the logic now is that I have to re-clear the interlock to re-arm it after I disarmed it. So if I set `armed=false`, I cannot re-arm by setting `armed=true`, but I have to clear the interrupt (even though that didn't trigger). I find that OK and if I wanted to be able to directly re-arm I would have to memorize what happened last.
> 
> Maybe then you can merge `armed` and `clear` so that a transition `false` -> `true` on `armed` is required to clear and will implicitly do so.

I think I'd prefer one Setting for this anyways.

PR #610, opened 2022-08-30
[WIP]: Control Pounder through MQTT

occheung:
## Description
This PR implements control of devices on Pounder through MQTT.

These parameters are configurable through MQTT (topic prefix: `<prefix>/settings/pdh/ch/<n>`).
```rust
pub struct PDHChannel {
    pub osc_frequency: f32,     // Frequency of the IN/OUT DDS pair
    pub out_phase: f32,         // Phase offset of the OUT DDS channel
    pub in_phase_offset: f32,   // Phase offset of the IN DDS channel relative to the OUT DDS channel
    pub amplitude: f32,         // Amplitude of the OUT DDS channel
    pub in_attenuation: f32,    // Attenuation of the signal injected to INx connector
    pub out_attenuation: f32,   // Attenuation of the OUT DDS channel
    pub enabled: bool,          // Amplitude of the IN DDS channel (0.0 or 1.0)
}
```

Temperature reading from Pounder (LM75) & input powers (after attenuators) are reported in telemetry.

I also tried reporting power level through the LED, but I got issues like AttRstN being stuck at low voltage. I have [another branch in my fork](https://github.com/occheung/stabilizer/tree/pounder) that has working LED using the mcp23017 library.

### DIFF
--- a/ad9959/src/lib.rs
+++ b/ad9959/src/lib.rs
@@ -516,7 +516,7 @@ impl<I: Interface> Ad9959<I> {
 /// Represents a means of serializing a DDS profile for writing to a stream.
 pub struct ProfileSerializer {
     // heapless::Vec<u8, 32>, especially its extend_from_slice() is slow
-    data: [u8; 32],
+    data: [u8; 64],
     index: usize,
     // make mode u32 to work around https://github.com/japaric/heapless/issues/305
     mode: u32,
@@ -530,7 +530,7 @@ impl ProfileSerializer {
     pub fn new(mode: Mode) -> Self {
         Self {
             mode: mode as _,
-            data: [0; 32],
+            data: [0; 64],
             index: 0,
         }
     }
--- a/ad9959/src/lib.rs
+++ b/ad9959/src/lib.rs
@@ -568,6 +568,44 @@ impl ProfileSerializer {
         }
     }
 
+    /// Update the system clock configuration.
+    ///
+    /// # Args
+    /// * `channels` - A set of channels to apply the configuration to.
+    /// * `ftw` - If provided, indicates a frequency tuning word for the channels.
+    /// * `pow` - If provided, indicates a phase offset word for the channels.
+    /// * `acr` - If provided, indicates the amplitude control register for the channels. The ACR
+    ///   should be stored in the 3 LSB of the word. Note that if amplitude scaling is to be used,
+    ///   the "Amplitude multiplier enable" bit must be set.
+    #[inline]
+    pub fn update_system_clock(
+        &mut self,
+        reference_clock_frequency: f32,
+        multiplier: u8,
+    ) -> Result<f32, Error> {
+        if multiplier != 1 && !(4..=20).contains(&multiplier) {
+            return Err(Error::Bounds);
+        }
+
+        let frequency = multiplier as f32 * reference_clock_frequency as f32;
+        if frequency > 500_000_000.0f32 {
+            return Err(Error::Frequency);
+        }
+
+        // The enabled channel will be updated after clock reconfig
+        let mut fr1: [u8; 3] = [0, 0, 0];
+
+        // All FR1 fields are kept untouched except VCO & PLL divider
+        // These 2 fields will be modified anyway
+        fr1[0].set_bits(2..=6, multiplier);
+
+        let vco_range = frequency > 255e6;
+        fr1[0].set_bit(7, vco_range);
+
+        self.add_write(Register::FR1, &fr1);
+        Ok(frequency)
+    }
+
     /// Add a register write to the serialization data.
     fn add_write(&mut self, register: Register, value: &[u8]) {
         let data = &mut self.data[self.index..];
--- a/ad9959/src/lib.rs
+++ b/ad9959/src/lib.rs
@@ -46,11 +46,11 @@ pub enum Mode {
 bitflags! {
     /// Specifies an output channel of the AD9959 DDS chip.
     pub struct Channel: u8 {
-        const ONE   = 0b00010000;
-        const TWO   = 0b00100000;
-        const THREE = 0b01000000;
-        const FOUR  = 0b10000000;
-        const ALL   = Self::ONE.bits | Self::TWO.bits | Self::THREE.bits | Self::FOUR.bits;
+        const ZERO  = 0b00010000;
+        const ONE   = 0b00100000;
+        const TWO   = 0b01000000;
+        const THREE = 0b10000000;
+        const ALL   = Self::ZERO.bits | Self::ONE.bits | Self::TWO.bits | Self::THREE.bits;
     }
 }
 
--- a/src/bin/dual-iir.rs
+++ b/src/bin/dual-iir.rs
@@ -47,8 +47,7 @@ use stabilizer::{
         pounder::{
             attenuators::AttenuatorInterface, dds_output::DdsOutput,
             rf_power::PowerMeasurementInterface, Ad9959PdhSettings,
-            Channel as PounderChannel, PDHLockGeneratorConfig,
-            PounderDevices,
+            Channel as PounderChannel, PDHLockGeneratorConfig, PounderDevices,
         },
         signal_generator::{self, SignalGenerator},
         timers::SamplingTimer,
--- a/src/bin/dual-iir.rs
+++ b/src/bin/dual-iir.rs
@@ -44,6 +44,12 @@ use stabilizer::{
         afe::Gain,
         dac::{Dac0Output, Dac1Output, DacCode},
         hal,
+        pounder::{
+            attenuators::AttenuatorInterface, dds_output::DdsOutput,
+            rf_power::PowerMeasurementInterface, Ad9959PdhSettings,
+            Channel as PounderChannel, PDHLockGeneratorConfig,
+            PounderDevices,
+        },
         signal_generator::{self, SignalGenerator},
         timers::SamplingTimer,
         DigitalInput0, DigitalInput1, SystemTimer, Systick, AFE0, AFE1,
@@ -142,6 +148,15 @@ pub struct Settings {
     /// # Value
     /// See [signal_generator::BasicConfig#miniconf]
     signal_generator: [signal_generator::BasicConfig; 2],
+
+    /// Pounder output control
+    ///
+    /// # Path
+    /// `pdh`
+    ///
+    /// # Value
+    /// See [pounder::PDHLockGeneratorConfig#miniconf]
+    pdh: PDHLockGeneratorConfig,
 }
 
 impl Default for Settings {
@@ -165,6 +180,8 @@ impl Default for Settings {
             signal_generator: [signal_generator::BasicConfig::default(); 2],
 
             stream_target: StreamTarget::default(),
+
+            pdh: PDHLockGeneratorConfig::default(),
         }
     }
 }
@@ -183,6 +200,7 @@ mod app {
         settings: Settings,
         telemetry: TelemetryBuffer,
         signal_generator: [SignalGenerator; 2],
+        pounder_devices: Option<PounderDevices>,
     }
 
     #[local]
@@ -192,6 +210,7 @@ mod app {
         afes: (AFE0, AFE1),
         adcs: (Adc0Input, Adc1Input),
         dacs: (Dac0Output, Dac1Output),
+        ddses: Option<DdsOutput>,
         iir_state: [[iir::Vec5<f32>; IIR_CASCADE_LENGTH]; 2],
         generator: FrameGenerator,
         cpu_temp_sensor: stabilizer::hardware::cpu_temp_sensor::CpuTempSensor,
@@ -202,13 +221,19 @@ mod app {
         let clock = SystemTimer::new(|| monotonics::now().ticks() as u32);
 
         // Configure the microcontroller
-        let (stabilizer, _pounder) = hardware::setup::setup(
+        let (stabilizer, pounder) = hardware::setup::setup(
             c.core,
             c.device,
             clock,
             BATCH_SIZE,
             SAMPLE_TICKS,
         );
+
+        let (pounder_devices, dds_output) = if let Some(pounder_dev) = pounder {
+            (Some(pounder_dev.pounder), Some(pounder_dev.dds_output))
+        } else {
+            (None, None)
+        };
 
         let mut network = NetworkUsers::new(
             stabilizer.net.stack,
@@ -243,6 +268,7 @@ mod app {
                         .unwrap(),
                 ),
             ],
+            pounder_devices,
         };
 
         let mut local = Local {
@@ -251,6 +277,7 @@ mod app {
             afes: stabilizer.afes,
             adcs: stabilizer.adcs,
             dacs: stabilizer.dacs,
+            ddses: dds_output,
             iir_state: [[[0.; 5]; IIR_CASCADE_LENGTH]; 2],
             generator,
             cpu_temp_sensor: stabilizer.temperature_sensor,
@@ -401,7 +428,7 @@ mod app {
         }
     }
 
-    #[task(priority = 1, local=[afes], shared=[network, settings, signal_generator])]
+    #[task(priority = 1, local=[afes, ddses], shared=[network, settings, signal_generator, pounder_devices])]
     fn settings_update(mut c: settings_update::Context) {
         let settings = c.shared.network.lock(|net| *net.miniconf.settings());
         c.shared.settings.lock(|current| *current = settings);
@@ -424,6 +451,87 @@ mod app {
                 ),
             }
         }
+
+        // Update PDH channels on Pounder
+        c.shared.pounder_devices.lock(|dev| {
+            if let (Some(ddses), Some(devices)) = (c.local.ddses, dev) {
+                devices
+                    .set_ext_clk(settings.pdh.clock_config.external_clock)
+                    .unwrap();
+
+                let mut builder = ddses.builder();
+
+                let f_sys = builder
+                    .update_system_clock(
+                        settings.pdh.clock_config.reference_clock,
+                        settings.pdh.clock_config.multiplier,
+                    )
+                    .unwrap();
+
+                let Ad9959PdhSettings {
+                    in_channel_dds,
+                    out_channel_dds,
+                } = settings.pdh.ch[0].try_into_dds_config_mu(f_sys).unwrap();
+
+                builder.update_channels(
+                    ad9959::Channel::ZERO,
+                    Some(out_channel_dds.ftw),
+                    Some(out_channel_dds.pow),
+                    Some(out_channel_dds.acr),
+                );
+                builder.update_channels(
+                    ad9959::Channel::ONE,
+                    Some(in_channel_dds.ftw),
+                    Some(in_channel_dds.pow),
+                    Some(in_channel_dds.acr),
+                );
+
+                let Ad9959PdhSettings {
+                    in_channel_dds,
+                    out_channel_dds,
+                } = settings.pdh.ch[1].try_into_dds_config_mu(f_sys).unwrap();
+
+                builder.update_channels(
+                    ad9959::Channel::TWO,
+                    Some(out_channel_dds.ftw),
+                    Some(out_channel_dds.pow),
+                    Some(out_channel_dds.acr),
+                );
+                builder.update_channels(
+                    ad9959::Channel::THREE,
+                    Some(in_channel_dds.ftw),
+                    Some(in_channel_dds.pow),
+                    Some(in_channel_dds.acr),
+                );
+
+                builder.write();
+
+                devices
+                    .set_attenuation(
+                        PounderChannel::Out0,
+                        settings.pdh.ch[0].out_attenuation,
+                    )
+                    .unwrap();
+                devices
+                    .set_attenuation(
+                        PounderChannel::In0,
+                        settings.pdh.ch[0].in_attenuation,
+                    )
+                    .unwrap();
+                devices
+                    .set_attenuation(
+                        PounderChannel::Out1,
+                        settings.pdh.ch[1].out_attenuation,
+                    )
+                    .unwrap();
+                devices
+                    .set_attenuation(
+                        PounderChannel::In1,
+                        settings.pdh.ch[1].in_attenuation,
+                    )
+                    .unwrap();
+            }
+        });
 
         let target = settings.stream_target.into();
         c.shared.network.lock(|net| net.direct_stream(target));
--- a/src/bin/dual-iir.rs
+++ b/src/bin/dual-iir.rs
@@ -537,7 +537,7 @@ mod app {
         c.shared.network.lock(|net| net.direct_stream(target));
     }
 
-    #[task(priority = 1, shared=[network, settings, telemetry], local=[cpu_temp_sensor])]
+    #[task(priority = 1, shared=[network, settings, telemetry, pounder_devices], local=[cpu_temp_sensor])]
     fn telemetry(mut c: telemetry::Context) {
         let telemetry: TelemetryBuffer =
             c.shared.telemetry.lock(|telemetry| *telemetry);
@@ -547,11 +547,30 @@ mod app {
             .settings
             .lock(|settings| (settings.afe, settings.telemetry_period));
 
+        let (pounder_temp, input_powers) =
+            c.shared.pounder_devices.lock(|pounder_dev| {
+                if let Some(dev) = pounder_dev {
+                    let input_powers = [
+                        dev.measure_power(PounderChannel::In0).unwrap(),
+                        dev.measure_power(PounderChannel::In1).unwrap(),
+                    ];
+
+                    (
+                        Some(dev.lm75.read_temperature().unwrap()),
+                        Some(input_powers),
+                    )
+                } else {
+                    (None, None)
+                }
+            });
+
         c.shared.network.lock(|net| {
             net.telemetry.publish(&telemetry.finalize(
                 gains[0],
                 gains[1],
                 c.local.cpu_temp_sensor.get_temperature().unwrap(),
+                pounder_temp,
+                input_powers,
             ))
         });
 
--- a/src/hardware/pounder/dds_output.rs
+++ b/src/hardware/pounder/dds_output.rs
@@ -157,6 +157,16 @@ impl<'a> ProfileBuilder<'a> {
         self
     }
 
+    #[inline]
+    pub fn update_system_clock(
+        &mut self,
+        reference_clock_frequency: f32,
+        multiplier: u8,
+    ) -> Result<f32, ad9959::Error> {
+        self.serializer
+            .update_system_clock(reference_clock_frequency, multiplier)
+    }
+
     /// Write the profile to the DDS asynchronously.
     #[allow(dead_code)]
     #[inline]
--- a/src/hardware/pounder/mod.rs
+++ b/src/hardware/pounder/mod.rs
@@ -129,10 +129,10 @@ impl From<Channel> for ad9959::Channel {
     /// Translate pounder channels to DDS output channels.
     fn from(other: Channel) -> Self {
         match other {
-            Channel::In0 => Self::TWO,
-            Channel::In1 => Self::FOUR,
-            Channel::Out0 => Self::ONE,
-            Channel::Out1 => Self::THREE,
+            Channel::In0 => Self::ONE,
+            Channel::Out0 => Self::ZERO,
+            Channel::In1 => Self::THREE,
+            Channel::Out1 => Self::TWO,
         }
     }
 }
--- a/src/hardware/pounder/mod.rs
+++ b/src/hardware/pounder/mod.rs
@@ -4,6 +4,7 @@ use super::hal;
 use crate::hardware::{shared_adc::AdcChannel, I2c1Proxy};
 use embedded_hal::blocking::spi::Transfer;
 use enum_iterator::Sequence;
+use miniconf::{Miniconf, MiniconfAtomic};
 use serde::{Deserialize, Serialize};
 
 pub mod attenuators;
@@ -91,38 +92,129 @@ impl From<Channel> for GpioPin {
     }
 }
 
-#[derive(Serialize, Deserialize, Copy, Clone, Debug)]
-pub struct DdsChannelState {
-    pub phase_offset: f32,
-    pub frequency: f32,
-    pub amplitude: f32,
-    pub enabled: bool,
-}
-
-#[derive(Serialize, Deserialize, Copy, Clone, Debug)]
-pub struct ChannelState {
-    pub parameters: DdsChannelState,
-    pub attenuation: f32,
-}
-
-#[derive(Serialize, Deserialize, Copy, Clone, Debug)]
-pub struct InputChannelState {
-    pub attenuation: f32,
-    pub power: f32,
-    pub mixer: DdsChannelState,
-}
-
-#[derive(Serialize, Deserialize, Copy, Clone, Debug)]
-pub struct OutputChannelState {
-    pub attenuation: f32,
-    pub channel: DdsChannelState,
-}
-
-#[derive(Serialize, Deserialize, Copy, Clone, Debug)]
+#[derive(Serialize, Deserialize, Copy, Clone, Debug, MiniconfAtomic)]
 pub struct DdsClockConfig {
     pub multiplier: u8,
     pub reference_clock: f32,
     pub external_clock: bool,
+}
+
+impl Default for DdsClockConfig {
+    fn default() -> Self {
+        Self {
+            multiplier: super::design_parameters::DDS_MULTIPLIER,
+            reference_clock: super::design_parameters::DDS_REF_CLK.to_Hz()
+                as f32,
+            external_clock: false,
+        }
+    }
+}
+
+#[derive(Serialize, Deserialize, Copy, Clone, Debug, Miniconf)]
+pub struct PDHLockGeneratorConfig {
+    pub clock_config: DdsClockConfig,
+    pub ch: [PDHChannel; 2],
+}
+
+impl Default for PDHLockGeneratorConfig {
+    fn default() -> Self {
+        Self {
+            clock_config: DdsClockConfig::default(),
+            ch: [PDHChannel::default(); 2],
+        }
+    }
+}
+
+#[derive(Serialize, Deserialize, Copy, Clone, Debug, Miniconf)]
+pub struct PDHChannel {
+    pub osc_frequency: f32,
+    pub out_phase: f32,
+    pub in_phase_offset: f32,
+    pub amplitude: f32,
+    pub in_attenuation: f32,
+    pub out_attenuation: f32,
+    pub enabled: bool,
+}
+
+impl Default for PDHChannel {
+    fn default() -> Self {
+        Self {
+            osc_frequency: 10.0e6,
+            out_phase: 0.0,
+            in_phase_offset: 0.0,
+            amplitude: 0.0,
+            in_attenuation: 31.5,
+            out_attenuation: 31.5,
+            enabled: false,
+        }
+    }
+}
+
+pub struct Ad9959PdhSettings {
+    pub in_channel_dds: Ad9959DdsSettings,
+    pub out_channel_dds: Ad9959DdsSettings,
+}
+
+pub struct Ad9959DdsSettings {
+    pub ftw: u32,
+    pub pow: u16,
+    pub acr: u32,
+}
+
+impl PDHChannel {
+    pub fn try_into_dds_config_mu(
+        self,
+        system_frequency: f32,
+    ) -> Result<Ad9959PdhSettings, Error> {
+        let ftw = Self::frequency_to_ftw(self.osc_frequency, system_frequency)?;
+        Ok(Ad9959PdhSettings {
+            in_channel_dds: Ad9959DdsSettings {
+                ftw,
+                pow: Self::phase_to_pow(self.out_phase + self.in_phase_offset)?,
+                acr: if self.enabled {
+                    Self::amplitude_to_acr(1.0)?
+                } else {
+                    Self::amplitude_to_acr(0.0)?
+                },
+            },
+            out_channel_dds: Ad9959DdsSettings {
+                ftw,
+                pow: Self::phase_to_pow(self.out_phase)?,
+                acr: Self::amplitude_to_acr(self.amplitude)?,
+            },
+        })
+    }
+
+    fn frequency_to_ftw(
+        dds_frequency: f32,
+        system_frequency: f32,
+    ) -> Result<u32, Error> {
+        if dds_frequency < 0.0 || dds_frequency > system_frequency {
+            return Err(Error::Bounds);
+        }
+        // The function for channel frequency is `f_out = FTW * f_s / 2^32`, where FTW is the
+        // frequency tuning word and f_s is the system clock rate.
+        Ok(((dds_frequency as f32 / system_frequency)
+            * 1u64.wrapping_shl(32) as f32) as u32)
+    }
+
+    fn phase_to_pow(phase_turns: f32) -> Result<u16, Error> {
+        Ok((phase_turns * (1 << 14) as f32) as u16 & 0x3FFFu16)
+    }
+
+    fn amplitude_to_acr(amplitude: f32) -> Result<u32, Error> {
+        if !(0.0..=1.0).contains(&amplitude) {
+            return Err(Error::Bounds);
+        }
+
+        if amplitude == 1.0 {
+            return Ok(0x13FF);
+        }
+
+        let amplitude_control: u16 = (amplitude * (1 << 10) as f32) as u16;
+
+        Ok(((amplitude_control & 0x3FF) | (1 << 12)) as u32)
+    }
 }
 
 impl From<Channel> for ad9959::Channel {
--- a/src/net/telemetry.rs
+++ b/src/net/telemetry.rs
@@ -59,6 +59,12 @@ pub struct Telemetry {
 
     /// The CPU temperature in degrees Celsius.
     pub cpu_temp: f32,
+
+    /// The pounder temperature
+    pub pounder_temp: Option<f32>,
+
+    /// The detected RF power into PDH input channels
+    pub pdh_input_powers: Option<[f32; 2]>,
 }
 
 impl Default for TelemetryBuffer {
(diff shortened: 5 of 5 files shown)


occheung:
Added the same Pounder telemetry to `lockin` variant.

@jordens What are the plans of Pounder for `lockin`?

src/bin/dual-iir.rs
```diff
@@ -142,6 +147,15 @@ pub struct Settings {
...
     signal_generator: [signal_generator::BasicConfig; 2],
+
+    /// Pounder output control
+    ///
+    /// # Path
+    /// `pdh`
+    ///
+    /// # Value
+    /// See [PDHLockGeneratorConfig#miniconf]
+    pdh: PDHLockGeneratorConfig,
```
ryan-summers, line 158: This should be wrapped in an `Option`. It should only be `Some(pdh)` if pounder is found to be present. Miniconf will transparently handle the `Some` or `None` variants as "existing" or "not existing" at run time.

ryan-summers: That way we can support for `lockin` with and without pounder present.

ad9959/src/lib.rs
```diff
@@ -530,7 +530,7 @@ impl ProfileSerializer {
     pub fn new(mode: Mode) -> Self {
         Self {
             mode: mode as _,
-            data: [0; 32],
+            data: [0; 64],
```
jordens, line 533: No. Why?

That's more than the qspi fifo.

Instead just recreate the builder and wait/delay for flush.

ad9959/src/lib.rs
```diff
@@ -568,6 +568,44 @@ impl ProfileSerializer {
...
+
+        let frequency = multiplier as f32 * reference_clock_frequency as f32;
+        if frequency > 500_000_000.0f32 {
+            return Err(Error::Frequency);
+        }
+
+        // The enabled channel will be updated after clock reconfig
+        let mut fr1: [u8; 3] = [0, 0, 0];
+
+        // All FR1 fields are kept untouched except VCO & PLL divider
```
jordens, line 598: How do you keep them untouched? Aren't you writing them?

src/bin/dual-iir.rs
```diff
@@ -439,11 +546,30 @@ mod app {
...
+                    (None, None)
+                }
+            });
+
         c.shared.network.lock(|net| {
             net.telemetry.publish(&telemetry.finalize(
                 gains[0],
                 gains[1],
                 c.local.cpu_temp_sensor.get_temperature().unwrap(),
+                pounder_temp,
```
jordens, line 571: No need to shuttle these through `finalize()`. They are not converted with any gains. Continuously appending to the `finalize()` arguments doesn't scale well.
Instead add fields to the `TelemetryBuffer` and let `finalize()` copy everything that doesn't need conversion.
(Same holds for cpu temperature.)

src/net/telemetry.rs
```diff
@@ -59,6 +59,12 @@ pub struct Telemetry {
 
     /// The CPU temperature in degrees Celsius.
     pub cpu_temp: f32,
+
+    /// The pounder temperature
+    pub pounder_temp: Option<f32>,
```
jordens, line 64: Just make Pounder telemetry one `Option`, not two.

src/bin/dual-iir.rs
```diff
@@ -142,6 +147,15 @@ pub struct Settings {
...
     /// See [signal_generator::BasicConfig#miniconf]
     signal_generator: [signal_generator::BasicConfig; 2],
+
+    /// Pounder output control
+    ///
+    /// # Path
+    /// `pdh`
+    ///
+    /// # Value
+    /// See [pounder::PDHLockGeneratorConfig#miniconf]
```
jordens: There is nothing to see there. It's missing documentation.

src/hardware/pounder/mod.rs
```diff
@@ -91,48 +92,139 @@ impl From<Channel> for GpioPin {
...
+    }
+}
+
+#[derive(Serialize, Deserialize, Copy, Clone, Debug, Miniconf)]
+pub struct PDHLockGeneratorConfig {
+    pub clock_config: DdsClockConfig,
+    pub ch: [PDHChannel; 2],
+}
+
+impl Default for PDHLockGeneratorConfig {
```
jordens, line 119: Derive it.

jordens (changes requested):

jordens:
There is no plan to add Pounder to `lockin`. Instead the is FLS.

src/hardware/pounder/mod.rs
```diff
@@ -91,48 +92,139 @@ impl From<Channel> for GpioPin {
...
+    fn default() -> Self {
+        Self {
+            clock_config: DdsClockConfig::default(),
+            ch: [PDHChannel::default(); 2],
+        }
+    }
+}
+
+#[derive(Serialize, Deserialize, Copy, Clone, Debug, Miniconf)]
+pub struct PDHChannel {
```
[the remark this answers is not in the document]
occheung: It is based on the parameters that can be controlled independently in a PDH servo loop.
image
For example, since the oscillator is tied to both the phase modulator and the mixer (through phase shifter), I tied both DDS of the same InOut pair the same.

ad9959/src/lib.rs
occheung: Ok.

Should I keep track of the DDS parameters (or read back), then only update when necessary?

jordens: AFAICT frequency/phase/amplitude all update without transients. Writing them always should be fine. But may well be worth checking on hardware.
NB: Changing the max buffer length would also require changes to the IO_UPDATE timing and would break assumptions about the max duration of such an update elsewhere.

src/hardware/pounder/mod.rs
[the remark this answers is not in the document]
jordens: I know how PDH works. But it's a bad idea to assume that this is the only application of Pounder.
Also since you don't control the relative timing of the frequency updates between in and out, the relative phase isn't deterministic and your assumptions break.

ad9959/src/lib.rs
jordens: Then the comment should state that. As it is now it sounds like you are actively ensuring that the bits remain the same.

ad9959/src/lib.rs
```diff
@@ -568,6 +568,44 @@ impl ProfileSerializer {
...
+        &mut self,
+        reference_clock_frequency: f32,
+        multiplier: u8,
+    ) -> Result<f32, Error> {
+        if multiplier != 1 && !(4..=20).contains(&multiplier) {
+            return Err(Error::Bounds);
+        }
+
+        let frequency = multiplier as f32 * reference_clock_frequency as f32;
+        if frequency > 500_000_000.0f32 {
```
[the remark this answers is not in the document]
occheung: The nominal minimum system clock frequency stated on the datasheet is literally 0.
However, there is a 10 MHz minimum for the reference clock frequency. Checking this minimum seems more appropriate.

src/bin/dual-iir.rs
```diff
@@ -165,6 +179,8 @@ impl Default for Settings {
             signal_generator: [signal_generator::BasicConfig::default(); 2],
 
             stream_target: StreamTarget::default(),
+
+            pdh: PDHLockGeneratorConfig::default(),
```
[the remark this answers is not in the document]
occheung: Well Pounder is said to be the PDH lock generator, so I geared towards making it (and the PR by extension) based on the context of PDH.
We can just drop this context and use the DDS configuration/state structs that were already in `pounder`.

src/bin/dual-iir.rs
occheung: [f9459ae](https://github.com/quartiq/stabilizer/pull/610/commits/f9459aee8fdcd0f0ec6efc95cfac0f0d46d9229e)

ad9959/src/lib.rs
occheung: [9720452](https://github.com/quartiq/stabilizer/pull/610/commits/97204524007b541a59a9eff4d5f6938d7633b90d)

src/hardware/pounder/mod.rs
```diff
@@ -91,48 +92,139 @@ impl From<Channel> for GpioPin {
...
+    fn phase_to_pow(phase_turns: f32) -> Result<u16, Error> {
+        Ok((phase_turns * (1 << 14) as f32) as u16 & 0x3FFFu16)
+    }
+
+    fn amplitude_to_acr(amplitude: f32) -> Result<u32, Error> {
+        if !(0.0..=1.0).contains(&amplitude) {
+            return Err(Error::Bounds);
+        }
+
+        if amplitude == 1.0 {
```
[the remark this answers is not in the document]
occheung: 3509991

src/hardware/pounder/mod.rs
```diff
@@ -91,48 +92,139 @@ impl From<Channel> for GpioPin {
...
+                ftw,
+                pow: Self::phase_to_pow(self.out_phase)?,
+                acr: Self::amplitude_to_acr(self.amplitude)?,
+            },
+        })
+    }
+
+    fn frequency_to_ftw(
+        dds_frequency: f32,
+        system_frequency: f32,
```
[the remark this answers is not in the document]
occheung: 58ee3ee
I changed `f_sys` to `system_frequency` instead.

src/hardware/pounder/mod.rs
```diff
@@ -91,48 +92,139 @@ impl From<Channel> for GpioPin {
...
+    ) -> Result<u32, Error> {
+        if dds_frequency < 0.0 || dds_frequency > system_frequency {
```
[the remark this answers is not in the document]
occheung: 077b634

src/hardware/pounder/mod.rs
```diff
@@ -91,48 +92,139 @@ impl From<Channel> for GpioPin {
...
+    pub acr: u32,
 }
 
-#[derive(Serialize, Deserialize, Copy, Clone, Debug)]
-pub struct DdsClockConfig {
-    pub multiplier: u8,
-    pub reference_clock: f32,
-    pub external_clock: bool,
+impl PDHChannel {
+    pub fn try_into_dds_config_mu(
```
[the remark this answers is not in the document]
occheung: We need the system clock frequency to derive the frequency tuning word, which should not be an attribute of a singular DDS channel.

occheung:
Most of the system clock calculation & parameter conversions are direct copy of those in `ad9959`. What about this:

- Make associated functions in `ad9959` for frequency/amplitude/phase to ftw/acr/pow conversions.
- Refactor existing methods in `ad9959` that use these conversions.
- Fix the frequency boundary checks for clock configuration functions in `ad9959`.

src/bin/dual-iir.rs
```diff
@@ -202,14 +220,20 @@ mod app {
...
-        let (stabilizer, _pounder) = hardware::setup::setup(
+        let (stabilizer, pounder) = hardware::setup::setup(
             c.core,
             c.device,
             clock,
             BATCH_SIZE,
             SAMPLE_TICKS,
         );
 
+        let (pounder_devices, dds_output) = if let Some(pounder_dev) = pounder {
```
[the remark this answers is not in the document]
occheung: It was to separate the DDS modules as a local resource instead of global resource. DDS is only needed when settings are updated.
Though it is not like everything in the `PounderDevices` struct is needed in multiple tasks.

ad9959/src/lib.rs
occheung: > Instead just recreate the builder and wait/delay for flush.
Done. The QSPI enters indirect write mode after `DdsOutput` initializes and starts streaming. So when writing to a full FIFO, the write operation will stall until there are enough space for the data.

src/bin/dual-iir.rs
```diff
@@ -425,26 +457,118 @@ mod app {
...
 
-    #[task(priority = 1, shared=[network, settings, telemetry], local=[cpu_temp_sensor])]
+    #[task(priority = 1, shared=[network, settings, telemetry, pounder], local=[cpu_temp_sensor])]
     fn telemetry(mut c: telemetry::Context) {
-        let telemetry: TelemetryBuffer =
+        let mut telemetry: TelemetryBuffer =
             c.shared.telemetry.lock(|telemetry| *telemetry);
 
+        telemetry.cpu_temp = c.local.cpu_temp_sensor.get_temperature().unwrap();
+        telemetry.pounder = c.shared.pounder.lock(|pounder| {
```
jordens, line 539: This is convoluted. By default `telemetry.pounder` is `None`. No need to reset it all the time.
And can't this code go into `PounderTelemetry` or `PounderDevices` so that you don't have to copy it everywhere?

src/bin/dual-iir.rs
```diff
@@ -425,26 +457,118 @@ mod app {
             }
         }
 
+        // Update Pounder configurations
```
jordens, line 460: This should also go into `PounderSettings` and/or `Pounder`. Don't spread it into every application.

ad9959/src/lib.rs
```diff
@@ -568,6 +609,58 @@ impl ProfileSerializer {
...
+    /// Update a number of channels with fully defined profile settings.
+    ///
+    /// # Args
+    /// * `channels` - A set of channels to apply the configuration to.
+    /// * `profile` - The complete DDS profile, which defines the frequency tuning word,
+    ///   amplitude control register & the phase offset word of the channels. Note that the ACR
+    ///   should be stored in the 3 LSB of the word. If amplitude scaling is to be used, the
+    ///   "Amplitude multiplier enable" bit must be set.
+    #[inline]
+    pub fn update_channels_with_profile(
```
jordens, line 621: There is no point in duplicating the code in `update_channels`.
This method is not needed.

ad9959/src/lib.rs
```diff
@@ -568,6 +609,58 @@ impl ProfileSerializer {
...
+        self.add_write(Register::CPOW0, &profile.pow.to_be_bytes());
+        self.add_write(Register::ACR, &profile.acr.to_be_bytes()[1..]);
+    }
+
+    /// Update the system clock configuration.
+    ///
+    /// # Args
+    /// * `reference_clock_frequency` - The reference clock frequency provided to the AD9959 core.
+    /// * `multiplier` - The frequency multiplier of the system clock. Must be 1 or 4-20.
+    #[inline]
```
jordens, line 638: No inline needed here.

ad9959/src/lib.rs
```diff
@@ -568,6 +609,58 @@ impl ProfileSerializer {
...
+    pub fn update_system_clock(
+        &mut self,
+        reference_clock_frequency: f32,
+        multiplier: u8,
+    ) -> Result<f32, Error> {
+        let frequency = Ad9959::<()>::validate_system_clock(
```
jordens, line 644: No. That's a layering violation. Don't do the conversions and these checks in the serializer but elsewhere.

src/hardware/pounder/mod.rs
```diff
@@ -91,48 +93,87 @@ impl From<Channel> for GpioPin {
...
 }
 
-#[derive(Serialize, Deserialize, Copy, Clone, Debug)]
-pub struct ChannelState {
-    pub parameters: DdsChannelState,
-    pub attenuation: f32,
+impl Default for DdsChannelConfig {
+    fn default() -> Self {
+        Self {
+            frequency: 10.0e6,
```
jordens, line 106: Why 10 MHz? Just 0.

src/hardware/pounder/mod.rs
```diff
@@ -91,48 +93,87 @@ impl From<Channel> for GpioPin {
...
 pub struct DdsClockConfig {
     pub multiplier: u8,
     pub reference_clock: f32,
     pub external_clock: bool,
 }
 
+impl Default for DdsClockConfig {
+    fn default() -> Self {
+        Self {
+            multiplier: super::design_parameters::DDS_MULTIPLIER,
```
jordens, line 154: If you make this configurable, then these are defnitely not "design parameters" anymore. You need to remove them if someone relies on them being constants.

src/hardware/pounder/mod.rs
```diff
@@ -91,48 +93,87 @@ impl From<Channel> for GpioPin {
...
+            reference_clock: super::design_parameters::DDS_REF_CLK.to_Hz()
+                as f32,
+            external_clock: false,
+        }
+    }
+}
+
+#[derive(Serialize, Deserialize, Copy, Clone, Debug, Default, Miniconf)]
+pub struct PounderConfig {
+    pub dds_clock: DdsClockConfig,
```
jordens, line 164: just `pub clock: ClockConfig` no need to proliferate DDS everywhere.

src/hardware/pounder/mod.rs
```diff
@@ -91,48 +93,87 @@ impl From<Channel> for GpioPin {
...
+        }
+    }
 }
 
-#[derive(Serialize, Deserialize, Copy, Clone, Debug)]
-pub struct InputChannelState {
-    pub attenuation: f32,
-    pub power: f32,
-    pub mixer: DdsChannelState,
+impl DdsChannelConfig {
```
jordens, line 113: Make a real `impl From<(ClockConfig, ChannelConfig)> for Profile` and do the conversion in that `impl`, not in `Profile::new()`.

src/hardware/pounder/dds_output.rs
```diff
@@ -157,6 +157,44 @@ impl<'a> ProfileBuilder<'a> {
...
+    ///   should be stored in the 3 LSB of the word. If amplitude scaling is to be used, the
+    ///   "Amplitude multiplier enable" bit must be set.
+    #[inline]
+    pub fn set_system_clock(
+        &mut self,
+        reference_clock_frequency: f32,
+        multiplier: u8,
+    ) -> Result<&mut Self, ad9959::Error> {
+        self.serializer
+            .update_system_clock(reference_clock_frequency, multiplier)?;
```
jordens, line 194: Also call that `set_system_clock` then. These are just wrappers.

src/hardware/pounder/dds_output.rs
```diff
@@ -157,6 +157,44 @@ impl<'a> ProfileBuilder<'a> {
...
+    ///   Note that the ACR should be stored in the 3 LSB of the word.
+    ///   If amplitude scaling is to be used, the "Amplitude multiplier enable" bit must be set.
+    #[inline]
+    pub fn update_channels_with_profile(
+        &mut self,
+        channels: Channel,
+        profile: DdsProfile,
+    ) -> &mut Self {
+        self.serializer
+            .update_channels_with_profile(channels, profile);
```
jordens, line 175: Just call `update_channels()` instead.

src/bin/dual-iir.rs
```diff
@@ -425,26 +457,118 @@ mod app {
...
+        c.shared.pounder.lock(|pounder| {
+            if let (
```
jordens, line 462: I'm having trouble reading this. Please find a way to split that `let`.

jordens (changes requested):
Sorry the review took so long. There are still a couple of correctness and maintainability things we need to sort out in this PR.

ad9959/src/lib.rs
```diff
@@ -606,3 +699,35 @@ impl ProfileSerializer {
...
+    /// * `amplitude` - A normalized amplitude setting [0, 1].
+    /// * `system_clock_frequency` - The system clock frequency of Ad9959.
+    pub fn new(
+        frequency: f32,
+        phase_turns: f32,
+        amplitude: f32,
+        system_clock_frequency: f32,
+    ) -> Result<Self, Error> {
+        Ok(DdsProfile {
+            ftw: Ad9959::<()>::frequency_to_ftw(
```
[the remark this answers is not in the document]
occheung: So where should I be putting the converter methods (like `frequency_to_ftw`)? `Ad9959` is using them, otherwise I am duplicating all these code for the builder/serializer.

If we expose `Ad9959` back in `Pounder`, it sounds very much against the idea that we are freezing the `Ad9959` after destructing it and snatching the QSPI interface from it.

src/hardware/pounder/attenuators.rs
```diff
@@ -68,6 +67,9 @@ pub trait AttenuatorInterface {
         // care) would contain erroneous data.
         let attenuation_code = (!channels[channel as usize]) >> 2;
 
+        // The write-back transfer is performed. Staging register is now restored.
+        self.transfer_attenuators(&mut channels)?;
```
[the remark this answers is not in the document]
occheung: The first transfer reads back the attenuation value and replace it with 0s (31.5 dB). When you trigger the second transfer immediately, you read back the the 0s (31.5 dB) and restore the original attenuation before we performed the read.

The original code does the read back-to-back, so you will always read 31.5 dB no matter what, because it is the dummy value that we set a few lines before this.

I only used the value to check the equality of attenuation before overwriting it, so all it does is do prevent unnecessary overwriting.

ad9959/src/lib.rs
```diff
@@ -406,27 +396,11 @@ impl<I: Interface> Ad9959<I> {
...
-            acr[2] = masked_control.to_be_bytes()[1];
-
-            // Enable the amplitude multiplier
-            acr[1].set_bit(4, true);
-        }
-
-        self.modify_channel(channel, Register::ACR, &acr)?;
+        let acr = Self::amplitude_to_acr(amplitude)?;
+        // Isolate the amplitude scaling factor from ACR
+        let amplitude_control = acr | ((1 << 10) - 1);
```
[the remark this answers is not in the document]
occheung: Alright that looks really dubious...

Isolation should be bitwise AND, and it should consider the case where the amplitude multiplier is off.

src/hardware/pounder/mod.rs
occheung: This would be implementing a trait for a struct in a foreign crate, `impl` it on a wrapper instead.

ad9959/src/lib.rs
```diff
@@ -216,23 +216,16 @@ impl<I: Interface> Ad9959<I> {
...
-            return Err(Error::Frequency);
-        }
-
         // TODO: Update / disable any enabled channels?
         let mut fr1: [u8; 3] = [0, 0, 0];
         self.read(Register::FR1, &mut fr1)?;
         fr1[0].set_bits(2..=6, multiplier);
 
-        let vco_range = frequency > 255e6;
+        let vco_range = frequency > 200e6;
```
ryan-summers, line 228: Please move the 200e6 to a `const` variable with a name at the top of the file. Also, can you please document where this limit is coming from? It's not clear to me (nor can I remember why the original 255e6 was selected)

ad9959/src/lib.rs
```diff
@@ -404,29 +395,18 @@ impl<I: Interface> Ad9959<I> {
...
-
-        self.modify_channel(channel, Register::ACR, &acr)?;
+        let acr = amplitude_to_acr(amplitude)?;
+        let amplitude = if (acr & (1 << 12)) != 0 {
+            // Isolate the amplitude scaling factor from ACR
+            (acr & ((1 << 10) - 1)) as f32 / (1 << 10) as f32
+        } else {
+            // Amplitude is always at full-scale with amplitude multiplier disabled
+            1.0
+        };
```
ryan-summers, line 405: It really doesn't feel to me like the changes here are an improvement over the previous implementation - in fact, I find this implementation _much_ more opaque. What are the functional changes here? If there are none, let's leave this code as-is.

ad9959/src/lib.rs
```diff
@@ -468,16 +448,16 @@ impl<I: Interface> Ad9959<I> {
 
         // The function for channel frequency is `f_out = FTW * f_s / 2^32`, where FTW is the
         // frequency tuning word and f_s is the system clock rate.
-        let tuning_word: u32 = ((frequency / self.system_clock_frequency())
-            * 1u64.wrapping_shl(32) as f32)
-            as u32;
+        let tuning_word: u32 =
+            ((frequency as f32 / self.system_clock_frequency())
```
ryan-summers, line 452: See Github actions lint - casting f32->f32

ad9959/src/lib.rs
```diff
@@ -468,16 +448,16 @@ impl<I: Interface> Ad9959<I> {
...
+                * 1u64.wrapping_shl(32) as f32) as u32;
 
         self.modify_channel(
             channel,
             Register::CFTW0,
             &tuning_word.to_be_bytes(),
         )?;
-        Ok((tuning_word as f32 / 1u64.wrapping_shl(32) as f32)
+        Ok((tuning_word as f32 / (1u64 << 32) as f32)
```
ryan-summers, line 460: I'd prefer to be explicit that we're doing a wrapping_shl() call. I've had shifts be implicitly modified due to compiler semantics before, and explicit is better than implicit.

ad9959/src/lib.rs
```diff
@@ -511,6 +491,72 @@ impl<I: Interface> Ad9959<I> {
...
+///
+/// Returns:
+/// The system clock frequency to be configured.
+pub fn validate_clocking(
+    reference_clock_frequency: f32,
+    multiplier: u8,
+) -> Result<f32, Error> {
+    if multiplier != 1 && !(4..=20).contains(&multiplier)
+        || (multiplier != 1 && reference_clock_frequency < 10e6)
+        || reference_clock_frequency < 1e6
```
ryan-summers, line 508: Please break out (and document) each of this conditionals. It's not clear to me what specific conditions each one is testing for.

ad9959/src/lib.rs
```diff
@@ -511,6 +491,72 @@ impl<I: Interface> Ad9959<I> {
...
+    {
+        return Err(Error::Bounds);
+    }
+
+    let frequency = multiplier as f32 * reference_clock_frequency;
+    if !(255e6..=500e6).contains(&frequency)
+        && !(100e6..=160e6).contains(&frequency)
+        && (multiplier != 1 || !(0.0..100e6).contains(&frequency))
```
ryan-summers, line 516: Similar to above - please break this out into multiple conditionals and document why each conditional is required. Also, break the magic numbers out into named constants at the file-scope so it's clear what they represent (and potentially could be combined).

ad9959/src/lib.rs
```diff
@@ -511,6 +491,72 @@ impl<I: Interface> Ad9959<I> {
...
+    }
+
+    Ok(frequency)
+}
+
+pub fn frequency_to_ftw(
+    dds_frequency: f32,
+    system_clock_frequency: f32,
+) -> Result<u32, Error> {
+    if !(0.0..=(system_clock_frequency / 2.0)).contains(&dds_frequency) {
```
ryan-summers, line 528: Please mention why this check is being conducted (I assume it's because of nyquist). This function could also use some documentation

ad9959/src/lib.rs
```diff
@@ -511,6 +491,72 @@ impl<I: Interface> Ad9959<I> {
...
+        return Err(Error::Bounds);
+    }
+    // The function for channel frequency is `f_out = FTW * f_s / 2^32`, where FTW is the
+    // frequency tuning word and f_s is the system clock rate.
+    Ok(((dds_frequency / system_clock_frequency) * (1u64 << 32) as f32) as u32)
+}
+
+pub fn phase_to_pow(phase_turns: f32) -> Result<u16, Error> {
```
ryan-summers, line 536: Add function documentation

ad9959/src/lib.rs
```diff
@@ -511,6 +491,72 @@ impl<I: Interface> Ad9959<I> {
...
+    Ok((phase_turns * (1 << 14) as f32) as u16 & ((1 << 14) - 1))
```
ryan-summers, line 537: I think this conversion could definitely use some documentation. It feels like we could also just simplify it. Why is (1<< 14) - 1 better than just `0x3FFF`?

ad9959/src/lib.rs
```diff
@@ -511,6 +491,72 @@ impl<I: Interface> Ad9959<I> {
...
+}
+
+pub fn amplitude_to_acr(amplitude: f32) -> Result<u32, Error> {
```
ryan-summers, line 540: Please document this function

ad9959/src/lib.rs
```diff
@@ -511,6 +491,72 @@ impl<I: Interface> Ad9959<I> {
...
+    if !(0.0..=1.0).contains(&amplitude) {
+        return Err(Error::Bounds);
+    }
+
+    let amplitude_control: u16 = (amplitude * (1 << 10) as f32) as u16;
+
+    // Enable the amplitude multiplier for the channel if required. The amplitude control has
+    // full-scale at 0x3FF (amplitude of 1), so the multiplier should be disabled whenever
+    // full-scale is used.
+    let acr = if amplitude_control < (1 << 10) {
```
ryan-summers, line 550: This is a pretty opaque way to figure out if the amplitude multipler is required. Why not just check to see if `amplitude` is 1 or not. That makes it clear that the ACR is set if you use any non-1 amplitude multiplier, as opposed to checking against some magic 1<<10 constant.

ad9959/src/lib.rs
```diff
@@ -566,6 +612,33 @@ impl ProfileSerializer {
...
+        multiplier: u8,
+    ) -> Result<f32, Error> {
+        let frequency = reference_clock_frequency * multiplier as f32;
+
+        // The enabled channel will be updated after clock reconfig
+        let mut fr1 = [0u8; 3];
+
+        // The ad9959 crate does not modify FR1[0:17]. These bits keep their default value.
+        // These bits by default are 0.
+        fr1[0].set_bits(2..=6, multiplier);
```
ryan-summers, line 632: What are the other bits of FR1? Can we be sure we won't modify them later in some other portion of the code? I don't like baking assumptions like this into the software. Say definitively why we won't touch the other bits, or just read the register instead.

If we can't read the register, document why and add this to the function limitation docs.

ad9959/src/lib.rs
```diff
@@ -566,6 +612,33 @@ impl ProfileSerializer {
...
+
+        // Frequencies within the VCO forbidden range (160e6, 255e6) are already rejected.
+        let vco_range = frequency > 200e6;
```
ryan-summers, line 635: Document this magic value and pull it out into a `const`. The comment immediately above this doesn't seem to make any sense with the 200e6 value.

src/hardware/pounder/attenuators.rs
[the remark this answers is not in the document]
ryan-summers: I think the better approach is to copy `channels` into a new buffer for the write-back. That way, we keep around the `channels` value (which mimics what's actually in the staging register). Keeping them `channels` valid for only 1 line of code seems like a quick foot-gun. for the future.

src/hardware/pounder/mod.rs
```diff
@@ -91,40 +93,84 @@ impl From<Channel> for GpioPin {
...
+            external_clock: false,
+        }
+    }
+}
+
+#[derive(Copy, Clone, Debug, Default, Miniconf)]
+pub struct PounderConfig {
+    pub clock: ClockConfig,
+    #[miniconf(defer)]
+    pub in_ch: miniconf::Array<ChannelConfig, 2>,
```
ryan-summers, line 169: Just call them `in` and `out`, as this mimics how they are labeled on the front pannel. Alternatively, `channel_in` and `channel_out` would be fine.

src/hardware/setup.rs
```diff
@@ -133,6 +143,80 @@ pub struct PounderDevices {
...
+                .set_ext_clk(settings.clock.external_clock)
+                .unwrap();
+
+            self.dds_output
+                .builder()
+                .set_system_clock(
+                    settings.clock.reference_clock,
+                    settings.clock.multiplier,
+                )
+                .unwrap()
```
ryan-summers, line 169: Won't this panic if the user provides invalid settings? We definitely don't want that. Same with above w.r.t. clock settings

src/hardware/setup.rs
```diff
@@ -133,6 +143,80 @@ pub struct PounderDevices {
...
+                PounderChannel::In0,
+                PounderChannel::In1,
+                PounderChannel::Out0,
+                PounderChannel::Out1,
+            ])
+        {
+            let dds_profile = ProfileWrapper::from((
+                clocking.clone(),
+                channel_config.clone(),
+            ))
```
ryan-summers, line 186: See clippy lints - don't need to call `clone` on these

src/net/telemetry.rs
```diff
@@ -38,6 +38,8 @@ pub struct TelemetryBuffer {
     pub dacs: [DacCode; 2],
     /// The latest digital input states during processing.
     pub digital_inputs: [bool; 2],
+    /// The latest measured status related to pounder. See [PounderTelemetry].
+    pub pounder: Option<PounderTelemetry>,
```
ryan-summers, line 42: This doesn't need to exist in `TelemetryBuffer`, as there's no need to maintain the pounder telemetry at all points in the app like there is a need for the dacs and IOs. Instead, just have it exist in the final `Telemetry` structure.

ryan-summers (changes requested):
This is looking much better! Most of my comments are related to documentation and clarifying magic numbers in various files.

There's quite a few things Robert pointed out that still need clean up as well

src/hardware/setup.rs
```diff
@@ -133,6 +143,80 @@ pub struct PounderDevices {
...
+            if self.pounder.get_attenuation(pounder_channel).unwrap()
+                != channel_config.attenuation
+            {
+                self.pounder
+                    .set_attenuation(
+                        pounder_channel,
+                        channel_config.attenuation,
+                    )
+                    .unwrap();
+            }
```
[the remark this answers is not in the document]
occheung: Then should we just have a function that sets attenuation of all channels in 1 go? We are currently repeating the SPI write 4 times when only 1 write is needed.

ad9959/src/lib.rs
occheung: Documented. Its selection is somewhat arbitrary.

src/bin/dual-iir.rs
```diff
@@ -431,15 +461,30 @@ mod app {
...
         }
 
+        // Update Pounder configurations
+        c.shared.pounder.lock(|pounder| {
+            if let (Some(pounder), Some(pounder_settings), Some(clocking)) =
+                (pounder, settings.pounder.as_ref(), c.local.dds_clock_state)
+            {
+                pounder.update_dds(*pounder_settings, clocking);
+            }
+        });
```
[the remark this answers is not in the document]
occheung: All 3 of them should be either `Some` together or `None` together. We can just check if 1 of them is a `Some` (maybe `pounder`), and directly get the inner values of all 3 if we found a `Some`.

ad9959/src/lib.rs
```diff
@@ -4,6 +4,22 @@ use bit_field::BitField;
 use bitflags::bitflags;
 use embedded_hal::{blocking::delay::DelayUs, digital::v2::OutputPin};
 
+/// An arbitrary frequency that sits between the 160 MHz to 255 MHz zone, where system clock
+/// frequency is not supported by both high frequency and low frequency VCO settings.
+const VCO_MIN_FREQUENCY: f32 = 200e6;
```
ryan-summers, line 9: We don't allow configuration in the 160-255MHz range based on the clocking checks below. We should remove this constant and just use either the min (high gain) or max (low gain) as a switch for the VCO gain bit.

ad9959/src/lib.rs
```diff
@@ -4,6 +4,22 @@ use bit_field::BitField;
...
+/// The minimum reference clock input frequency with REFCLK multiplier enabled.
+const MIN_MULTIPLIED_REFCLK_FREQUENCY: f32 = 10e6;
+/// The maximum system clock frequency with high frequency range VCO.
+const MAX_VCO_OUTPUT_FREQUENCY: f32 = 500e6;
+/// The minimum system clock frequency with high frequency range VCO.
+const MIN_VCO_OUTPUT_FREQUENCY: f32 = 255e6;
+/// The maximum system clock frequency with low frequency range VCO.
+const MAX_OUTPUT_FREQUENCY: f32 = 160e6;
+/// The minimum system clock frequency with low frequency range VCO.
+const MIN_OUTPUT_FREQUENCY: f32 = 100e6;
```
ryan-summers, line 21:
```suggestion
/// The system clock frequency range with low gain configured for the internal VCO.
const LOW_GAIN_VCO_RANGE: Range<f32> = core::ops::Range { start: 100e6, end: 160e6 };
```

ad9959/src/lib.rs
```diff
@@ -4,6 +4,22 @@ use bit_field::BitField;
...
+/// frequency is not supported by both high frequency and low frequency VCO settings.
+const VCO_MIN_FREQUENCY: f32 = 200e6;
+/// The minimum reference clock input frequency with REFCLK multiplier disabled.
+const MIN_REFCLK_FREQUENCY: f32 = 1e6;
+/// The minimum reference clock input frequency with REFCLK multiplier enabled.
+const MIN_MULTIPLIED_REFCLK_FREQUENCY: f32 = 10e6;
+/// The maximum system clock frequency with high frequency range VCO.
+const MAX_VCO_OUTPUT_FREQUENCY: f32 = 500e6;
+/// The minimum system clock frequency with high frequency range VCO.
+const MIN_VCO_OUTPUT_FREQUENCY: f32 = 255e6;
```
ryan-summers, line 17:
```suggestion
/// The system clock frequency range with high gain configured for the internal VCO.
const HIGH_GAIN_VCO_RANGE: Range<f32> = core::ops::Range { start: 255e6, end: 500e6 };
```

ad9959/src/lib.rs
```diff
@@ -511,6 +518,114 @@ impl<I: Interface> Ad9959<I> {
...
+    // If the REFCLK multiplier is enabled, the REFCLK frequency must be at least 10 MHz.
+    if multiplier != 1
+        && reference_clock_frequency < MIN_MULTIPLIED_REFCLK_FREQUENCY
+    {
+        return Err(Error::Bounds);
+    }
+    // The REFCLK frequency must be at least 1 MHz with REFCLK multiplier disabled.
+    if multiplier == 1 && reference_clock_frequency < MIN_REFCLK_FREQUENCY {
+        return Err(Error::Bounds);
+    }
```
ryan-summers, line 547: Move this to the top of the function and remove the check for `multiplier == 1`. In reality, 1MHz is the absolute lowest allowed REFCLK, but the limit _might_ be higher. We never should allow a lower refclk.

ad9959/src/lib.rs
```diff
@@ -511,6 +518,114 @@ impl<I: Interface> Ad9959<I> {
...
+/// Returns:
+/// The corresponding phase offset word.
+pub fn phase_to_pow(phase_turns: f32) -> Result<u16, Error> {
+    Ok((phase_turns * (1 << 14) as f32) as u16 & 0x3FFFu16)
+}
+
+/// Convert amplitude into amplitude control register values.
+///
+/// Arguments:
+/// * `amplitude` - The amplitude scaling factor of a DDS channel.
```
ryan-summers, line 605:
```suggestion
/// * `amplitude` - The normalized amplitude of a DDS channel.
```

ad9959/src/lib.rs
```diff
@@ -511,6 +518,114 @@ impl<I: Interface> Ad9959<I> {
...
+    }
+    // The function for channel frequency is `f_out = FTW * f_s / 2^32`, where FTW is the
+    // frequency tuning word and f_s is the system clock rate.
+    Ok(((dds_frequency / system_clock_frequency) * (1u64 << 32) as f32) as u32)
+}
+
+/// Convert phase into phase offset word.
+///
+/// Arguments:
+/// * `phase_turns` - The phase of a DDS channel.
```
ryan-summers, line 594:
```suggestion
/// * `phase_turns` - The normalized number of phase turns of a DDS channel.
```

ad9959/src/lib.rs
```diff
@@ -511,6 +518,114 @@ impl<I: Interface> Ad9959<I> {
...
+
+    // Enable the amplitude multiplier for the channel if required. The amplitude control has
+    // full-scale at 0x3FF (amplitude of 1), so the multiplier should be disabled whenever
+    // full-scale is used.
+    let acr = if amplitude != 1.0 {
+        // Enable the amplitude multiplier
+        (amplitude_control & 0x3FF) | (1 << 12)
+    } else {
+        0
+    };
```
ryan-summers, line 624:
```suggestion
    let acr: u32 = 0u32.set_bits(0..=9, (amplitude * (1 << 10) as f32) as u16).set_bit(12, amplitude != 1.0);
```

ryan-summers: Use the `bit_field` API: https://docs.rs/bit_field/0.10.1/bit_field/

src/bin/dual-iir.rs
```diff
@@ -199,6 +214,7 @@ mod app {
         adcs: (Adc0Input, Adc1Input),
         dacs: (Dac0Output, Dac1Output),
         iir_state: [[iir::Vec5<f32>; IIR_CASCADE_LENGTH]; 2],
+        dds_clock_state: Option<ClockConfig>,
```
ryan-summers, line 217: This never actually changes in the code. Why do we store it as a `Resource`? It seems like this should just exist within `Pounder` instead.

src/net/telemetry.rs
```diff
@@ -58,6 +58,23 @@ pub struct Telemetry {
...
+/// # Note
+/// This structure should be generated on-demand by the buffer when required to minimize conversion
+/// overhead.
+#[derive(Copy, Clone, Serialize)]
+pub struct PounderTelemetry {
+    /// The Pounder temperature in degree Celsius
+    pub temperature: f32,
+
+    /// The detected RF power into IN channels
+    pub input_powers: [f32; 2],
```
ryan-summers, line 77:
```suggestion
    pub input_power: [f32; 2],
```

It's already plural because it's an array, naming it `input_powers` implies each channel has multiple power measurements

src/net/telemetry.rs
```diff
@@ -58,6 +58,23 @@ pub struct Telemetry {
...
+}
+
+/// The structure that holds the telemetry related to Pounder.
+///
+/// # Note
+/// This structure should be generated on-demand by the buffer when required to minimize conversion
+/// overhead.
+#[derive(Copy, Clone, Serialize)]
+pub struct PounderTelemetry {
+    /// The Pounder temperature in degree Celsius
```
ryan-summers, line 73:
```suggestion
    /// The Pounder temperature in degrees Celsius
```

src/hardware/setup.rs
```diff
@@ -133,6 +143,85 @@ pub struct PounderDevices {
...
+        PounderTelemetry {
+            temperature: self.pounder.lm75.read_temperature().unwrap(),
+            input_powers: [
+                self.pounder.measure_power(PounderChannel::In0).unwrap(),
+                self.pounder.measure_power(PounderChannel::In1).unwrap(),
+            ],
+        }
+    }
+}
+
```
ryan-summers, line 224: This belongs in `hardware/pounder/` somewhere, not the setup utilities.

ryan-summers (changes requested):
Looking better, but there's still a few things to clean up. Thanks for working on this. I think there's also some outstanding conversations form @jordens and I earlier as well that still need to be addressed.

src/hardware/setup.rs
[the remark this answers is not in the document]
ryan-summers: It really just comes down to how often we're updating attenuation and when we do it. If we need the efficiency, then by all means write all of them in one go, but if it doesn't matter, lets leave what we have in place and open an issue for the future

ad9959/src/lib.rs
ryan-summers: I talk about this below, but we should just use the min of the high VCO gain config or max of the low VCO gain config constants instead of this arbitrary value.

PR #612, opened 2022-09-01
change buffer size and log error

nkrackow:
Tested with a smaller buffer size and I get the error log as expected.

### DIFF
--- a/src/net/telemetry.rs
+++ b/src/net/telemetry.rs
@@ -20,7 +20,7 @@ use minimq::embedded_nal::IpAddr;
 
 /// The telemetry client for reporting telemetry data over MQTT.
 pub struct TelemetryClient<T: Serialize> {
-    mqtt: minimq::Minimq<NetworkReference, SystemTimer, 512, 1>,
+    mqtt: minimq::Minimq<NetworkReference, SystemTimer, 1024, 1>,
     telemetry_topic: String<128>,
     _telemetry: core::marker::PhantomData<T>,
 }
@@ -146,6 +146,7 @@ impl<T: Serialize> TelemetryClient<T> {
                 Retain::NotRetained,
                 &[],
             )
+            .map_err(|e| log::error!("Telemetry publishing error: {:?}", e))
             .ok();
     }
 


jordens (approved):

jordens (approved):

src/net/telemetry.rs
```diff
@@ -20,7 +20,7 @@ use minimq::embedded_nal::IpAddr;
 
 /// The telemetry client for reporting telemetry data over MQTT.
 pub struct TelemetryClient<T: Serialize> {
-    mqtt: minimq::Minimq<NetworkReference, SystemTimer, 512, 1>,
+    mqtt: minimq::Minimq<NetworkReference, SystemTimer, 1024, 1>,
```
ryan-summers, line 23: Hmm... We're definitely starting to get up there in terms of telemetry size. If this continues being an issue in the future, we may want to start breaking telemetry out into separate clients. 1KB buffers in MQTT is going to very quickly start eating up our RAM

ryan-summers:
```
$ cargo size --bin dual-iir --release
   Compiling stabilizer v0.6.0 (C:\Users\rsummers\Documents\repositories\quartiq\stabilizer)
    Finished release [optimized + debuginfo] target(s) in 26.14s
   text    data     bss     dec     hex filename
 242248     112   41096  283456   45340 dual-iir

rsummers@DESKTOP-MEQ1A20 MINGW64 ~/Documents/repositories/quartiq/stabilizer (rs/issue-524/asm-delay)
$ cargo size --bin lockin --release
   Compiling stabilizer v0.6.0 (C:\Users\rsummers\Documents\repositories\quartiq\stabilizer)
    Finished release [optimized + debuginfo] target(s) in 12.44s
   text    data     bss     dec     hex filename
 215064     112   40928  256104   3e868 lockin
```

We have 128KB RAM memory (and we have an additional 512KB of AXISRAM and other RAM utilities. Currently, were only using about 41kB, so we're doing fine on the RAM front. This leaves us with > 80kB for stack usage, which should be plenty.

ryan-summers (commented):
bors r+

PR #614, opened 2022-09-13
Nk/driver output sm

nkrackow:
Draft PR for the Driver output enabling/disabling sequence state machine.
Implements https://github.com/quartiq/ai-artiq/issues/28 with the exception that all checks (temperatue interlock, output voltage) will be done in other tasks. Tested working on Stabilizer. 

I think I would prefer the Realy SM to be aggregated into the Output SM. This would make the logic easier to follow since it's not spread between two different SMs where one SM triggers events in the other. I think we should even change the relay SM logic a bit to enable a better disabling sequence. 

Also there are quite a few rough edges like the settings structure that I will clean up once the design is more settled.
We could even think about hidingparts of the functionality like the IIR parameters for a first, simple, open-loop version. Then we don't expose the possibility for user error during "first light".


src/bin/driver.rs
```diff
@@ -487,6 +505,46 @@ mod app {
...
         c.shared.network.lock(|net| net.direct_stream(target));
+
+        (c.shared.output_state, c.shared.relay_state).lock(|output, relay| {
+            for (i, (new_output_enabled, old_output_enabled)) in new_settings
+                .output_enabled
+                .iter()
+                .zip(old_settings.output_enabled)
+                .enumerate()
+            {
+                if *new_output_enabled && !old_output_enabled {
```
jordens, line 516: You can get rid of having to deref here if you do `for (i, (&new_output_enabled, ...` above.

src/hardware/driver/mod.rs
```diff
@@ -23,3 +22,17 @@ pub enum Channel {
     LowNoise = 0,
     HighPower = 1,
 }
+
+#[derive(Clone, Copy, Debug)]
+pub struct ChannelIndexError;
+
+impl TryFrom<usize> for Channel {
```
jordens, line 29: [TryFromPrimitive](https://docs.rs/num_enum/latest/num_enum/#traits) as elsewhere

src/hardware/driver/output.rs
```diff
@@ -0,0 +1,120 @@
...
+        if self.context().ramp_iir.y_offset >= iir.y_offset {
+            self.process_event(sm::Events::CurrentFinal).unwrap();
+            None
+        } else {
+            self.process_event(sm::Events::CurrentStep).unwrap();
+            Some(Output::RAMP_DELAY)
+        }
+    }
+
+    pub fn iir(&mut self) -> iir::IIR<f32> {
```
jordens, line 113: Better give back a reference here. This will always copy. Instead if you let the user do the copy (`let iir = *state_machine.iir()`, it's a bit more flexible.

src/hardware/driver/output.rs
```diff
@@ -0,0 +1,120 @@
...
+    }
+}
+
+pub mod sm {
+    use super::*;
+    statemachine! {
+        transitions: {
+            *Disabled + Enable = EnableWaitRelay,
+            EnableWaitRelay + RelayDone = RampCurrent,
+            RampCurrent + CurrentStep / increment_current = RampCurrent,
```
jordens, line 31: Can't you `RampCurrent + Enable = RampCurrent` and the same for `EnableWaitRelay + Enable`, and `DisableWaitRelay + Disable`, and `RampCurrent + Disable / reset_iir = DisableWaitRelay`?

jordens (approved):
Very nice and clean.

src/bin/driver.rs
```diff
@@ -487,6 +494,38 @@ mod app {
...
+                    if let Ok(Some(delay)) = output[i].disable().map_err(|e| {
+                        log::error!("Cannot disable output {:?}! {:?}", i, e)
+                    }) {
+                        handle_output_event::spawn_after(
+                            delay.convert(),
+                            i.try_into().unwrap(),
+                        )
+                        .unwrap();
+                    }
+                }
```
jordens, line 526: I think this could be a bit shorter if the API was `output[i].set_enable(new_output_enabled: bool)` and then `if new_output_enabled != old_output_enabled { ... }`. But not critical. Can also remain like this.

src/bin/driver.rs
```diff
@@ -549,13 +588,20 @@ mod app {
         });
     }
 
-    // Task for waiting the relay transition times.
-    #[task(priority = 1, shared=[relay_state])]
-    fn handle_relay(mut c: handle_relay::Context, channel: driver::Channel) {
-        let delay = (c.shared.relay_state)
-            .lock(|state| state[channel as usize].handle_relay());
+    // Task for handling the Powerup/-down sequence
+    #[task(priority = 1, shared=[output_state, settings])]
```
jordens, line 592: What if both outputs are enabled almost simultaneously. The task queue may overflow and the `spawn_after()` will `Err`, `unwrap` and `panic`.

src/bin/driver.rs
```diff
@@ -549,13 +588,20 @@ mod app {
...
+    ) {
+        let iir = c
+            .shared
+            .settings
+            .lock(|settings| settings.iir_ch[channel as usize]);
+        let delay = (c.shared.output_state)
+            .lock(|state| state[channel as usize].handle_output_event(iir));
         if let Some(del) = delay {
-            handle_relay::Monotonic::spawn_after(del.convert(), channel)
+            handle_output_event::Monotonic::spawn_after(del.convert(), channel)
```
jordens, line 604: Slightly more idiomatic:

```rust
handle_output_event(iir)
    .map(|del|
        handle_output_event::Monotonic::spawn_after(del.convert(), channel).unwrap()
);`

File: src/hardware/driver/output.rs [hunk 8]
```diff
@@ -0,0 +1,176 @@
...
+            Ok(Some(Relay::<I2C>::K1_DELAY)) // engage K1 first
+        } else {
+            Ok(None)
+        }
+    }
+
+    /// Handle an event that happens during the enabling/disabling sequence.
+    pub fn handle_output_event(
+        &mut self,
+        iir: iir::IIR<f32>,
```
jordens, line 127: pass this by reference, not value.

src/hardware/driver/output.rs
```diff
@@ -0,0 +1,176 @@
...
+            // Enable sequence
+            *Disabled + Enable / engage_k0 = EnableWaitK0,
+            EnableWaitK0 + RelayDone / disengage_k1 = EnableWaitK1,
+            EnableWaitK1 + RelayDone = RampCurrent,
+            RampCurrent + CurrentStep / increment_current = RampCurrent,
+            RampCurrent + CurrentFinal = Enabled,
+
+            // Abort transitions
+            EnableWaitK0 | EnableWaitK1 | RampCurrent + Disable = Abort,
+            Abort + RelayDone / engage_k1_and_hold_iir = DisableWaitK1,
```
jordens, line 49: What about
`Abort + Disable | Enable`
and
`DisableWaitK1 | DisableWaitK0 + Enable | Disable`
?

src/hardware/driver/output.rs
```diff
@@ -0,0 +1,176 @@
...
+    I2C: WriteRead<Error = E> + Write<Error = E>,
+    E: Debug,
+{
+    /// Start enabling sequence. Returns `Some(relay delay)` or None if we started an Abort
+    /// or an error if SM is in a state where we can't enable.
+    pub fn enable(
+        &mut self,
+    ) -> Result<Option<fugit::MillisDuration<u64>>, sm::Error> {
+        self.process_event(sm::Events::Enable)?;
+        if *self.state() != sm::States::Abort {
```
jordens, line 104: Look at the `process_event()` signature!

```suggestion
        if self.process_event(sm::Events::Enable)? != sm::States::Abort {
```

src/hardware/driver/output.rs
```diff
@@ -0,0 +1,176 @@
...
+}
+
+impl<I2C, E> sm::StateMachine<Output<I2C>>
+where
+    I2C: WriteRead<Error = E> + Write<Error = E>,
+    E: Debug,
+{
+    /// Start enabling sequence. Returns `Some(relay delay)` or None if we started an Abort
+    /// or an error if SM is in a state where we can't enable.
+    pub fn enable(
```
jordens, line 100: Also not critical. But I'm wondering whether you can't put most of that logic in `enable/disable/handle_output_even` can't just go into the state machinery. Using event data and state data.
https://github.com/korken89/smlang-rs/tree/master/examples

jordens (approved):

src/hardware/driver/output.rs
nkrackow: I've been thinking about it for a while but anytime your next state depends on the relation of internal and external data you have to do that outsider an event-action.

src/bin/driver.rs
nkrackow: Yes, I'm aware that currently the outputs cannot be enabled/disabled simultaneously. 
I didn't set the capacity to 2 yet because I wanted things to panic when things go wrong with one channel and we try to spawn another handle.
But I would say just setting the capacity to 2 is fine later.

src/hardware/driver/output.rs
nkrackow: I was thinking that I'll leave those as errors. It's fairly unlikely in practice that another enable/disable comes within 200 ms after the last (aborting/disabling just takes the two relay transition times). And if this occures it's not a problem since I just print the error that the enable/disable was impossible and nothing bad happens.
https://github.com/quartiq/stabilizer/blob/f0b3e916a8a73f4dcd8d5e9a2bff81c29b2824fe/src/bin/driver.rs#L506-L507

jordens (approved):

PR #616, opened 2022-09-19
Nk/driver dac

nkrackow:
This PR adds the Driver main DAC functionality. 
Besides the DAC driver I have integrated the DAC outputs into the usual signal processing flow:
- All processing is now done in units of Volt/Ampere and not in LSBs.
- The Stabilizer DAC still outputs a Voltage equivalent to the Driver current.
- No batch processing, just a single sample per process at a rate of 97.66 kHz.
- The DAC Spi writes are located in a secondary rtic task in order to make them effectively non-blocking for the higher pro tasks. Other priority levels all moved up.

- There are issues with using the SPI at a higher freuency (annotated in the code):
            // - With blocking DAC writes, a clock rate of 25 MHz (next possible step from 12.5 MHz) somehow seems to lead to
            //   longer CPU-blocking SPI writes which in turn means the Stabilizer DAC panics (at 97.6 kHz process rate).
            // - With non-blocking DAC writes I also see spurious SPI Overrun errors after ~30 seconds of program runtime at 
            //   a rate or 25 MHz with an effective DAC update rate of 97.6 kHz (triggered by the DSP process)
            // - 50 MHz immediately leads to FullDuplex error
            //   https://github.com/stm32-rs/stm32h7xx-hal/blob/778ecf7a14d713e4387dbd477a2455d2f71a3424/src/spi.rs#L101-L103
- Using a 12.5 MHz SPI clock the transfers timing usually looks like this (SCLK blue, one of the two CS yellow):
Screenshot from 2022-09-19 15-28-42
- Sometimes (I think when the Driver ADC needs to be handled) I get a slightly delayed output like this:
Screenshot from 2022-09-19 12-19-33
- I am pretty sure this can not lead to overall drift or missed DAC samples since the `dac_spi_write()` task has a capacity of two, so if the second channel was not done before a new DAC sample comes it would panic. I also looked at the [etf-trace](https://github.com/quartiq/etf-trace) outputs and I see a few us between the end of the second `dac_spi_write()` and the start of `process()`
- The `dac_spi_write()` task hast to have a priority higher than `settings_update()` etc. or it will panic. This leaves not too much processing time for telemetry etc. but those tasks also still work fine.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -64,6 +64,7 @@ smlang = "0.5.1"
 enum-iterator = "1.1.3"
 rand_xorshift = "0.3.0"
 rand_core = "0.6.3"
+shared-bus-rtic = "0.2.2"
 
 [dependencies.stm32h7xx-hal]
 features = ["stm32h743v", "rt", "ethernet", "xspi"]
--- a/src/hardware/driver/dac.rs
+++ b/src/hardware/driver/dac.rs
@@ -117,9 +117,9 @@ pub struct DacCode(u32);
 impl DacCode {
     // DAC constants
     const MAX_DAC_WORD: i32 = 1 << 20; // maximum DAC dataword (exclusive)
-    const VREF_DAC: f32 = 5.0; // Difference between
-    const R_OUT_LN: f32 = 100.0; // Low noise side output resistor
-    const R_OUT_HP: f32 = 40.0; // High power side output resistor
+    const VREF_DAC: f32 = 10.0; // Difference between positive and negaitiv reference pin
+    const R_OUT_LN: f32 = 40.0; // Low noise side output resistor
+    const R_OUT_HP: f32 = 0.68; // High power side output resistor
 }
 
 impl TryFrom<(f32, ChannelVariant)> for DacCode {
@@ -133,15 +133,18 @@ impl TryFrom<(f32, ChannelVariant)> for DacCode {
             ChannelVariant::HighPowerAnodeGrounded => (Self::R_OUT_HP, true),
             ChannelVariant::HighPowerCathodeGrounded => (Self::R_OUT_HP, false),
         };
-        let dac_code = (r_out
+        let mut dac_code = (r_out
             * current
             * (DacCode::MAX_DAC_WORD as f32 / DacCode::VREF_DAC))
             as i32;
+        log::info!("dac_code before: {:?}", dac_code);
+
         if is_inverted {
             // Convert to inverted dac output for anode grounded Driver channel variants.
-            // These variants need (VREF_DAC - V_CURR) to produce the current CURR. 
-            dac_code = !dac_code;
+            // These variants need (VREF_DAC - V_CURR) to produce the current CURR.
+            dac_code = DacCode::MAX_DAC_WORD - dac_code + 1;
         }
+        log::info!("dac_code after: {:?}", dac_code);
         if !(0..DacCode::MAX_DAC_WORD).contains(&dac_code) {
             return Err(Error::Bounds);
         };
@@ -183,9 +186,9 @@ where
 
     pub fn set(&mut self, current: f32) -> Result<(), Error> {
         let dac_code = DacCode::try_from((current, self.channel))?;
-        let mut bytes = (DAC_ADDR::DAC_DATA | (dac_code.0 << 4)).to_be_bytes();
+        let bytes = (DAC_ADDR::DAC_DATA | (dac_code.0 << 4)).to_be_bytes();
         self.sync_n.set_low();
-        self.spi.write(&mut bytes).unwrap();
+        self.spi.write(&bytes).unwrap();
         self.sync_n.set_high();
         Ok(())
     }
--- a/src/hardware/driver/dac.rs
+++ b/src/hardware/driver/dac.rs
@@ -4,6 +4,7 @@ use embedded_hal::blocking::spi::{Transfer, Write};
 use stm32h7xx_hal::gpio;
 
 use super::ChannelVariant;
+use embedded_hal::blocking::delay::DelayUs;
 
 // DAC11001 register addresses
 #[allow(unused, non_camel_case_types)]
@@ -183,6 +184,7 @@ where
         spi: SPI,
         sync_n: gpio::ErasedPin<gpio::Output>,
         channel: ChannelVariant,
+        delay: &mut impl DelayUs<u8>,
     ) -> Self {
         let mut dac = Dac {
             spi,
@@ -195,17 +197,19 @@ where
         // set to 10 V plusminus 1.25 V referenece span
         dac.write(DAC_ADDR::CONFIG1, CONFIG1::VREFVAL::SPAN_10V);
         // perform calibration
-        dac.calibrate();
+        dac.calibrate(delay);
 
         dac
     }
 
     /// Perform temperature drift calibration (waits until calibration is done).
-    pub fn calibrate(&mut self) {
+    pub fn calibrate(&mut self, delay: &mut impl DelayUs<u8>) {
         self.write(DAC_ADDR::CONFIG1, CONFIG1::EN_TMP_CAL::ENABLED);
         self.write(DAC_ADDR::TRIGGER, TRIGGER::RCLTMP::RECAL);
         // continuously read recalibration done bit until it is set
-        while (self.read(DAC_ADDR::STATUS) & STATUS::ALM::RECALIBRATED) == 0 {}
+        while (self.read(DAC_ADDR::STATUS, delay) & STATUS::ALM::RECALIBRATED)
+            == 0
+        {}
     }
 
     /// Set the DAC to produce a voltage corresponding to `current`.
@@ -224,12 +228,19 @@ where
     }
 
     /// Perform a read of a DAC register. First writes a word and then reads one.
-    pub fn read(&mut self, addr: DAC_ADDR) -> u32 {
-        let mut buf = [0u8; 8];
-        buf[0] = 1u8 << 7 | (addr as u32 >> 24) as u8;
+    pub fn read(
+        &mut self,
+        addr: DAC_ADDR,
+        delay: &mut impl DelayUs<u8>,
+    ) -> u32 {
+        let mut bytes = (1u32 << 31 | addr as u32).to_be_bytes();
         self.sync_n.set_low();
-        self.spi.transfer(&mut buf).unwrap();
+        self.spi.write(&bytes).unwrap();
         self.sync_n.set_high();
-        u64::from_be_bytes(buf) as u32 // just cut off the upper word where no data was received
+        delay.delay_us(1); // has to be high for at least 100 ns
+        self.sync_n.set_low();
+        self.spi.transfer(&mut bytes).unwrap();
+        self.sync_n.set_high();
+        u32::from_be_bytes(bytes)
     }
 }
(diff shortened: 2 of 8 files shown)


Cargo.toml
```diff
@@ -64,6 +64,7 @@ smlang = "0.5.1"
 enum-iterator = "1.1.3"
 rand_xorshift = "0.3.0"
 rand_core = "0.6.3"
+shared-bus-rtic = "0.2.2"
```
jordens, line 67: Check with @ryan-summers whether this is still the preferred solution.

src/hardware/driver/dac.rs
```diff
@@ -0,0 +1,240 @@
...
+
+#[allow(non_snake_case)]
+pub mod CONFIG1 {
+    pub mod EN_TMP_CAL {
+        pub const DISABLED: u32 = 0 << 23;
+        pub const ENABLED: u32 = 1 << 23;
+    }
+    pub mod TNH_MASK {
+        pub const JUMP14: u32 = 00 << 18;
+        pub const JUMP15: u32 = 01 << 18;
```
jordens, line 28: There are a couple clippy warnings here.
Pretty sure many of your numbers are actually wrong. If you want binary, write `0b01`.

src/hardware/setup.rs
```diff
@@ -1049,11 +1052,53 @@ pub fn setup(
             )),
         ];
 
+        let config = hal::spi::Config::new(hal::spi::Mode {
+            polarity: hal::spi::Polarity::IdleHigh,
```
jordens, line 1056: Wrong mode. CPOL=0 is IdleLow.

image

image

image

src/hardware/setup.rs
```diff
@@ -1049,11 +1052,53 @@ pub fn setup(
...
+            // - 50 MHz immediately leads to FullDuplex error
+            //   https://github.com/stm32-rs/stm32h7xx-hal/blob/778ecf7a14d713e4387dbd477a2455d2f71a3424/src/spi.rs#L101-L103
+            12500.kHz(),
+            ccdr.peripheral.SPI1,
+            &ccdr.clocks,
+        );
+
+        let dac_bus_manager =
+            shared_bus_rtic::new!(dac_spi, hal::spi::Spi<SPI1,Enabled,u8>);
+        let dac0_cs = gpiog.pg10.into_push_pull_output().erase();
```
jordens, line 1081: Setup and deassert (AFAICT they are bot asserted during the first init the way you do it here) the chip selects before the SPI bus!

src/hardware/driver/dac.rs
```diff
@@ -0,0 +1,240 @@
...
+    const R_OUT_LN: f32 = 40.0; // Low noise side output resistor
+    const R_OUT_HP: f32 = 0.68; // High power side output resistor
+}
+
+impl TryFrom<(f32, ChannelVariant)> for DacCode {
+    type Error = Error;
+    /// Convert an f32 representing a current int the corresponding DAC output code for the respective channel.
+    fn try_from(current_channel: (f32, ChannelVariant)) -> Result<Self, Error> {
+        let (current, channel) = current_channel;
+        let (r_out, is_inverted) = match channel {
```
jordens, line 136: Just `-Self::R_OUT_LN` if inverted. And implement that as `fn transimpedance(&self) -> f32` on `ChannelVariant`.

And then your is masking and bounds check should just become check for overflow (recommend do it on paper first).

src/hardware/driver/dac.rs
```diff
@@ -0,0 +1,240 @@
...
+
+    /// Set the DAC to produce a voltage corresponding to `current`.
+    pub fn set(&mut self, current: f32) -> Result<(), Error> {
+        let dac_code = DacCode::try_from((current, self.channel))?;
+        self.write(dac_code.0 << 4, DAC_ADDR::DAC_DATA);
+        Ok(())
+    }
+
+    /// Write 24 bit [data] to the DAC at [addr].
+    pub fn write(&mut self, data: u32, addr: DAC_ADDR) {
```
jordens, line 224:
```suggestion
    pub fn write(&mut self, addr: DAC_ADDR, data: u32) {
```

for consistency.

src/hardware/driver/dac.rs
```diff
@@ -0,0 +1,240 @@
...
+        let mut dac = Dac {
+            spi,
+            sync_n,
+            channel,
+        };
+
+        // reset DAC
+        dac.write(TRIGGER::SRST::RESET, DAC_ADDR::TRIGGER);
+        // set to 10 V plusminus 1.25 V referenece span
+        dac.write(CONFIG1::VREFVAL::SPAN_10V_1_25V, DAC_ADDR::CONFIG1);
```
jordens, line 201: No need to encode the tolerance in the const or the comment. IMO.

src/hardware/driver/dac.rs
```diff
@@ -0,0 +1,240 @@
...
+        self.spi.write(&bytes).unwrap();
+        self.sync_n.set_high();
+    }
+
+    /// Perform a read of a DAC register. First writes a word and then reads one.
+    pub fn read(&mut self, addr: DAC_ADDR) -> u32 {
+        let mut buf = [0u8; 8];
+        buf[0] = 1u8 << 7 | (addr as u32 >> 24) as u8;
+        self.sync_n.set_low();
+        self.spi.transfer(&mut buf).unwrap();
```
jordens, line 236: You probably want to deassert SYNC in between. That then also makes the payload data conversion a bit easier and give more value to you already having shifted the address 24 bit left in the enum.

image

src/hardware/driver/dac.rs
```diff
@@ -0,0 +1,240 @@
...
+        self.write(CONFIG1::EN_TMP_CAL::ENABLED, DAC_ADDR::CONFIG1);
+        self.write(TRIGGER::RCLTMP::RECAL, DAC_ADDR::TRIGGER);
+        // continuously read recalibration done bit until it is set
+        while (self.read(DAC_ADDR::STATUS) & STATUS::ALM::RECALIBRATED) == 0 {}
+    }
+
+    /// Set the DAC to produce a voltage corresponding to `current`.
+    pub fn set(&mut self, current: f32) -> Result<(), Error> {
+        let dac_code = DacCode::try_from((current, self.channel))?;
+        self.write(dac_code.0 << 4, DAC_ADDR::DAC_DATA);
```
jordens, line 219: Avoid the 4 bit left shift and just pretend the dac has 24 bits in `impl From<(channel, current)>`.

jordens (changes requested):

Cargo.toml
nkrackow: I can't use `shared-bus` in rtic resources since it does not impl. send/sync. I think I can also not make it a [local initialized resource](https://rtic.rs/1/book/en/by-example/resources.html#task-local-initialized-resources) since I need to do the whole DAC setup in `setup()`. 
I've also asked @ryan-summers about this.

src/hardware/driver/dac.rs
nkrackow: Ok. I use a 1 us delay for the 100 ns SYNC high timing now. Since we currently only need to read during the DAC initialization we can use the `setup()` `delay` that implements `delay_us()`.

src/hardware/driver/dac.rs
nkrackow: What if I accidentally want to produce a negative current?

nkrackow: Sorry, I cant think of a nice way to check for all the necessary bounds without explicit checks if we are inverted or not right now.. 
The current code is just there to illustrate my problem.

nkrackow:
Thanks for the feedback. 
I just can't figure out the bounds check right now...

And I spent some more time on the higher frequency SPI problems without much success:

- I still don't understand how I can get spurious RX FIFO overruns at 25 MHz if I'm just writing data. I see that the stm32h7hal just [always uses transfer](https://github.com/stm32-rs/stm32h7xx-hal/blob/778ecf7a14d713e4387dbd477a2455d2f71a3424/src/spi.rs#L1268) but that should still not lead to overflows.

- And The DuplexFailed Errors at 50 MHz seem to imply the SPI clock is re-sampled at the pin that it is expelled which fails at 50 MHz and therefore the RX FIFO (that I usually don't need) is empty.

src/hardware/driver/dac.rs
jordens: How did you resolve this?

src/hardware/driver/dac.rs
```diff
@@ -0,0 +1,245 @@
...
+            as i32;
+
+        if !(-DacCode::MAX_DAC_WORD..DacCode::MAX_DAC_WORD).contains(&dac_code)
+        {
+            return Err(Error::Bounds);
+        };
+
+        Ok(Self((dac_code & 0xfffff) as u32)) // translate to valid DAC code
+    }
+}
```
jordens, line 151: Something like this:

```suggestion
    fn try_from(current_channel: (f32, ChannelVariant)) -> Result<Self, Error> {
        let (current, channel) = current_channel;
        let scale = channel.transimpedance()
            * (DacCode::MAX_DAC_WORD as f32 / DacCode::VREF_DAC);
        let mut code = (current * scale) as i32;
        if scale < 0. {
            // Flip the sign and overflow bits
            code ^= !(DacCode::MAX_DAC_WORD - 1);
        }
        if !(0..DacCode::MAX_DAC_WORD).contains(&code) {
            Err(Error::Bounds)
        } else {
            Ok(Self((code & (DacCode::MAX_DAC_WORD - 1)) as u32))
        }
    }
}
```

src/hardware/driver/dac.rs
jordens: https://github.com/quartiq/stabilizer/pull/616#pullrequestreview-1113981259

jordens:
IIRC you need something like `f_SPI_peripheral_ker >= 3x f_SCLK`. Maybe play with the clock routing and check that it is like this for the other SPI buses. Do they also use less than 50 MHz?

src/hardware/driver/dac.rs
nkrackow: Oh sry I accidentally clicked resolved..

src/hardware/driver/dac.rs
jordens: But do review this. I may have missed something.

src/hardware/driver/dac.rs
nkrackow: Ah, I was missing line 150 and thought about doing it without an if.

src/hardware/driver/dac.rs
nkrackow: I checked it for upper/lower bound in both cases. The bounds are as expected (and annoying since I immediately panic at 0.0 A for the inverted channels).

jordens (approved):

nkrackow:
Since the changes I am now making to the Outputs aren't really related to the DAC driver, I'll merge this and make a new PR for that. I think the state right now is reasonable. I'll postpone the higher freq SPI problem for now.

src/hardware/driver/mod.rs
```diff
@@ -22,3 +28,25 @@ pub enum Channel {
...
     HighPower = 1,
 }
+
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+#[repr(usize)]
+pub enum ChannelVariant {
+    LowNoiseAnodeGrounded,
+    LowNoiseCathodeGrounded,
+    HighPowerAnodeGrounded,
+    HighPowerCathodeGrounded,
```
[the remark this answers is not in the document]
nkrackow: makes sense.

jordens (approved):

PR #618, opened 2022-09-22
Nk/driver output refactor

nkrackow:
This PR reworks the output such that only the low noise channel is processed and updated at the sample rate.
The high power channel is only updated in `settings_update()` or during the output enable sequence.

It also adapts the miniconf settings for Driver.

I'll rework parts of the Telemetry and Streaming in the next PR, although I think I might have to cut some corners on the streaming if that takes me too long.

### DIFF
--- a/src/bin/driver.rs
+++ b/src/bin/driver.rs
@@ -94,13 +94,13 @@ pub struct Settings {
     /// Low noise channel output settings
     ///
     /// # Value
-    /// See [LowNoiseSettings]
+    /// See [driver::output::LowNoiseSettings]
     pub low_noise: driver::output::LowNoiseSettings,
 
     /// High power channel output settings
     ///
     /// # Value
-    /// See [HighPowerSettings]
+    /// See [driver::output::HighPowerSettings]
     pub high_power: driver::output::HighPowerSettings,
 }
 
--- a/src/bin/driver.rs
+++ b/src/bin/driver.rs
@@ -263,35 +263,31 @@ mod app {
                         && settings.output.low_noise.allow_hold);
 
                 (adc0, adc1, dac0, dac1).lock(|adc0, adc1, dac0, dac1| {
-                    let adc_samples = [adc0, adc1];
-                    let dac_samples = [dac0, dac1];
-
                     // Preserve instruction and data ordering w.r.t. DMA flag access.
                     fence(Ordering::SeqCst);
 
                     // Perform processing for  ADC channel 0 --> Driver channel 0 (LowNoise)
-                    adc_samples[0]
-                        .iter()
-                        .zip(dac_samples[0].iter_mut())
-                        .map(|(ai, di)| {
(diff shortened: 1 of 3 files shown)


src/hardware/driver/output.rs
```diff
@@ -6,10 +6,90 @@ use core::fmt::Debug;
...
 use mcp230xx::{Mcp23008, Mcp230xx};
+use miniconf::Miniconf;
 use smlang::statemachine;
 
 use super::{relay::Relay, Channel};
 
+const LN_MAX_I_DEFAULT: f32 = 0.2; // default maximum current for the low noise channel in ampere
+
+#[derive(Clone, Copy, Debug, Miniconf, Default)]
+pub struct OutputSettings {
```
jordens, line 17: I'd remove this and just LN and HP settings into the main Settings.

src/hardware/driver/dac.rs
```diff
@@ -110,7 +110,7 @@ pub mod CONFIG2 {
 /// DAC value out of bounds error.
 #[derive(Debug)]
 pub enum Error {
-    Bounds,
+    Bounds(f32, i32),
```
jordens, line 113: Is that for debugging or do you want to generally do it like that?

src/bin/driver.rs
```diff
@@ -448,11 +387,21 @@ mod app {
...
+                [
+                    &old_settings.output.low_noise.output_enabled,
+                    &old_settings.output.high_power.output_enabled,
+                ],
+            );
+            for (i, (&new_output_enabled, &old_output_enabled)) in
+                new_output_enabled
+                    .iter()
+                    .zip(old_output_enabled)
+                    .enumerate()
```
jordens, line 404: The two arrays can be in-line, `for .. in [new_settings..., new_settings....].iter().zip([old..., old...]).enumerate() {`

src/bin/driver.rs
```diff
@@ -309,46 +258,36 @@ mod app {
...
-                                write_dac_spi::spawn(
-                                    channel.try_into().unwrap(),
-                                    y,
-                                )
-                                .unwrap();
-                            })
-                            .last();
-                    }
+                    // Perform processing for  ADC channel 0 --> Driver channel 0 (LowNoise)
+                    let x = f32::from(adc0[0]); // get adc sample in volt
```
jordens, line 270: Probably goo didea to apply the AFE gain here (you have the settings) to avoid confusion.

jordens (changes requested):
Minor changes.

src/hardware/driver/dac.rs
nkrackow: I thought there's generally no downside to returning this in the error. Ofc the error should usually not happen. But with the current sitiation that 0.0 A is a bounds error for the current sink channel it might be useful later.

jordens (approved):

PR #620, opened 2022-09-23
Nk/driver tele

nkrackow:
Small PR that implements the Driver telemetry.

### DIFF
--- a/src/bin/driver.rs
+++ b/src/bin/driver.rs
@@ -18,10 +18,11 @@ use mutex_trait::prelude::*;
 
 use idsp::iir;
 
+use serde::Serialize;
 use stabilizer::{
     hardware::{
         self,
-        adc::{Adc0Input, Adc1Input, AdcCode},
+        adc::{Adc0Input, Adc1Input},
         afe::Gain,
         dac::{Dac0Output, Dac1Output, DacCode},
         design_parameters, driver,
@@ -37,7 +38,6 @@ use stabilizer::{
     net::{
         data_stream::{FrameGenerator, StreamFormat, StreamTarget},
         miniconf::Miniconf,
-        telemetry::{Telemetry, TelemetryBuffer},
         NetworkState, NetworkUsers,
     },
 };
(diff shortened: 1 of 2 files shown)


src/bin/driver.rs
```diff
@@ -119,6 +119,28 @@ impl Default for Settings {
     }
 }
 
+#[derive(Serialize, Copy, Clone, Default, Debug)]
+pub struct Monitor {
+    output_current: [f32; 2],
```
jordens, line 124: Please check the names. Do you need `output` in all these (and elsewhere)? If there is `output`, why is there no `input` but `feedback`? `feedback` is also probably a bad name as that input it does not necessarily need to be for feedback at all.

src/bin/driver.rs
```diff
@@ -119,6 +119,28 @@ impl Default for Settings {
...
+    output_voltage: [f32; 2],
+    cpu_temp: f32,
+    overtemp: bool,
+}
+
+#[derive(Serialize, Copy, Clone, Default, Debug)]
+pub struct LowNoise {
```
jordens, line 131: `LowNoiseControl{adc0, current}`?

Try sticking to the names on the schematic/panel.
If this is a current sink, is `output_current` really a useful name? What about just `current`? It's clear from the context that this is not the monitor data.

jordens (changes requested):
Just name bikeshedding.

src/bin/driver.rs
nkrackow: Ok, I mean yes there ate inputs on stabilizer. Those might not be exposed in telemetry but I thought just `voltage` might be confusing for someone who has not used Drive+Stabilizer before.

src/bin/driver.rs
jordens: What about the struct name?

src/bin/driver.rs
jordens: Then call the struct `DriverMonitor` to make that clear. And use two of them, not one with two arrays.

PR #622, opened 2022-10-12
Nk/laser interlock

nkrackow:
This PR adds the laser interlock. The laser interlock is tripped (pin goes low) when the interlock voltage/current is surpassed on a channel or when the thermostat interlock trips.

### DIFF
--- a/src/bin/driver.rs
+++ b/src/bin/driver.rs
@@ -133,6 +133,7 @@ pub struct Monitor {
     voltage: [f32; 2],
     cpu_temp: f32,
     header_temp: f32,
+    laser_interlock: bool,
 }
 
 #[derive(Serialize, Copy, Clone, Default, Debug)]
@@ -378,7 +379,7 @@ mod app {
         c.shared.interlock_handle.lock(|handle| {
             if let Some(action) = new_settings.thermostat_interlock.action(
                 path.as_ref()
-                    .map(|path| path.as_str().trim_start_matches("interlock/")),
+                    .map(|path| path.as_str().trim_start_matches("thermostat_interlock/")),
                 handle.is_some(),
                 &old_settings.thermostat_interlock,
             ) {
@@ -466,7 +467,7 @@ mod app {
         }
     }
 
-    #[task(priority = 1, shared=[network, settings, telemetry, internal_adc], local=[cpu_temp_sensor, lm75])]
+    #[task(priority = 1, shared=[network, settings, telemetry, internal_adc, laser_interlock_pin], local=[cpu_temp_sensor, lm75])]
     fn telemetry(mut c: telemetry::Context) {
         let mut telemetry: Telemetry =
             c.shared.telemetry.lock(|telemetry| *telemetry);
@@ -489,6 +490,10 @@ mod app {
                     ],
                 )
             });
+        telemetry.monitor.laser_interlock = c
+            .shared
+            .laser_interlock_pin
+            .lock(|pin| pin.get_state() == hal::gpio::PinState::High);
         let telemetry_period = c
             .shared
             .settings
@@ -573,10 +578,11 @@ mod app {
         c.local.driver_dac[channel as usize].set(current).unwrap()
     }
 
-    #[task(priority = 1, shared=[interlock_handle])]
+    #[task(priority = 1, shared=[interlock_handle, laser_interlock_pin])]
     fn trip_interlock(mut c: trip_interlock::Context) {
         c.shared.interlock_handle.lock(|handle| *handle = None);
         log::error!("interlock tripped");
+        c.shared.laser_interlock_pin.lock(|pin| pin.set_low());
     }
 
     #[task(binds = ETH, priority = 1)]
--- a/src/bin/driver.rs
+++ b/src/bin/driver.rs
@@ -28,7 +28,7 @@ use stabilizer::{
         design_parameters, driver,
         driver::{
             interlock::{Action, Interlock},
-            output, Channel,
+            output,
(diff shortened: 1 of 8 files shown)


src/hardware/design_parameters.rs
```diff
@@ -53,3 +54,7 @@ pub const DDS_SYNC_CLK_DIV: u8 = 4;
 pub const MAX_SAMPLE_BUFFER_SIZE: usize = 32;
 
 pub type SampleBuffer = [u16; MAX_SAMPLE_BUFFER_SIZE];
+
+/// Period for checking the Driver output interlock voltage/current levels.
+pub const DRIVER_MONITOR_PERIOD: Duration<u64, 1, 10000> =
```
jordens, line 59: Maybe better as a miniconf setting?

src/hardware/driver/internal_adc.rs
```diff
@@ -79,4 +80,25 @@ impl InternalAdc {
         const OFFSET: f32 = 0.0; // Current sense offset         ToDo
         (ratio + OFFSET) * SCALE
     }
+
+    pub fn read_all(&mut self) -> [(f32, DriverMonitorChannel); 4] {
```
jordens, line 84: Ah. This is not a convenient API. The return array looses all info about the values. And you have to reconstruct it manually from the indices. Better use `read_output_current/voltage()` directly.

src/hardware/setup.rs
```diff
@@ -1097,12 +1097,16 @@ pub fn setup(
             ),
         ];
 
+        let mut laser_interlock_pin = pounder_pgood.into_push_pull_output();
+        laser_interlock_pin.set_low();
```
jordens, line 1101: Comment what this means.

src/hardware/driver/interlock.rs
```diff
@@ -25,7 +25,7 @@ pub struct Interlock {
     ///
     /// # Value
     /// "true" or "false"
-    armed: bool,
+    pub armed: bool,
```
jordens, line 28: Where is this needed?

src/bin/driver.rs
```diff
@@ -549,10 +564,40 @@ mod app {
...
+        let limits = c.shared.settings.lock(|settings| {
+            [
+                settings.low_noise.interlock_current,
+                settings.high_power.interlock_current,
+                settings.low_noise.interlock_voltage,
+                settings.high_power.interlock_voltage,
+            ]
+        });
+        let measurements = c.local.internal_adc.read_all();
+        for (limit, measurement) in limits.iter().zip(measurements.iter()) {
```
jordens, line 585: Not critical, but a direct API that reads one value would probably be not more code but more readable. See `internal_adc`

src/bin/driver.rs
```diff
@@ -549,10 +564,40 @@ mod app {
...
+            if (measurement.0 > *limit)
+                && c.shared
+                    .laser_interlock_pin
```
jordens, line 588: This is wrong. You are reading the pin state in each iteration. And suppressing the print on multiple limit faults.
Read it once.

src/bin/driver.rs
```diff
@@ -549,10 +564,40 @@ mod app {
...
+                    .lock(|pin| pin.get_state() == PinState::High)
+            {
+                c.shared.laser_interlock_pin.lock(|pin| pin.set_low());
+                log::info!("Driver output Overvoltage/current condition! {:?} Laser interlock tripped.", measurement.1);
+            }
+        }
+        c.shared.telemetry.lock(|tele| {
```
jordens, line 595: This would also become more readable once the `internal_adc` api matches the `settings.x` api and the `telemetry.monitor` structure.
I'd make `internal_adc` have two functions returning 2-arrays. Then the telemetry becomes easy.
And the iteration as well. Maybe use `flatten()` if it helps.

Also why are the settings called `interlock` and the things here called `limit`?

src/bin/driver.rs
```diff
@@ -456,19 +480,10 @@ mod app {
...
-                    [
-                        adc.read_output_voltage(Channel::LowNoise),
-                        adc.read_output_voltage(Channel::HighPower),
-                    ],
-                )
-            });
+        telemetry.monitor.laser_interlock = c
+            .shared
+            .laser_interlock_pin
+            .lock(|pin| pin.get_state() == hal::gpio::PinState::High);
```
jordens, line 486: You need to make sure it's clear what high and low means. Which is asserted? Either comment it everywhere or make an API that makes it clear.

jordens (changes requested):

src/hardware/design_parameters.rs
nkrackow: Imo this should not be exposed to a user. I think this should be optimized to run as fast as possible at a later stage.

jordens: But then it can remain a constant in the app, right? It's not shared anywhere.

src/hardware/driver/mod.rs
```diff
@@ -51,3 +53,21 @@ impl ChannelVariant {
         }
     }
 }
+
+#[derive(
+    Clone, Copy, Debug, Default, PartialEq, Eq, Serialize, Deserialize,
+)]
+pub enum LaserInterlock {
+    Asserted,
```
jordens, line 61: I'd hesitate calling an interlock that is in it's "not tripped" state "asserted".
Propose `NotTripped` to resolve ambiguities.

jordens (changes requested):
One renaming. Otherwise LGTM

PR #623, opened 2022-10-20
Nk/fix settings update

nkrackow:
This PR fixes two closely related bugs (that I put in):

* If a settings topic is longer than 64 chars but shorter that 128 it would panic on the `push_string().unwrap()`.
* Since the closure potentially gets called multiple times insider `handled_update()` (eg. when many retained settings come in at once) `.push_string()` continues to append every setting to `settings_path`.


nkrackow:
Tested with dual-iir and lockin.

jordens (approved):

PR #626, opened 2022-10-21
Nk/interlock tele

nkrackow:
This PR
* adds the laser interlock state to the telemetry. 
* refactors the laser interlock to store information about the tripping reason.
* refactors the logic of the thermostat interlock such that it can be reset by resetting the laser interlock.

### DIFF
--- a/src/bin/driver.rs
+++ b/src/bin/driver.rs
@@ -13,14 +13,13 @@ use core::mem::MaybeUninit;
 use core::sync::atomic::{fence, Ordering};
 
 use fugit::{Duration, ExtU64, TimerInstantU64};
-use hal::gpio::PinState;
 use heapless::String;
 use mutex_trait::prelude::*;
 
 use idsp::iir;
 
 use serde::Serialize;
-use stabilizer::hardware::driver::LaserInterlock;
+use stabilizer::hardware::driver::{LaserInterlock, LaserInterlockState};
 use stabilizer::{
     hardware::{
         self,
@@ -135,7 +134,7 @@ pub struct Monitor {
     voltage: [f32; 2],
     cpu_temp: f32,
     header_temp: f32,
-    laser_interlock: LaserInterlock,
+    laser_interlock: LaserInterlockState,
 }
 
 #[derive(Serialize, Copy, Clone, Default, Debug)]
@@ -172,7 +171,7 @@ mod app {
         header_adc_data: [u16; 8],
         output_state: [output::sm::StateMachine<output::Output<I2c1Proxy>>; 2],
         interlock_handle: Option<trip_interlock::SpawnHandle>,
-        laser_interlock_pin: hal::gpio::Pin<'B', 13, hal::gpio::Output>,
+        laser_interlock: LaserInterlock,
     }
 
     #[local]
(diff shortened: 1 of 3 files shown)


nkrackow:
The clippy lints at the end are due to commented out parts that don't work without Driver HW right now.

src/hardware/driver/mod.rs
```diff
@@ -57,17 +58,37 @@ impl ChannelVariant {
...
+
+    pub fn set(&mut self, state: Option<LaserInterlockTripped>) {
         match state {
-            PinState::High => Self::NotTripped,
-            PinState::Low => Self::Tripped,
+            Some(LaserInterlockTripped::Reset) => self.pin.set_low(),
+            Some(LaserInterlockTripped::Thermostat) => self.pin.set_low(),
+            Some(LaserInterlockTripped::Overcurrent(_)) => self.pin.set_low(),
+            Some(LaserInterlockTripped::Overvoltage(_)) => self.pin.set_low(),
+            None => self.pin.set_high(),
```
jordens, line 90:
```suggestion
            self.pin.set_state(state.is_none().into());
```

src/bin/driver.rs
```diff
@@ -417,6 +415,34 @@ mod app {
...
         });
 
+        // Only set interlock pin high if all channels disabled and false -> true transition.
+        if !old_settings.reset_laser_interlock
+            && new_settings.reset_laser_interlock
+        {
+            if c.shared.output_state.lock(|output| {
+                output[driver::Channel::HighPower as usize].is_enabled()
+            }) {
+                log::error!("Cannot reset laser interlock while HighPower channel is enabled.")
```
jordens, line 425: Always good idea to tell the user what to do here.

src/hardware/driver/mod.rs
```diff
@@ -57,17 +58,37 @@ impl ChannelVariant {
...
+pub enum LaserInterlockTripped {
     #[default]
-    Tripped,
+    Reset, // Tripped after device reset
+    Thermostat,
+    Overcurrent(Channel),
+    Overvoltage(Channel),
+}
+pub struct LaserInterlock {
+    pub state: Option<LaserInterlockTripped>,
```
jordens, line 69: Don't make these `pub`!
Better do `pub fn state(&self) -> ...`.

jordens, line 69: And better call this `reason`.
Call the `enum` `Reason`.

src/bin/driver.rs
```diff
@@ -600,20 +616,21 @@ mod app {
...
         for (i, (interlock_current, read)) in interlock_current
             .iter()
             .zip((current_reads).iter())
             .enumerate()
         {
             if (read > interlock_current) & interlock_asserted {
-                c.shared.laser_interlock_pin.lock(|pin| pin.set_low());
+                let ch = Channel::try_from(i).unwrap();
+                c.shared.laser_interlock.lock(|ilock| {
+                    ilock.set(Some(LaserInterlockTripped::Overcurrent(ch)))
```
jordens, line 629: `interlock_asserted`  becomes inconsistent with the pin and and `ilock.state` becomes inconsistent with the state that caused the interlock to fire!

jordens (changes requested):

src/hardware/driver/mod.rs
nkrackow: nice..

src/bin/driver.rs
nkrackow: Ok, but does that matter if `interlock_asserted` is dropped anyways at the end of the task? If we have an overvoltage condition as well that would trp the interlock a second time, but that isn't a problem either right?

PR #629, opened 2022-10-27
Nk/driver alarm

nkrackow:
This PR changes the thermostat interlock to "alarm". Functionality is the same with the exception that the polarity of the critical topic is flipped (alarm=true is bad while interlocked=true was good)

### DIFF
--- a/src/bin/driver.rs
+++ b/src/bin/driver.rs
@@ -28,7 +28,7 @@ use stabilizer::{
         dac::{Dac0Output, Dac1Output, DacCode},
         design_parameters, driver,
         driver::{
-            interlock::{Action, Interlock},
+            alarm::{Action, Alarm},
             output, Channel,
         },
         hal,
@@ -83,12 +83,12 @@ pub struct Settings {
     /// See [StreamTarget#miniconf]
     stream_target: StreamTarget,
 
-    /// Thermostat interlock settings.
+    /// Thermostat alarm settings.
     ///
     ///
     /// # Value
-    /// [Interlock]
-    thermostat_interlock: Interlock,
+    /// [Alarm]
+    alarm: Alarm,
 
     /// Laser interlock reset.
     /// A false -> true transition will reset a tripped laser interlock
@@ -120,7 +120,7 @@ impl Default for Settings {
             // The default telemetry period in seconds.
             telemetry_period: 1,
             stream_target: StreamTarget::default(),
-            thermostat_interlock: Interlock::default(),
+            alarm: Alarm::default(),
             reset_laser_interlock: false,
             low_noise: driver::output::LowNoiseSettings::default(),
             high_power: driver::output::HighPowerSettings::default(),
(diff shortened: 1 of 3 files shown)


src/hardware/driver/alarm.rs
```diff
@@ -1,23 +1,23 @@
...
+///! The alarm is implemented via MQTT. See [Alarm] for details about the MQTT interface.
 use fugit::ExtU64;
 use miniconf::Miniconf;
 
 #[derive(Clone, Copy, Debug, Miniconf)]
-pub struct Interlock {
-    /// "Interlocked" topic. Publishing "true" onto this topic renews the interlock timeout.
-    /// Publishing "false" or failing to publish `true` for `timeout` trips the interlock.
+pub struct Alarm {
+    /// "Alarm" topic. Publishing "false" onto this topic renews the alarm timeout.
```
jordens, line 10:
```suggestion
    /// "Alarm" topic. Publish "false" onto this topic to indicate valid operating conditions. This renews the alarm timeout.
```

src/hardware/driver/alarm.rs
```diff
@@ -1,23 +1,23 @@
...
+    /// Publishing "true" or failing to publish "false" for `timeout` trips the alarm.
```
jordens, line 11: And then does what?
What do I do to clear?

src/hardware/driver/alarm.rs
```diff
@@ -49,29 +49,30 @@ impl Default for Interlock {
...
         match path {
-            Some("interlocked") => {
-                if self.interlocked && handle_is_some {
+            Some("alarm") => {
+                if !self.alarm && handle_is_some {
                     Some(Action::Reschedule(self.timeout.millis()))
-                } else if !self.interlocked && handle_is_some {
-                    Some(Action::Reschedule(0.millis()))
+                } else if self.alarm && handle_is_some {
+                    Some(Action::Trip(100.millis()))
```
jordens, line 67: Why 100?

src/bin/driver.rs
```diff
@@ -83,12 +83,12 @@ pub struct Settings {
     /// See [StreamTarget#miniconf]
     stream_target: StreamTarget,
 
-    /// Thermostat interlock settings.
+    /// Thermostat alarm settings.
```
jordens, line 86: I would not mention thermostat so prominently here and elsewhere. The feature is not limited to thermostat.
Do mention it as an example.

src/hardware/driver/mod.rs
```diff
@@ -61,7 +61,8 @@ impl ChannelVariant {
 pub enum Reason {
     #[default]
     Reset, // Tripped after device reset
-    Thermostat,
+    ThermostatLimits,
```
jordens, line 64: This isn't specific to `Thermostat` or `Limits`. Just call the two variants `Alarm` and `AlarmTimeout`.

jordens (changes requested):

src/hardware/driver/alarm.rs
nkrackow: oops, that was from debugging

jordens (approved):

PR #631, opened 2022-11-02
Nk/selftest

nkrackow:


src/hardware/driver/output.rs
```diff
@@ -231,19 +209,66 @@ where
...
+        if let Some(tests) = tests {
+            for ((range, &read), &reason) in
+                tests.0.iter().zip(reads.iter()).zip(tests.1.iter())
+            {
+                if !range.contains(&read) {
+                    interlock.set(Some(Reason::Selftest(Selftest {
+                        reason,
+                        read,
+                        channel,
+                    })));
```
jordens, line 260: If you want to reduce function size (and you should), this and a match for the range constants could have probably gone into the SelfTest enum impl. But also fine as it is.

jordens (approved):
All good.
One hint.
And you have accumulated a bunch of lints you may want to look at.

nkrackow:
> All good. One hint. And you have accumulated a bunch of lints you may want to look at.

Yeah they are from me just commenting out things so that the code works without Driver or the Headboard. I'll make a compile time option for the Headboard.

PR #632, opened 2022-11-03
new miniconf, minimq

jordens:


src/hardware/signal_generator.rs
```diff
@@ -20,7 +20,7 @@ pub enum Signal {
 /// Where `<signal>` may be any of [Signal] variants, `frequency` specifies the signal frequency
 /// in Hertz, `symmetry` specifies the normalized signal symmetry which ranges from 0 - 1.0, and
 /// `amplitude` specifies the signal amplitude in Volts.
-#[derive(Copy, Clone, Debug, Miniconf, Deserialize)]
+#[derive(Copy, Clone, Debug, Miniconf, Serialize, Deserialize)]
```
ryan-summers, line 23: I don't think we should need `Serialize` or `Deserialize` on this structure anymore, since it is never configured as a single value, correct?

jordens: Ack. But it doesn't hurt.

ryan-summers: Agreed that it doesn't hurt. Just trying to validate my understanding of the current state of `miniconf` :)

jordens: Yes. One early annoyance I hit is that I can't make a `Miniconf` struct `Serialize` or `Deserialize` if it contains a `miniconf::Array`. Not that it's needed urgently right now, but it does show the impact of those newtypes.

Cargo.toml
```diff
@@ -45,14 +45,13 @@ embedded-hal = "0.2.7"
 nb = "1.0.0"
 num_enum = { version = "0.5.7", default-features = false }
 paste = "1"
-idsp = "0.8"
+idsp = "0.9"
 ad9959 = { path = "ad9959", version = "0.2.0" }
-miniconf = "0.5"
+miniconf = { git = "https://github.com/quartiq/miniconf.git", branch = "rj/bump-serde-json-core" }
```
ryan-summers, line 50:
```suggestion
miniconf = "0.6.0"
```

Cargo.toml
```diff
@@ -66,7 +65,9 @@ rand_core = "0.6.4"
 
 [dependencies.stm32h7xx-hal]
 features = ["stm32h743v", "rt", "ethernet", "xspi"]
-version = "0.13.0"
+# version = "0.13.0"
+git = "https://github.com/stm32-rs/stm32h7xx-hal.git"
+branch = "master"
```
ryan-summers, line 70: Can we hold off on this change for another PR? This doesn't really make sense for miniconf/minimq updates

ryan-summers (changes requested):
I think we need to bump MSRV to 1.65

book/src/setup.md
```diff
@@ -68,22 +68,22 @@ docker run -p 1883:1883 --name mosquitto -v ${pwd}/mosquitto.conf:/mosquitto/con
 ## Building
 
 1. Get and install [rustup](https://rustup.rs/) and use it to install a current stable Rust toolchain.
-    The minimum supported Rust version (MSRV) is 1.62.0 (2021 edition).
-1. Install target support
+    Stabilizer tracks stable Tust. The minimum supported Rust version (MSRV) is specified in the manifest (`Cargo.toml`).
```
ryan-summers, line 71:
```suggestion
    Stabilizer tracks stable Rust. The minimum supported Rust version (MSRV) is specified in the manifest (`Cargo.toml`).
```

ryan-summers (approved):
bors r+

Not sure what's up with deadlinks

jordens:
I suspect that's a miniconf issue.

ryan-summers:
Hmm, indeed it looks like Miniconf is generating a broken link in docs because of `DerefMut` pulling in `std` docs: https://docs.rs/miniconf/0.6.0/miniconf/struct.Option.html#examples-2

ryan-summers:
Perhaps we should make the link-check informational-only and not required as part of bors. I'm not sure if there's anything we can actually do to fix this outside of removing `DerefMut` from miniconf

jordens:
But I don't see what we can do about that. More of a bug in rust as I'd expect the `core` docs should build without without `std`.

PR #640, opened 2022-11-07
compile time features for the cases without Driver and without the headboard

nkrackow:
I'm not sure where exactly to put this best. Having the state machine transitions different depending on compile time features is not possible, right?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -73,6 +73,8 @@ version = "0.12.2"
 [features]
 nightly = [ ]
 pounder_v1_0 = [ ]
+no_driver = [ ]
+no_headboard = [ ]
 
 [profile.dev]
 codegen-units = 1
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -71,10 +71,10 @@ features = ["stm32h743v", "rt", "ethernet", "xspi"]
 version = "0.12.2"
 
 [features]
+default = ["ai_artiq_laser_module"]
 nightly = [ ]
 pounder_v1_0 = [ ]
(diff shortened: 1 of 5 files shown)


jordens (changes requested):
what's breadboard? what's it's real name?
You can also `#[allow(unused something something)]` to suppress the lint. Whatever feels easier.
Negative `no_X` features are clearly a bad idea. Especially if you then negate them all again! Instead make them positive and change `default-features` if you are really worried about having to pass them.
And consider rather having one feature that allows you to work in your current mockup environment, not two.

jordens:
Then why not call it "laser module" as in the design? It's certainly not a breadboard.

Are you sure the use case without the laser module is well defined? (I.e. written down). Then you can judge whether it's worth it carrying it through as a feature, or whether run-time detection is much easier (`Option<Mcp...>`), or whether it's a hack done by editing the code.

nkrackow:
I mean the likelyhood of driving _a_ laser module with driver is fairly high. It might just not be the ai-artiq module with its specific carrier board. I've called it "ai-artiq-laser-module" now.
I would argue that this is best implemented as a feature, will certainly happen at FBH that they test Driver on another laser module.
The case without Driver is a hack for development, so I've reverted back to just commenting out the one line in this case.

nkrackow:
Do you think run-time detection for the laser module is better? I guess it could be that I wanted to connect the headboard but just failed to do so correctly.

jordens:
It's only better if the features stuff becomes annoying.

jordens (approved):

PR #641, opened 2022-11-07
Removing doc requirements from bors

ryan-summers:
There's an issue with Rust where docs included in example code are not compiled in to `miniconf`. We can't fix this issue without fixing some of the core libs. As such, ignore doc requirements from Bors for now.


jordens (approved):

ryan-summers:
bors r=jordens

Issue #647, opened 2022-11-11
Peripheral clock seems to be off.


nkrackow:
While migrating to the HAL V13.1 I expected the program to panic because the HAL now [checks that the internal ADC core clock is not above 80 MHz in out case](https://github.com/stm32-rs/stm32h7xx-hal/blob/master/src/adc.rs#L578-L583).
[Point in the datasheet](https://www.st.com/resource/en/reference_manual/dm00314099-stm32h742-stm32h743-753-and-stm32h750-value-line-advanced-arm-based-32-bit-mcus-stmicroelectronics.pdf#page=353).

https://github.com/quartiq/stabilizer/blob/3562cdc377093fbfabdef91c47467ef83d024654/src/hardware/setup.rs#L278
Using a debug print in the HAL adc setup code I determined that the ADC core and therefore `per_ck` is clocked at 64 MHz and not at 100 MHz as the line above suggests.
Is this plausible with other peripherals or am I missing something?

jordens:
Maybe its just integer divider and rounding.

nkrackow:
Ok, so not a problem? I thought because it's some `TIMER_FREQUENCY` it might matter that it's actually that value.

jordens:
Not sure. The timer tick is (also) set at 
https://github.com/quartiq/stabilizer/blob/3562cdc377093fbfabdef91c47467ef83d024654/src/hardware/setup.rs#L324
and other places. I don't know off hand what clock the timers are clocked by.

Should definitely be checked and either fixed or commented.
That said I was also not entirely convinced I understood the internal ADC clocking frequencies and constraints.

PR #650, opened 2022-11-17
[NFC] clarify per_ck

nkrackow:
closes https://github.com/quartiq/stabilizer/issues/647

### DIFF
--- a/src/hardware/setup.rs
+++ b/src/hardware/setup.rs
@@ -275,7 +275,7 @@ pub fn setup(
         .use_hse(8.MHz())
         .sysclk(design_parameters::SYSCLK.convert())
         .hclk(200.MHz())
-        .per_ck(64.MHz()) // 64 MHz HSI clk selected per default (this setting changes nothing)
+        .per_ck(64.MHz()) // fixed frequency HSI, only used for internal ADC. This is not the "peripheral" clock for timers and others.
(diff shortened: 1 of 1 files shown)


src/hardware/setup.rs
```diff
@@ -275,7 +275,7 @@ pub fn setup(
         .use_hse(8.MHz())
         .sysclk(design_parameters::SYSCLK.convert())
         .hclk(200.MHz())
-        .per_ck(design_parameters::TIMER_FREQUENCY.convert())
+        .per_ck(64.MHz()) // 64 MHz HSI clk selected per default (this setting changes nothing)
```
jordens, line 278:
```suggestion
        .per_ck(64.MHz()) // HSI, only used for internal ADC. This is not the "peripheral" clock for timers and others.
```

jordens (changes requested):

src/hardware/setup.rs
ryan-summers: Technically speaking, the `per_ck` _can_ be the peripheral "kernel" clock for some peripherals (i.e. SPI), but we don't use it as that: 
image

jordens: Yes. As the comment says, it is only used for ADC and not for anything else.

ryan-summers (approved):

PR #666, opened 2022-12-29
Nk/driver output sense

nkrackow:
This PR implements proper output current/voltage sensing as well as some fixes for the DAC like offsetting by one LSB for a current source `ChannelVariant`. 
(I will fix the docs in a separate PR)


src/hardware/driver/dac.rs
```diff
@@ -180,24 +187,28 @@ where
...
-                | CONFIG1::LDACMODE::ASYNC, // set to use SPI SYNC
+            | CONFIG1::DSDO::ENABLED // set to retain default
+            | CONFIG1::FSET::ENABLED // set to retain default
+            | CONFIG1::LDACMODE::ASYNC, // set to use SPI SYNC
         );
         // perform calibration
         // don't try to calibrate without driver because it will wait forever
         // dac.calibrate(delay);
-
+        delay.delay_us(200);
```
nkrackow, line 202: I have found a delay of more than 10 us necessary here, otherwise the current setting in the next line does not work. The delay after the reset does not have an effect (even though its probaply better to give it some time after a reset) so it must be specifically after writing the CONFIG1 reg. I couldn't find this in the datasheet.. Everywhere else >100 ns delays are enough between writes as per datasheet.

jordens: Probably a good idea to check the ALARM bit. It may well indicate that the DAC isn't ready yet.

src/hardware/driver/internal_adc.rs
```diff
@@ -1,16 +1,19 @@
...
     gpio::{gpiof::*, Analog},
-    stm32::{ADC1, ADC3},
+    stm32::{ADC1, ADC2, ADC3},
 };
 
 const V_REF: f32 = 2.048; // ADC reference voltage
-const R_SENSE: f32 = 0.1; // Driver output current sense resistor (Will maybe be something else on HW)
+
+// See schematic "diagnostics" sheet. Output voltage sense.
+const V_OFFSET: f32 = 1.; // divider/shifter opamp offset
```
jordens, line 12: Units? For all constants.

src/hardware/driver/mod.rs
```diff
@@ -51,7 +50,7 @@ impl ChannelVariant {
     const R_OUT_LN: f32 = 40.0; // Low noise side output resistor
     const R_OUT_HP: f32 = 0.068; // High power side output resistor
     const DIVIDER_HP: f32 = 1. / 41.; // High power side DAC output divider ratio
-    fn transimpedance(&self) -> f32 {
+    fn scale(&self) -> f32 {
```
jordens, line 53: Scale for what? Transimpedance of what? What's the difference? Please document and find unambiguous names.

src/hardware/driver/mod.rs
```diff
@@ -51,7 +50,7 @@ impl ChannelVariant {
     const R_OUT_LN: f32 = 40.0; // Low noise side output resistor
     const R_OUT_HP: f32 = 0.068; // High power side output resistor
     const DIVIDER_HP: f32 = 1. / 41.; // High power side DAC output divider ratio
```
jordens, line 52: Please ref the schematics and stick to the names. This is the "follower" right? So maybe "GAIN_FOLLOWER_HP"

src/hardware/driver/mod.rs
```diff
@@ -51,7 +50,7 @@ impl ChannelVariant {
     const R_OUT_LN: f32 = 40.0; // Low noise side output resistor
     const R_OUT_HP: f32 = 0.068; // High power side output resistor
```
jordens, line 51: Clear names, e.g. "R_SHUNT_LN". And always document units.
What does the "OUT" mean?

src/hardware/driver/internal_adc.rs
```diff
@@ -76,8 +85,8 @@ impl InternalAdc {
                 self.output_current.1.read_normalized().unwrap()
             }
         };
-        const SCALE: f32 = V_REF / R_SENSE; // Current sense scale       ToDo
-        const OFFSET: f32 = 0.0; // Current sense offset         ToDo
-        (ratio + OFFSET) * SCALE
+        let scale =
+            V_REF * I_SCALE / self.variant[ch as usize].transimpedance(); // current sense scale
```
jordens, line 89: Invert the function return to avoid the division.
And push `V_REF*I_SCALE` into the (single-use) method.

jordens (changes requested):

src/hardware/driver/dac.rs
nkrackow: Yes, that is what is done in 'calibration()' once we have fixed the DAC readback in HW.

jordens: Sure. I'd just keep track of it.

src/hardware/driver/internal_adc.rs
nkrackow: I would leave those constants here since they have to do with the ADC and current sensing. At least I will surely forget that the ADC reference voltage is to be found in the `ChannelVariant`.

src/hardware/driver/internal_adc.rs
nkrackow: I have commented the units now. Or did you mean I should be using a unit wrapper for the float value?

src/hardware/driver/internal_adc.rs
jordens: Ok. But do invert the method. `current_sense_{transconductance,scale,gain}()` maybe.

src/hardware/driver/internal_adc.rs
jordens: No. Just spell out the (hopefully SI) units in the comments.

src/hardware/driver/internal_adc.rs
nkrackow: its `current_sense_admittance` now

src/hardware/driver/internal_adc.rs
nkrackow: Yes, all SI

jordens (approved):

Issue #655, opened 2022-11-30
move the alarm topic somewhere else or reduce traffic


jordens:
As it's a really high rate topic, as part of settings it dominates all traffic there.
Move it elsewhere, either into a separate MQTT client use a will to have the broker handle the timeout.

jordens:
what about the will idea?

PR #681, opened 2023-02-20
Nk/driver alarm mqtt client

nkrackow:
Refactor of the MQTT alarm. Now there is another minimq client in the network stack that only subscribes to the `alarm` topic, separated from settings and telemetry. The alarm settings like timeout and reset are still there.
Addresses https://github.com/quartiq/stabilizer/issues/655


nkrackow:
Unf. I had to touch the lockin and dual-iir mains like this but I still think this is cleaner than the hack with the settings path before.

nkrackow:
ai-artiq-py is updated for this and thermostat doesn't need changes

src/net/alarm.rs
```diff
@@ -0,0 +1,183 @@
+use heapless::String;
+use miniconf::embedded_time;
+use minimq::embedded_nal::{IpAddr, TcpClientStack};
+
+use fugit::ExtU64;
+use miniconf::Miniconf;
+use serde::{Deserialize, Serialize};
+
+const TRUE: &[u8] = &[116, 114, 117, 101]; // "true" in raw bytes
```
jordens, line 9: `b"true"`?
And you use it exactly once. So inline it.

src/net/alarm.rs
```diff
@@ -0,0 +1,183 @@
...
+const FALSE: &[u8] = &[102, 97, 108, 115, 101]; // "false" in raw bytes
+
+#[derive(Clone, Copy, Debug, Miniconf, Serialize, Deserialize)]
+/// Driver alarm functionality
```
jordens, line 13: This isn't driver anymore. You'll need to make the docs generic.

src/net/alarm.rs
```diff
@@ -0,0 +1,183 @@
...
+///
+/// Driver features an alarm to ensure safe co-operation with other devices.
+/// The alarm is implemented via MQTT as a separate topic besides settings and telemetry.
+/// Publish "false" onto the alarm topic to indicate valid operating conditions. This renews the alarm timeout.
+/// Publishing "true" or failing to publish "false" for [`Self::timeout`] trips the alarm.
+/// In the case of Driver this will trip the [super::LaserInterlock].
+/// After the alarm is tripped, it has to be [rearm](Self::rearm())ed.
+/// For Driver this will happen when the [super::LaserInterlock] is cleared.
+pub struct AlarmSettings {
```
jordens, line 22: Just `Settings`. No need to prefix things with `Alarm`.

src/net/alarm.rs
```diff
@@ -0,0 +1,183 @@
...
+    pub fn update(
+        &mut self,
+    ) -> Result<Option<bool>, minimq::Error<Stack::Error>> {
+        if !self.mqtt.client().is_connected() {
+            // reset subscription if client disconnects
+            self.subscribed = false;
+        }
+        if !self.subscribed && self.mqtt.client().is_connected() {
+            self.subscribed = true;
+            log::info!("Alarm subscribed to {:?}", self.prefix);
```
jordens, line 165: Move this below the `subscribe()`.

src/net/alarm.rs
```diff
@@ -0,0 +1,183 @@
...
+            self.mqtt
+                .client()
+                .subscribe(&[(self.prefix.as_str()).into()], &[])
+                .map(|_| None) // return None if we just subscribed
+        } else {
+            self.mqtt
+                .poll(|_client, _topic, message, _properties| match message {
```
jordens, line 172: I think you can https://docs.rs/serde-json-core/latest/serde_json_core/fn.from_slice.html# this right away. No need to do it yourself.

src/net/alarm.rs
```diff
@@ -0,0 +1,183 @@
...
+            None
+        }
+    }
+}
+
+/// Stabilizer mqtt alarm listener.
+pub struct Alarm<
+    Stack,
+    Clock,
+    const MESSAGE_SIZE: usize,
```
jordens, line 114: Why not just 5 fixed?

jordens (commented):
couple nits. otherwise ok.
We'll want to overhaul the network users interface to reduce the coupling between different apps.

src/net/alarm.rs
```diff
@@ -161,17 +155,19 @@ where
...
+            log::info!("Alarm subscribed to {:?}", self.prefix);
+            result
         } else {
             self.mqtt
-                .poll(|_client, _topic, message, _properties| match message {
-                    TRUE => Some(true),
-                    FALSE => Some(false),
+                .poll(|_client, _topic, message, _properties| match from_slice(message) {
+                    Ok((true, 4)) => Some(true),
+                    Ok((false, 5)) => Some(false),
```
jordens, line 170:
```suggestion
                    Ok((b, _)) => Some(b),
```

And then clippy might be even wiser.

Issue #706, opened 2023-03-21
Turn off when LM75 temperatures are out of range


nkrackow:
There can currently be runaway heating conditions with Driver. https://github.com/sinara-hw/Driver/issues/18 
To prevent this in SW we should keep an eye on the LM75 temperatures and set the outputs to minimal current in case of a (hardcoded?) overtemperature read.

nkrackow:
Now implemented in https://github.com/quartiq/stabilizer/commit/136f0f54deb44afd4b8de392f94e0663c7facccf
The DAC does not need to be set to min current any more since the MCU interlock now switches off the LN and HP power supplies.

PR #708, opened 2023-03-22
Nk/driver temperature shutdown

nkrackow:
This PR adds a temperature monitoring task to check the Driver and headboard temperatures and turn off the channels to prevent runaway heating and damage in Driver. https://github.com/sinara-hw/Driver/issues/18

It was thereby necessary to refactor the way the interlock works by adding the feature that tripping the interlock implicitly also turns off both channels (aka triggers the state machines to transition through to the `Disabled` state). This means that the miniconf `output_enabled` state for both channels can be out of sync with the actual output state machine. This is a necessary sacrifice to be able to turn off the outputs and I think it doesn't matter that much. 
You still have to first set the miniconf setting to disabled first before resetting the laser interlock. So nothing changed from a users perspective except for no self destruction during a high output current interlock event. 
implements https://github.com/quartiq/stabilizer/issues/706

### DIFF
--- a/src/bin/driver.rs
+++ b/src/bin/driver.rs
@@ -523,17 +523,23 @@ mod app {
         if !old_settings.reset_laser_interlock
             && new_settings.reset_laser_interlock
         {
-            if c.shared.output_state.lock(|output| {
-                output[driver::Channel::HighPower as usize].is_enabled()
+            if c.shared.output_state.lock(|state| {
+                state[driver::Channel::HighPower as usize].is_enabled()
             }) {
                 log::error!("Cannot reset laser interlock while HighPower channel is enabled. Disable channel first.")
-            } else if c.shared.output_state.lock(|output| {
-                output[driver::Channel::LowNoise as usize].is_enabled()
+            } else if c.shared.output_state.lock(|state| {
+                state[driver::Channel::LowNoise as usize].is_enabled()
             }) {
                 log::error!("Cannot reset laser interlock while LowNoise channel is enabled. Disable channel first.")
             } else {
                 log::info!("Laser interlock reset.");
-                c.shared.laser_interlock.lock(|ilock| ilock.set(None, None));
+                c.shared.laser_interlock.lock(|ilock| {
+                    let _ = c // discard result, nothing to spawn here
+                        .shared
+                        .output_state
+                        .lock(|state| ilock.set(None, state));
+                });
+
                 c.shared.alarm_handle.lock(|handle| {
                     if let Some(millis) =
                         new_settings.alarm.rearm(handle.is_some())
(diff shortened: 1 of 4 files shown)


src/bin/driver.rs
```diff
@@ -770,12 +822,26 @@ mod app {
         {
             if (read > interlock_current) & interlock_asserted & output_en {
                 let ch = Channel::try_from(i).unwrap();
-                c.shared.laser_interlock.lock(|ilock| {
-                    ilock.set(Some(Reason::Overcurrent(Condition {
-                        channel: ch,
-                        threshold: *interlock_current,
-                        read: *read,
-                    })))
+                (c.shared.laser_interlock).lock(|ilock| {
```
nkrackow, line 825: The nested locking is done due to https://github.com/rtic-rs/rtic/issues/713

nkrackow:
The hardcoded valid temperature range obviously has to be discussed and potentially adapted still.

src/hardware/driver/output.rs
```diff
@@ -310,14 +310,20 @@ where
...
+            sm::States::RampCurrent => {
+                (Some(Output::<I2C>::RAMP_DELAY), result)
+            }
+            sm::States::SelftestShunt => {
+                (Some(Output::<I2C>::SET_DELAY), result)
+            }
+            sm::States::SelftestShort => {
+                (Some(Output::<I2C>::SET_DELAY), result)
+            }
+            _ => (None, result),
```
jordens, line 326: `(match .. {}, result)`

src/hardware/driver/mod.rs
```diff
@@ -123,11 +143,27 @@ impl LaserInterlock {
...
+        reason: Option<Reason>,
+        output_state: &mut [output::sm::StateMachine<output::Output<I2c1Proxy>>;
+                 2],
+    ) -> [Option<fugit::MillisDuration<u64>>; 2] {
         // only update if no reason yet or if clearing (remember first reason)
         if self.reason.is_none() || reason.is_none() {
             self.pin.set_state(reason.is_none().into());
+            // disable both outputs if interlock is tripped
             self.reason = reason;
+            if self.reason.is_some() {
```
jordens, line 157: A bit hard to parse mentally. Maybe clearer with a `match` or an inversion of thw two `if` s?

src/bin/driver.rs
```diff
@@ -719,13 +758,24 @@ mod app {
...
+                let delays = ilock.set(Some(reason), state);
+                delays.iter().enumerate().for_each(|(i, delay)| {
+                    if let Some(delay) = delay {
+                        handle_output_tick::spawn_after(
+                            delay.convert(),
+                            Channel::try_from(i).unwrap(),
+                        )
+                        .unwrap();
+                    }
+                });
```
jordens, line 776: This looks like something refactorable (it shows up five times). Or at least the part except the `spawn_after`.

nkrackow:
Not necessary any more due to https://github.com/quartiq/board_modifications/issues/6#issuecomment-1486856599 and https://github.com/quartiq/stabilizer/commit/136f0f54deb44afd4b8de392f94e0663c7facccf

PR #717, opened 2023-04-04
Updating runner to utilize specific probe

ryan-summers:
Updates the booster HITL to use a specific debug probe (there are now two, one for booster and one for stabilizer)


ryan-summers:
bors try

PR #722, opened 2023-04-11
Nk/more driver limits

nkrackow:
small changes to implement limits for HP channel and interlock tripping if limits out of range

### DIFF
--- a/src/bin/driver.rs
+++ b/src/bin/driver.rs
@@ -42,7 +42,10 @@ use stabilizer::{
         afe::Gain,
         dac::{Dac0Output, Dac1Output, DacCode},
         design_parameters, driver,
-        driver::{output, Channel, Condition, Location, Temperature},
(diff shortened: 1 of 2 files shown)


src/bin/driver.rs
```diff
@@ -550,22 +553,36 @@ mod app {
...
-                c.local.channel_range[0].end,
-            );
-        new_settings.low_noise.iir.y_max =
-            new_settings.low_noise.iir.y_max.clamp(
-                c.local.channel_range[0].start,
-                c.local.channel_range[0].end,
-            );
+        // check if the output limits are within the allowed range and trip the interlock if not
+        for ((range, limit), setting) in [
+            c.local.channel_range[0].clone(),
```
jordens, line 558: & should be enough. Why clone?

jordens (approved):

PR #745, opened 2023-07-04
update proc-macro2 (nightly regression)

jordens:


jordens:
It was supposed to address this: https://github.com/quartiq/stabilizer/actions/runs/5440429509/jobs/9893339611
But instead the previous dependabot updates pulled that update in.
Now nightly crashes instead (presumably not caused by the crate updates but unrelated things).
This leaves this PR as a simple indirect dependency updated across the board.

ryan-summers (approved):

PR #748, opened 2023-07-07
Fixing pounder bootup issues

### LINKED ISSUE #746: Stabilizer + Pounder panics on bootup
If the Pounder+Stabilizer combination is flashed and rebooted, it has been reported that it will panic without the debugger attached (as indicated by the red LEDs powered)

Upon using a debugger reset, the panic is resolved.

First reported by @jordens

ryan-summers:
Fixes #746 by waiting a fixed delay before initializing the AD9959 to allow the Pounder clock to stabilize

### DIFF
--- a/.cargo/config
+++ b/.cargo/config
@@ -1,7 +1,6 @@
 [target.'cfg(all(target_arch = "arm", target_os = "none"))']
-#runner = "probe-run --chip STM32H743ZITx --speed 30000"
+runner = "probe-run --chip STM32H743ZITx --speed 30000"
 # runner = "gdb-multiarch -q -x openocd.gdb"
-runner = "probe-rs-cli download --chip STM32H743ZITx"
 rustflags = [
     "-C", "link-arg=-Tlink.x",
     "-C", "link-arg=--nmagic",
--- a/.cargo/config
+++ b/.cargo/config
@@ -1,6 +1,7 @@
 [target.'cfg(all(target_arch = "arm", target_os = "none"))']
-runner = "probe-run --chip STM32H743ZITx --speed 30000"
+#runner = "probe-run --chip STM32H743ZITx --speed 30000"
 # runner = "gdb-multiarch -q -x openocd.gdb"
+runner = "probe-rs-cli download --chip STM32H743ZITx"
 rustflags = [
     "-C", "link-arg=-Tlink.x",
     "-C", "link-arg=--nmagic",
--- a/src/hardware/setup.rs
+++ b/src/hardware/setup.rs
@@ -870,35 +870,28 @@ pub fn setup(
             };
 
             #[cfg(not(feature = "pounder_v1_0"))]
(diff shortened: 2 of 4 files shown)


.cargo/config
```diff
@@ -1,6 +1,7 @@
 [target.'cfg(all(target_arch = "arm", target_os = "none"))']
-runner = "probe-run --chip STM32H743ZITx --speed 30000"
+#runner = "probe-run --chip STM32H743ZITx --speed 30000"
```
jordens, line 2: Revert

src/hardware/setup.rs
```diff
@@ -870,22 +870,35 @@ pub fn setup(
...
                 design_parameters::DDS_REF_CLK.to_Hz() as f32,
-                design_parameters::DDS_MULTIPLIER,
-            )
-            .unwrap();
+            );
+
+            // The oscillator to the AD9959 takes some time to turn on after first initialization.
+            // To accomodate this, we retry the initialization process until the configuration
+            // process completes successfully.
+            loop {
```
jordens, line 887: Nono. This would potentially wait forever.
Also it's not a good idea to do the configuration when the clock is just coming up and leave it there. There is no certain indicator that the clock is fine and the initialization is fine under those conditions.
Just a sufficient delay (take 10 ms) is the right thing to do.

jordens: Also we should explicitly `set_ext_clk(false)` once very early and wait after that for oscillator power up.

jordens (changes requested):

src/hardware/setup.rs
ryan-summers: >Just a sufficient delay (take 10 ms) is the right thing to do.

The main problem I have is that the datasheet for the internal clock does not specify a startup time. We have a watchdog going for 30s during initialization that would catch anything infinite. That being said, I have nothing opposed to inserting a predefined delay and calling it good. It would just be nicer to have justification for said delay.

> Also we should explicitly `set_ext_clk(false)` once very early and wait after that for oscillator power up.

This is done during `PounderDevices::new()` above, so we don't have to worry about it here.

jordens: The DDS PLL needs ~1 ms to lock. But that's not critical here since at this point we don't use it yet (but we should add a 1 ms delay after `configure_system_clock()` at this point). The XO has ~ 2 ms startup time. That's what we are waiting for. We do need to worry about that.

ryan-summers: Where are you getting the ~2ms startup time for the XO from? I don't see anything in the [datasheet](https://eu.mouser.com/datasheet/2/94/CCHD_950-2998865.pdf)

jordens: Extrapolation from [others of the same architecture/style](https://www.crystek.com/crystal/spec-sheets/clock/cchd-957.pdf) that have it spec'd.

jordens (approved):

PR #757, opened 2023-07-21
fix hitl, make beta ci optional

jordens:


.github/workflows/ci.yml
```diff
@@ -122,35 +122,3 @@ jobs:
           # Install depedencies at our cache location
           mv ../target/thumbv7em-none-eabihf/doc src/firmware
           mdbook build
-
-  hitl-trigger:
-    runs-on: ubuntu-latest
-    environment: hitl
-    if: ${{ github.event_name == 'merge_group' }}
```
ryan-summers, line 129: I think this condition here is critical to correct operation of the HITL + merge queue setup. If we don't have this condition, the job isn't marked as "Skipped" on a PR, but rather just never runs. With that setup, we can never actually ad it to the merge queue

ryan-summers (dismissed):

.github/workflows/ci.yml
```diff
@@ -58,8 +58,10 @@ jobs:
         include:
```
ryan-summers, line 58:
```suggestion
        continue-on-error: false
        include:
```

.github/workflows/ci.yml
```diff
@@ -49,7 +49,7 @@ jobs:
 
   compile:
     runs-on: ubuntu-latest
-    continue-on-error: ${{ matrix.toolchain == 'nightly' }}
+    continue-on-error: ${{ matrix.continue-on-error }}
```
ryan-summers, line 52: Maybe 
```suggestion
    continue-on-error: ${{ matrix.continue-on-error || 'false'}}
```

jordens: Ah. RTFM.

PR #767, opened 2023-08-10
new miniconf (trait split)

jordens:


src/bin/dual-iir.rs
```diff
@@ -82,7 +82,7 @@ pub struct Settings {
     ///
     /// # Value
     /// Any of the variants of [Gain] enclosed in double quotes.
-    #[miniconf(defer)]
+    #[tree()]
```
ryan-summers, line 85: Can we update it so that we can just annotate it with `#[tree]`? The parenthesis seem odd

jordens: https://github.com/quartiq/miniconf/pull/176/commits/fa6ff4bbc329c99fccfd1cc6cf4c776450959b89

ryan-summers (approved):

ryan-summers:
Reposting some of my information here. The HITL appears to be failing because the QUARTIQ office network has two addresses assigned to the `mqtt` name:
```
rsummers@irvine:/home/runner$ host -t a mqtt
mqtt.ber.quartiq.de is an alias for baby.ber.quartiq.de.
baby.ber.quartiq.de has address 10.34.17.1
baby.ber.quartiq.de has address 10.34.16.1

rsummers@irvine:/home/runner$ host -v -t a mqtt
Trying "mqtt.ber.quartiq.de"
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 52636
;; flags: qr aa rd ra; QUERY: 1, ANSWER: 3, AUTHORITY: 0, ADDITIONAL: 0

;; QUESTION SECTION:
;mqtt.ber.quartiq.de.           IN      A

;; ANSWER SECTION:
mqtt.ber.quartiq.de.    0       IN      CNAME   baby.ber.quartiq.de.
baby.ber.quartiq.de.    0       IN      A       10.34.17.1
baby.ber.quartiq.de.    0       IN      A       10.34.16.1

Received 102 bytes from fd60:7da2:8519:16::1#53 in 0 ms
```

The 10.34.17.1 address is provided first in the list, and pinging this address is successful, but it does not appear that there is an MQTT broker running on this machine. As such, the HITL is failing because Stabilizer is getting a bad IP address from DNS. Once we fix the DNS configuration, I believe HITL should succeed.

jordens:
Presumably Smoltcp should prefer the shortest route which is the .16.1 in this case. But I'll fix it for this case. 

Also we should also investigate srv record support and then use that as default.

ryan-summers:
> Presumably Smoltcp should prefer the shortest route which is the .16.1 in this case. But I'll fix it for this case.

I'm not sure how it would detect the "shortest" path route. Technically, smoltcp will give you _all_ of the routes, we just have it configured to only provide the first. I also have the NAL set up to only handle a single (first) result. How would you propose software identifies the "shortest" route. Would this be the "smallest" subnet that could be made between the device and the domain name?

> Also we should also investigate srv record support and then use that as default.

Hmm, I don't think the embedded-nal traits support providing a fully-qualified IP+port address back from the query currently, so this would require a NAL update as well.

jordens:
> > Presumably Smoltcp should prefer the shortest route which is the .16.1 in this case. But I'll fix it for this case.
> 
> I'm not sure how it would detect the "shortest" path route. Technically, smoltcp will give you _all_ of the routes, we just have it configured to only provide the first.

All of the routes?
The route to .16.1. needs zero hops while the one to .17.1 needs to go via the default route. Smoltcp knows that.
It's also should not discard records from the NS response but preferrably do something like happy eyes (what browsers do).

> I also have the NAL set up to only handle a single (first) result. How would you propose software identifies the "shortest" route. Would this be the "smallest" subnet that could be made between the device and the domain name?

on-link IP addresses should be preferred to those that need to go via the gateway.

> > Also we should also investigate srv record support and then use that as default.
> 
> Hmm, I don't think the embedded-nal traits support providing a fully-qualified IP+port address back from the query currently, so this would require a NAL update as well.

Yes. It's a further usability improvement step that came to mind. Independent of this issue.
It's also what mosquitto does.

PR #777, opened 2023-08-31
Adding support for DNS-based brokers

ryan-summers:
Adds support for using DNS-specified brokers. This PR also updates the minimq clients for the `settings` and `telemetry` users. We should probably wait until minimq and miniconf get released.

### DIFF
--- a/src/bin/dual-iir.rs
+++ b/src/bin/dual-iir.rs
@@ -220,7 +220,7 @@ mod app {
             clock,
             env!("CARGO_BIN_NAME"),
             stabilizer.net.mac_address,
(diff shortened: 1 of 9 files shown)


src/bin/dual-iir.rs
```diff
@@ -143,7 +143,7 @@ pub struct Settings {
     ///
     /// # Value
     /// See [signal_generator::BasicConfig#miniconf]
-    #[miniconf(defer(2))]
```
ryan-summers, line 146: Oops, depth needs to be set here.

jordens (approved):
LGTM

Issue #351, opened 2021-05-06
Store device identifier and broker in non-volatile memory


ryan-summers:
Currently, stabilizer generates an identifier using the binary name and MAC address. We should allow this default to be overridden by the user and the value should persist to non-volatile memory.

Using the unused flash bank for storage should be sufficient.

This is a follow-on to #276

jordens:
And in the same way we would store the broker (dns name or IP).

ryan-summers:
We also would want to store application-specific settings as well, e.g. for driver, as noted by @nkrackow

ryan-summers:
This will likely also necessitate the addition of a serial terminal to stabilizer to facilitate configuration of the initial network parameters. We may want to leverage some of the serial code on `Booster` for this

ryan-summers:
We should look into using https://crates.io/crates/sequential-storage to do this, as this is what the crate was designed for. Should eliminate excessive wear on the flash, but we may need to update the various linkers/flashing tools to not erase the storage part of flash during firmware updates.

ryan-summers:
@jordens I was looking in to this and I realized that Stabilizer doesn't expose a USB port on the front to use for communications. As such, I guess I'm not entirely sure how we'd expect the user to bootstrap some of these configurations, as right now the only communication interface is via Ethernet + MQTT.

However, the goal here is to provide some initial interface to allow users to set up that interface.

There's a UART debug header, but that seems like an error-prone way to ask users to configure things and doesn't necessarily feel like an improvement.

Did you have any ideas here as to how the user would update information like the MQTT broker or ID?

ryan-summers:
One difficulty I'm noticing here is that the sector size of flash on our device is 128KB. That means we're going to need to be able to buffer 128KB of data in RAM during an erase cycle. We have plenty of RAM (there's 1MB available), but I don't think the linker is set up to use it all. I'll look at how we can leverage it all. Specifically, using the sequential-storage crate, we'll need to allocate the 128KB on the stack.

I think what may make the most sense is to link the stack into a separate RAM section from the rest of memory, since the H7 segments the available RAM into a number of different contiguous sections.

jordens:
We only need to buffer the things to write, no?

ryan-summers:
The sequential storage crate requires a buffer of a flash sector in case it needs to do an element modification I believe. I'll have to look at why it requires this, it seems like an odd requirement.

jordens:
[...]

PR #789, opened 2023-10-13
Adding basic USB implementation

ryan-summers:
Adding a PR for basic USB terminal support.

For some reason, this isn't working yet on Windows (descriptor rquests are failing and windows can't recognize the device).

I suspect something's going wrong in the configuration of the peripheral and am still investigating.

Current USB interface is an echo chamber:
```
PS C:\Users\rsummers> python -m serial COM22
--- Miniterm on COM22  9600,8,N,1 ---
--- Quit: Ctrl+] | Menu: Ctrl+T | Help: Ctrl+T followed by Ctrl+H ---
echo: a
echo: s
echo: d
echo: f
echo: s
echo: a
echo: s
echo: d
echo: f
echo: s
echo: a
```

### DIFF
--- a/src/hardware/serial_terminal.rs
+++ b/src/hardware/serial_terminal.rs
@@ -12,7 +12,8 @@ impl Write for OutputBuffer {
         let data = s.as_bytes();
 
         // Write as much data as possible to the output buffer.
-        let Ok(mut grant) = self.producer.grant_max_remaining(data.len()) else {
+        let Ok(mut grant) = self.producer.grant_max_remaining(data.len())
+        else {
             // Output buffer is full, silently drop the data.
             return Ok(());
         };
--- a/src/hardware/serial_terminal.rs
+++ b/src/hardware/serial_terminal.rs
@@ -1,0 +1,71 @@
+use super::UsbBus;
+
+static OUTPUT_BUFFER: bbqueue::BBBuffer<1024> = bbqueue::BBBuffer::new();
+
+pub struct SerialTerminal {
+    usb_device: usb_device::device::UsbDevice<'static, UsbBus>,
+    usb_serial: usbd_serial::SerialPort<'static, UsbBus>,
+    output: bbqueue::Consumer<'static, 1024>,
+}
+
+impl SerialTerminal {
+    pub fn new(
+        usb_device: usb_device::device::UsbDevice<'static, UsbBus>,
+        usb_serial: usbd_serial::SerialPort<'static, UsbBus>,
+    ) -> Self {
+        let (_producer, consumer) = OUTPUT_BUFFER.try_split().unwrap();
+
+        Self {
+            usb_device,
+            usb_serial,
+            output: consumer,
+        }
+    }
+
+    fn flush(&mut self) {
+        let read = match self.output.read() {
+            Ok(grant) => grant,
+            Err(bbqueue::Error::InsufficientSize) => return,
+            err => err.unwrap(),
+        };
+
+        match self.usb_serial.write(read.buf()) {
+            Ok(count) => read.release(count),
+            Err(usbd_serial::UsbError::WouldBlock) => read.release(0),
+            Err(_) => {
+                let len = read.buf().len();
+                read.release(len);
+            }
+        }
+    }
+
+    pub fn process(&mut self) {
+        self.flush();
+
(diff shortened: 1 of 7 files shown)


ryan-summers:
The USB enumeration issue appears to be a result of https://github.com/rust-embedded-community/usb-device/issues/128 - I'll submit a patch to `usb-device` to get this fixed.

ryan-summers:
The intent is to follow-up with an actual settings interface in the future now that we have a USB interface, and then use the sequential-storage crate to store values in the unused flash. This supports #351

src/hardware/serial_terminal.rs
```diff
@@ -0,0 +1,77 @@
...
+                    // Currently, just echo it back.
+                    self.usb_serial.write(b"echo: ").ok();
+                    self.usb_serial.write(&[value]).ok();
+                    self.usb_serial.write(b"\n\r").ok();
+                }
+            }
+
+            Err(usbd_serial::UsbError::WouldBlock) => {}
+            Err(_) => {
+                // Flush the output buffer if USB is not connected.
```
jordens, line 66: Then you should be able to match the error explicitly.
And call this `clear the output buffer`.

src/hardware/serial_terminal.rs
```diff
@@ -0,0 +1,77 @@
...
+                self.output
+                    .read()
```
jordens, line 68: Are you sure once will be sufficient?

src/hardware/serial_terminal.rs
```diff
@@ -0,0 +1,77 @@
+use super::UsbBus;
+
+static OUTPUT_BUFFER: bbqueue::BBBuffer<1024> = bbqueue::BBBuffer::new();
```
jordens, line 3: let's start with 512

src/hardware/setup.rs
```diff
@@ -1003,6 +1009,67 @@ pub fn setup(
...
+            device.OTG2_HS_PWRCLK,
+            usb_n,
+            usb_p,
+            ccdr.peripheral.USB2OTG,
+            &ccdr.clocks,
+        );
+
+        // Generate a device serial number from the MAC address.
+        let serial_number =
+            cortex_m::singleton!(: Option<heapless::String<64>> = None)
```
jordens, line 1033: 17 is enough.

src/hardware/serial_terminal.rs
```diff
@@ -0,0 +1,77 @@
...
+        if !self.usb_device.poll(&mut [&mut self.usb_serial]) {
+            return;
+        }
+
+        let mut buffer = [0u8; 64];
+        match self.usb_serial.read(&mut buffer) {
+            Ok(count) => {
+                for &value in &buffer[..count] {
+                    // Currently, just echo it back.
+                    self.usb_serial.write(b"echo: ").ok();
```
jordens, line 58: Why not use the `producer` from the output queue?

src/hardware/serial_terminal.rs
ryan-summers: It's not exactly clear to me which error would indicate a USB connection not being present from the list of errors: https://docs.rs/usbd-serial/latest/usbd_serial/enum.UsbError.html - it seems reasonable to me to just clear the output buffer unconditionally if we can't read from the buffer for any reason, as there's definitely some communication breakdown that likely is occurring. Not sure how we should handle the errors otherwise.

src/hardware/serial_terminal.rs
ryan-summers: Fixed in 2bd057c35fc9230293f612668f01fc5aff581371

src/hardware/setup.rs
ryan-summers: Fixed in 2bd057c35fc9230293f612668f01fc5aff581371

src/hardware/serial_terminal.rs
ryan-summers: Fixed in 2bd057c35fc9230293f612668f01fc5aff581371

src/hardware/serial_terminal.rs
ryan-summers: Updated to do a while-loop approach in 2bd057c35fc9230293f612668f01fc5aff581371

jordens (approved):

PR #805, opened 2023-11-10
Adding initial support for configuring parameters via USB

### LINKED ISSUE #351: Store device identifier and broker in non-volatile memory

ryan-summers:
This PR fixes #351 by storing the MQTT broker address and the stabilizer ID in the unused flash bank.

TODO:
- [x] Update code to use parameters stored in flash on startup
- [x] Verify parameters are not lost during flashing

### DIFF
--- a/src/hardware/flash.rs
+++ b/src/hardware/flash.rs
@@ -1,0 +1,82 @@
+use stm32h7xx_hal::flash::{UnlockedFlashBank, LockedFlashBank};
+use embedded_storage::nor_flash::NorFlash;
+#[derive(Debug)]
+pub enum StorageError {
+    JsonDe(serde_json_core::de::Error),
+    JsonSer(serde_json_core::ser::Error),
+}
+
+impl From<serde_json_core::de::Error> for StorageError {
+    fn from(e: serde_json_core::de::Error) -> Self {
+        Self::JsonDe(e)
+    }
+}
+
+impl From<serde_json_core::ser::Error> for StorageError {
+    fn from(e: serde_json_core::ser::Error) -> Self {
+        Self::JsonSer(e)
+    }
+}
+
+impl sequential_storage::map::StorageItemError for StorageError {
+    fn is_buffer_too_small(&self) -> bool {
+        match self {
+            Self::JsonSer(serde_json_core::ser::Error::BufferFull) => true,
+            Self::JsonDe(serde_json_core::de::Error::EofWhileParsingString) => true,
+            Self::JsonDe(serde_json_core::de::Error::EofWhileParsingList) => true,
+            Self::JsonDe(serde_json_core::de::Error::EofWhileParsingObject) => true,
+            Self::JsonDe(serde_json_core::de::Error::EofWhileParsingNumber) => true,
+            Self::JsonDe(serde_json_core::de::Error::EofWhileParsingValue) => true,
+            _ => false,
+        }
+    }
+}
+
+macro_rules! storage_item {
+    ($len:literal, $key:literal, $name:ident) => {
+        pub struct $name(pub heapless::String<$len>);
+
+        impl sequential_storage::map::StorageItem for $name {
+            type Key = &'static str;
+            type Error = StorageError;
+
+            fn serialize_into(&self, buffer: &mut [u8]) -> Result<usize, Self::Error> {
+                Ok(serde_json_core::to_slice(&self.0, buffer)?)
+            }
+
+            fn deserialize_from(buffer: &[u8]) -> Result<(Self, usize), Self::Error> {
+                Ok(serde_json_core::from_slice(buffer).map(|(item, size)| (Self(item), size))?)
+            }
+
+            fn key(&self) -> Self::Key {
+                $key
+            }
+        }
+    }
+}
+
+
+storage_item!(23, "id", MqttIdentifier);
+storage_item!(255, "broker", BrokerAddress);
+
+pub struct FlashSettings {
+    flash: LockedFlashBank,
+}
+
+impl FlashSettings {
+    pub fn new(flash: LockedFlashBank) -> Self {
+        Self { flash }
+    }
+
+    pub fn store_item(&mut self, item: impl sequential_storage::map::StorageItem) {
+        let mut bank = self.flash.unlocked();
+        let range = (bank.address() as u32)..((bank.address() + bank.len()) as u32);
+        sequential_storage::map::store_item::<_, _, { <UnlockedFlashBank as NorFlash>::ERASE_SIZE } >(&mut bank, range, item).unwrap();
+    }
+
+    pub fn fetch_item<I: sequential_storage::map::StorageItem<Key=&'static str>>(&mut self, key: &'static str) -> Option<I> {
+        let mut bank = self.flash.unlocked();
+        let range = (bank.address() as u32)..((bank.address() + bank.len()) as u32);
+        sequential_storage::map::fetch_item(&mut bank, range, key).unwrap()
+    }
+}
--- a/src/hardware/flash.rs
+++ b/src/hardware/flash.rs
@@ -6,6 +6,9 @@ use stm32h7xx_hal::flash::{LockedFlashBank, UnlockedFlashBank};
 pub struct Settings {
     pub broker: heapless::String<255>,
     pub id: heapless::String<23>,
+    #[serde(skip)]
+    #[tree(skip)]
+    pub mac: smoltcp_nal::smoltcp::wire::EthernetAddress,
 }
 
 impl Settings {
@@ -16,7 +19,12 @@ impl Settings {
         Self {
             broker: "mqtt".into(),
             id,
+            mac,
         }
+    }
+
+    pub fn reset(&mut self) {
+        *self = Self::new(self.mac)
     }
 }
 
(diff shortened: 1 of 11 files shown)


src/hardware/serial_terminal.rs
```diff
@@ -25,28 +75,126 @@ impl Write for OutputBuffer {
...
+) {
+    let props: heapless::Vec<&'_ str, 2> = if let Some(prop) =
+        menu::argument_finder(item, args, "property").unwrap()
+    {
+        heapless::Vec::from_slice(&[prop]).unwrap()
+    } else {
+        heapless::Vec::from_slice(&["id", "broker"]).unwrap()
+    };
+
+    for prop in props {
```
ryan-summers, line 102: I'm not a huge fan of this. Maybe we just make the wrapper type a constant 255 buffer so we can maintain a standard type for both objects and then can simplify the iteration.

ryan-summers: Only difficulty is somehow assigning different keys to each object

src/hardware/flash.rs
```diff
@@ -0,0 +1,79 @@
...
+        };
+
+        Self { flash, settings }
+    }
+
+    pub fn save(&mut self) {
+        let mut bank = self.flash.unlocked();
+        let mut data = [0; 512];
+        let mut offset: usize = 0;
+        offset += serde_json_core::to_slice(&self.settings.broker, &mut data[offset..]).unwrap();
```
jordens, line 73: Why not just `postcard::to_slice(&settings, ...)` or `encdec` it?

src/hardware/serial_terminal.rs
```diff
@@ -1,5 +1,56 @@
...
+}
+
+const ROOT_MENU: menu::Menu<Context> = menu::Menu {
+    label: "root",
+    items: &[
+        &menu::Item {
+            command: "read",
+            help: Some("Read a property from the device:
+
+Available Properties:
```
jordens, line 26: Let's have a command `list` that replaces this help text as much as possible (same below for `write`). It can list the paths and print the doctring of the settings struct.

src/hardware/flash.rs
```diff
@@ -0,0 +1,89 @@
...
+        mut flash: LockedFlashBank,
+        mac: smoltcp_nal::smoltcp::wire::EthernetAddress,
+    ) -> Self {
+        let mut settings = Settings::new(mac);
+        let mut buffer = [0u8; 256];
+        let mut offset: usize = 0;
+
+        // We iteratively read the settings from flash to allow for easy expansion of the settings
+        // without losing data in the future when new fields are added.
+        flash.read(offset as u32, &mut buffer[..]).unwrap();
```
ryan-summers, line 39: This structure is kind of annoying and repetitive. We could use a proc-macro, but ideally we wouldn't need to and we could iterate over all the fields of the `settings` structure. However, the issue I have is that I need the `type` of the field as well to be able to deserialize it, and I don't think we can really iterate over types.

ryan-summers (commented):
@jordens This is now a functional implementation. A lot of this could easily be abstracted out into a separate crate that we could share between stabilizer/booster/thermostat as well. My only current question is figuring out how to be generic over the flash settings structure. It's a similar problem to miniconf, but subtley different.

Alternatively, if we don't really care about losing settings when the settings structure is modified, we could just serialize the whole structure directly into flash.

src/hardware/serial_terminal.rs
```diff
@@ -1,5 +1,65 @@
...
+    output: OutputBuffer,
+    flash: FlashSettings,
+}
+
+const ROOT_MENU: menu::Menu<Context> = menu::Menu {
+    label: "root",
+    items: &[
+        &menu::Item {
+            command: "reset",
+            help: Some("Reset Stabilizer to force new settings to take effect."),
```
jordens, line 17: * Let's make this already device and application agnostic. `s/Stabilizer/device`.
* Let's also add a command that resets the settings to their default (and distinguish clearly which command resets the CPU and which resets the settings).
* Does the term `property` match the term we use in `miniconf`? I think there it's `item`.
* Let's also state clearly here that these settings here are distinct from the application `mqtt` settings. Let's open an issue to discuss how we can use/expose the application settings through usb and save/load them in flash to be overriden by mqtt.
* Also I'd like to consider making this menu interface to miniconf (plus maybe persistent storage) a agnostic thing (crate).

src/hardware/serial_terminal.rs
```diff
@@ -1,5 +1,65 @@
...
+            item_type: menu::ItemType::Callback {
+                function: handle_property_write,
+                parameters: &[
+                    menu::Parameter::Mandatory {
+                        parameter_name: "property",
+                        help: Some("The name of the property to write."),
+                    },
+                    menu::Parameter::Mandatory {
+                        parameter_name: "value",
+                        help: Some("Specifies the value to be written to the property."),
```
jordens, line 54: ... `JSON` encoded.

src/hardware/serial_terminal.rs
```diff
@@ -25,28 +85,128 @@ impl Write for OutputBuffer {
...
+    }
+}
+
+fn handle_list(
+    _menu: &menu::Menu<Context>,
+    _item: &menu::Item<Context>,
+    _args: &[&str],
+    context: &mut Context,
+) {
+    writeln!(context, "Available properties:").unwrap();
```
jordens, line 104: Can this also print the docstring (or some other struct-associated string) of the settings struct?

src/hardware/serial_terminal.rs
```diff
@@ -25,28 +85,128 @@ impl Write for OutputBuffer {
...
+    for path in Settings::iter_paths::<heapless::String<32>>("/") {
+        let path = path.unwrap();
+        writeln!(context, "* {path}").unwrap();
```
jordens, line 107: Let's also print the current values as well as the ones that `reset` would establish while we are at it (instead of after each `read`?). Seems convenient and easy enough.

src/hardware/serial_terminal.rs
```diff
@@ -25,28 +85,128 @@ impl Write for OutputBuffer {
...
+                return;
+            }
+            Ok(len) => len,
+        };
+
+        let stringified = core::str::from_utf8(&buf[..len]).unwrap();
+        writeln!(context, "{path}: {stringified}").unwrap();
+    };
+
+    for path in Settings::iter_paths::<heapless::String<32>>("/") {
```
jordens, line 140: Ah. You had this. Let's remove it from `read` and just keep the dump in `list`.

src/hardware/serial_terminal.rs
```diff
@@ -25,28 +85,128 @@ impl Write for OutputBuffer {
...
 impl SerialTerminal {
     pub fn new(
         usb_device: usb_device::device::UsbDevice<'static, UsbBus>,
         usb_serial: usbd_serial::SerialPort<'static, UsbBus>,
+        flash: FlashSettings,
     ) -> Self {
         let (producer, consumer) = OUTPUT_BUFFER.try_split().unwrap();
 
+        let input_buffer =
+            cortex_m::singleton!(: [u8; 255] = [0; 255]).unwrap();
```
jordens, line 193: Why 255 (not 256)?
Is this the same 255/6 as the other one in the property read fn?

jordens (changes requested):
Feel free to defer things and convert into future issues where you see fit.

src/hardware/serial_terminal.rs
ryan-summers: Right now, no. I think we could expand the miniconf macro to expose the docstrings as metadata as well, but that would be a modification of miniconf. I'll open an issue.

ryan-summers: I added https://github.com/quartiq/miniconf/issues/187 for this functionality

src/hardware/serial_terminal.rs
ryan-summers: Fixed in https://github.com/quartiq/stabilizer/pull/805/commits/6f171975624442b05a68dee490f8802ecc7af40a

src/hardware/serial_terminal.rs
ryan-summers: Fixed in https://github.com/quartiq/stabilizer/pull/805/commits/6f171975624442b05a68dee490f8802ecc7af40a

src/hardware/serial_terminal.rs
ryan-summers: Fixed in 6182e1c790d6b1f1f5a8e824419c6f877bdcd4f1

src/hardware/serial_terminal.rs
ryan-summers: Fixed in 6182e1c790d6b1f1f5a8e824419c6f877bdcd4f1

src/hardware/serial_terminal.rs
jordens: I meant just the docstring of the struct, not of the fields. That would already cover a lot.

jordens (dismissed):
Nice.

src/hardware/serial_terminal.rs
```diff
@@ -1,5 +1,73 @@
...
+        &menu::Item {
+            command: "list",
+            help: Some("List all available settings and their current values."),
+            item_type: menu::ItemType::Callback {
+                function: handle_list,
+                parameters: &[],
+            },
+        },
+        &menu::Item {
+            command: "read",
```
jordens, line 40: Let's stick with `get`/`set` as in miniconf.

PR #811, opened 2023-11-15
Adding usage of sequential storage

ryan-summers:
Leveraging `sequential-storage` to make it faster to save settings and reduce wear on the flash. This PR also makes it so that the flash settings structure is not defined in flash itself, so if we add new items in the future, they will be dynamically loaded into the settings struct. This ensures that settings will not be lost between API changes.

This also aids in supporting https://github.com/quartiq/stabilizer/issues/822, as we can encode the settings item-by-item into the flash memory. This also means that settings for different profiles can co-exist (as long as they have unique settings names) so that they persist (and are shared) between different applications.


jordens (approved):
Sorry for the collision. LGTM!

PR #813, opened 2023-11-15
Moving serial settings to a separate crate

### LINKED ISSUE #808: Move serial port configuration menu to separate crate
The USB serial configuration menu is a useful utility for serial-based settings configuration used in multiple QUARTIQ firmwares. We should make it generic and have it exist as its own separate crate so we can leverage the functionality for Booster and Thermostat.

ryan-summers:
This PR fixes #808 by creating a new crate for serial settings and storage.

From the serial console:
```
Input was empty?

> help
AVAILABLE ITEMS:
  dfu
  service
  reboot
  factory-reset
  list
  get <item>
  set <item> <value>
  help [ <command> ]

> dfu
Reset to DFU is not supported on this device

> service
Service data not available

> list
Available items:
/broker: "test" [default: "mqtt"]
/id: "04-91-62-d2-a8-6f" [default: "04-91-62-d2-a8-6f"]
```


serial-settings/src/lib.rs
```diff
@@ -0,0 +1,306 @@
...
+
+impl<'a, UsbBus: usb_device::bus::UsbBus, Settings: SerialSettings, Flash: NorFlash + 'static> SerialTerminal<'a, UsbBus, Settings, Flash> {
+    pub fn new(
+        usb_device: usb_device::device::UsbDevice<'static, UsbBus>,
+        usb_serial: usbd_serial::SerialPort<'static, UsbBus>,
+        input_buffer: &'a mut [u8],
+        settings: Settings,
+        flash: Flash,
+    ) -> Self {
+        let (producer, consumer) = OUTPUT_BUFFER.try_split().unwrap();
```
ryan-summers, line 236: I suspect we can manage buffering without bbqueue, since there's synchronous access to the various objects.

serial-settings/src/lib.rs
```diff
@@ -0,0 +1,306 @@
...
+        let serialized = postcard::to_slice(&self.settings, &mut buf).unwrap();
+        self.flash.write(0, serialized).unwrap();
+    }
+}
+
+pub trait SerialSettings: for<'a> JsonCoreSlash<'a> + Serialize + Clone + 'static {
+    fn reset(&mut self);
+}
+
+fn make_menu<Settings: SerialSettings, Flash: NorFlash + 'static>() -> menu::Menu<'static, Context<Settings, Flash>> {
```
ryan-summers, line 30: I wonder if there's an easy way for us to make the menu options expandable by the user... Perhaps defer to the future

jordens: Maybe not needed now. All we want to add now is `service` and `dfu` AFAICT.

serial-settings/src/lib.rs
```diff
@@ -0,0 +1,306 @@
...
+    pub fn new(
+        usb_device: usb_device::device::UsbDevice<'static, UsbBus>,
+        usb_serial: usbd_serial::SerialPort<'static, UsbBus>,
+        input_buffer: &'a mut [u8],
+        settings: Settings,
+        flash: Flash,
+    ) -> Self {
+        let (producer, consumer) = OUTPUT_BUFFER.try_split().unwrap();
+
+        // TODO: Attempt to load from flash first.
```
ryan-summers, line 238: Need to figure out how to load settings from flash and fall-back if failing. In stabilizer, we modify the read-back settings, and I'm not sure the best way to handle that here.

serial-settings/src/lib.rs
```diff
@@ -0,0 +1,306 @@
...
+
+    pub fn process(&mut self) {
+        self.flush();
+
+        if !self.usb_device.poll(&mut [&mut self.usb_serial]) {
+            return;
+        }
+
+        let mut buffer = [0u8; 64];
+        match self.usb_serial.read(&mut buffer) {
```
ryan-summers, line 286: I think it may be beneficial to pull out any concept of USB and operate on `Read` / `Write` traits instead, as this would allow this to function over i.e. serial UART as well.

jordens: exactly.

serial-settings/src/lib.rs
```diff
@@ -0,0 +1,306 @@
...
+
+struct Context<Settings: SerialSettings, Flash: NorFlash + 'static> {
+    output: OutputBuffer,
+    flash: Flash,
+    settings: Settings,
+}
+
+impl<Settings: SerialSettings, Flash: NorFlash + 'static> Context<Settings, Flash> {
+    fn save(&mut self) {
+        self.flash.erase(0, self.flash.capacity() as u32).unwrap();
```
ryan-summers, line 18: No reason to erase all of the flash bank, just erase the area we're going to use.

serial-settings/src/lib.rs
jordens: be generic over `<S: Tree + ..., F: FnOnce() -> S>` with `F` providing the default fallback `S`.

serial-settings/src/lib.rs
jordens: Oh yes.

serial-settings/src/lib.rs
ryan-summers: Both `service` and `dfu` are not things we have on stabilizer. We can add them, but resetting to DFU mode is generally an involved process, so I'd imagine it's going to be a function provided on the trait object.

serial-settings/Cargo.toml
```diff
@@ -0,0 +1,16 @@
...
+name = "serial-settings"
+version = "0.1.0"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+embedded-storage = "0.3"
+miniconf = "0.9"
+menu = {path = "../../../rust-embedded-community/menu"}
```
ryan-summers, line 11: Currently need https://github.com/rust-embedded-community/menu/pull/15 to make this all work

serial-settings/src/lib.rs
ryan-summers: I added an abstraction in214d29fce5cf46e5cbbd317e5ac0e0051034937b, but I'm not the biggest fan of it because it means that there's `service` and `dfu` menu options in stabilizer that don't do anything currently (although we could add them).

Ideally, these items wouldn't even exist in the menu if they're not desired, but we can hold off on this for the future

ryan-summers:
@jordens I'm curious to see if you have any thoughts with the current design. It's definitely not perfect, but I think it's at a decent point for now. I'm going to poke at integrating it into Booster to see how it goes.

serial-settings/src/lib.rs
```diff
@@ -0,0 +1,418 @@
...
+}
+
+impl<T> Platform for T {}
+
+pub struct Context<
+    'a,
+    I: Interface,
+    S: Settings,
+    P: Platform,
+    Flash: NorFlash + 'static,
```
ryan-summers, line 60: This implies that the settings will be persisted to flash, but it's also perfectly valid to persist them to something else, like EEPROM. Ultimately, this just needs to be some non-volatile memory location.

Actually, thinking about it, it doesn't even have to be non-volatile. Should we look for a different interface to use for this?

ryan-summers: I think https://docs.rs/embedded-storage/latest/embedded_storage/trait.Storage.html is the correct trait bound here instead.

jordens: What about moving `save()` into the `Settings` trait and remove it completely from `Context`?

jordens: Or move it into `Platform`

jordens: Also maybe move `Settings` and `Interface` into aassociated types of `Platform`?

ryan-summers: I moved the various types into associated types of `Platform`, but I haven't touched `save()`.

I'm trying to understand why we would want to move it. Right now, it's abstracting away the concept of saving settings from the user. If we made it an associated function of `Platform`, that means the user could override it with some custom saving method. However, how does that make sense if we're internally loading the settings from the provided `storage` mechanism during initialization? If they were to save it differently, we couldn't load those settings.

It seems like what's currently implemented should be fine for now.

serial-settings/src/lib.rs
```diff
@@ -0,0 +1,374 @@
...
+        // TODO: Validate it first?
+        match context
+            .platform
+            .settings_mut()
+            .set_json(key, value.as_bytes())
+        {
+            Ok(_) => match context.platform.save(context.buffer) {
+                Ok(_) => {
+                    writeln!(
+                            context,
```
jordens, line 229: I'm a bit unhappy about writing to `context` vs writing to `context.platform.interface_mut()`. They are significantly different.

jordens: Also we should check that the worst case write doesn't unduly starve the other things happening at idle prio (network, streaming IIRC).

serial-settings/src/lib.rs
```diff
@@ -0,0 +1,374 @@
...
+    fn cmd(&mut self, cmd: &str);
+
+    /// Return a mutable reference to the `Interface`.
+    fn interface_mut(&mut self) -> &mut Self::Interface;
+
+    /// Return a reference to the `Settings`
+    fn settings(&self) -> &Self::Settings;
+
+    /// Return a mutable reference to the `Settings`.
+    fn settings_mut(&mut self) -> &mut Self::Settings;
```
ryan-summers, line 91: I'm not following why `P::Settings` aren't just a member of `Context`. It seems like extra work for the implementor of `platform` to have to provide these when in reality the settings are internally managed by this library.

serial-settings/src/lib.rs
```diff
@@ -0,0 +1,374 @@
...
+    type Settings: Settings;
+
+    /// `load()`/`save()` Error type
+    type Error: core::fmt::Debug;
+
+    /// Load the settings from storage
+    fn load(&mut self, buffer: &mut [u8]) -> Result<(), Self::Error>;
+
+    /// Save the settings to storage
+    fn save(&mut self, buffer: &mut [u8]) -> Result<(), Self::Error>;
```
ryan-summers, line 79: Seems like these should accept/return a `Self::Settings` instead of the buffer

serial-settings/src/lib.rs
```diff
@@ -0,0 +1,374 @@
...
+
+    /// Execute a platform specific command.
+    fn cmd(&mut self, cmd: &str);
+
+    /// Return a mutable reference to the `Interface`.
+    fn interface_mut(&mut self) -> &mut Self::Interface;
```
ryan-summers, line 85: Similarly, it feels odd that the interface is owned by the underlying `Platform` implementor instead of by the `Context`. The menu is intended to be the primary owner of the interface for communication (at least in my mind).

ryan-summers (commented):
@jordens Some questisons/comments on the design. Overall, I think we're pretty close to being able to publish this as something. Do we want to keep it in the stabilizer repo for now, or split it into (yet another) repo?

serial-settings/src/lib.rs
ryan-summers: Indeed! It should not be using `write_all()`, but instead just `write()`. However, I believe writing to `context` instead of `interface_mut()` is correct here, as writing to `interface_mut()` will indefinitely block if `WriteReady::write_ready()` is false, which will essentially stall the USB port and starve the idle task.

jordens: The problem is we don't write to `context` everywhere currently.

ryan-summers: Gotcha. I think this is a good reason why the interface should be explicitly owned by the `Context`, as it could then wrap the `Interface` in something that properly implements `core::fmt::Write`

serial-settings/src/lib.rs
jordens: The `Platform` equally accesses and uses the interface.
Ideally we'd want to merge `Platform` and `Context` into one (provided methods on the trait). But that seemed a bit hard when I tried.

serial-settings/src/lib.rs
jordens: In addition to the other points, it was less code to have Platform own settings and interface.

serial-settings/src/lib.rs
jordens: And also potentially get rid of the `Runner` as it doesn't seem to do much.

serial-settings/src/lib.rs
jordens: It doesn't seem appropriate to let this crate impl that trait.

ryan-summers: I fixed this in 2c391b669991052b5c450110bebcc7083ec989d4

serial-settings/src/lib.rs
ryan-summers: I've come around on `Platform` owning the interface, since indeed it seems like the platform really does need to use it quite often.

serial-settings/src/lib.rs
ryan-summers: I ended up removing `load()` entirely, since it didn't really serve much of a purpose (it was only being called once in the actual constructor of `Runner`). Instead, `Platform` is responsible for initializing its settings, which aligns with normal means of object construction.

ryan-summers:
> @jordens Some questisons/comments on the design. Overall, I think we're pretty close to being able to publish this as something. Do we want to keep it in the stabilizer repo for now, or split it into (yet another) repo?

Two last comments/discussions left above, but otherwise I think things here are at a nice state. Ping on the question above as well.

serial-settings/src/lib.rs
jordens: To me "proper ownership" doesn't have clear and easy rules.

Settings changing outside this crate is something I can see coming up soon, e.g. also saving the application settings in flash and using that as the startup value (shadowed then later by mqtt where available and retained, I think I mentioned this somewhere else as well).

jordens:
Let's split it once you want to use it on Booster. Is that now?

serial-settings/src/lib.rs
```diff
@@ -0,0 +1,372 @@
...
+        writer.write_str(s)
+    }
+}
+
+struct Writer<'a, T: WriteReady + EioWrite>(&'a mut T);
+
+impl<'a, T: embedded_io::WriteReady + embedded_io::Write> core::fmt::Write
+    for Writer<'a, T>
+{
+    fn write_str(&mut self, s: &str) -> core::fmt::Result {
```
jordens, line 313: I still have an issue with this. Why does this crate need to make the write best effort? That seems like a thing for a downstream crate to decide.

jordens (changes requested):

serial-settings/src/lib.rs
ryan-summers: Good point. I'll update it to not make assumptions about the writability of data.

jordens: I suspect that will also allow removing the dependency on eio traits from this crate.

ryan-summers: We need the `Read` component as well for the `process()` function to feed data to the menu interface. While we could replace the eio Write methods, it feels really odd to use half e-io and half core::fmt

jordens: agreed

jordens: But this does tie into the notion that `Runner` doesn't do much other than encapsulating `eio::Read`.

ryan-summers: Yeah, I recognize that. It largely comes down to whether we want this to be a "batteries included" approach or not. While I'm generally not a huge fan of making assumptions for the user, I think the `Runner` does eliminate a fair amount of boilerplate for the end user. Additionally, I added blanket-impls of the e-io traits for `usbd_serial::SerialPort`, so USB serial devices won't need any additional code to handle reading/writing data.

Ultimately, I'd say that we should keep the status quo and re-evaluate in the future if we want.

ryan-summers:
> Let's split it once you want to use it on Booster. Is that now?

Meh, there's no huge rush to move it onto Booster yet, so maybe we just let it live here for the time being. We can break it out into a separate crate when we feel like it's mature enough.

PR #820, opened 2023-11-30
Adding support for rebooting to DFU

### LINKED ISSUE #818: support reboot into dfu
* see `booster` for how this is done on F4
* research  robust sequence for H7

https://github.com/quartiq/stabilizer/blob/19c6fa6381f288058a1c49d7bddc3a1adf0a1ee8/src/hardware/flash.rs#L126

ryan-summers:
Added support for executing the system DFU bootloader when requested from the USB interface by setting a flag in noinit RAM and then resetting the systen. Upon system reset, it checks this noinit region for the flag and jumps to the DFU bootloader if it is set.

Fixes #818

### DIFF
--- a/src/hardware/flash.rs
+++ b/src/hardware/flash.rs
@@ -131,7 +131,7 @@ impl serial_settings::Platform for SerialSettingsPlatform {
             _ => {
                 writeln!(
                     self.interface_mut(),
-                    "Invalid platform command `{cmd}`"
+                    "Invalid platform command: `{cmd}` not in [`dfu`, `reboot`]"
                 )
                 .ok();
             }


jordens:
https://community.st.com/t5/stm32-mcus-embedded-software/jump-issue-with-custom-bootloader-stm32h7/m-p/186556/highlight/true#M11509 might also have some pieces.

The alternative approach would be to (instead of preparing peripherals manually) write a flag somewhere, do a full cpu reset, and then early in `setup()` check the flag and jump to dfu. Like [here](https://community.st.com/t5/stm32-mcus-embedded-software/jump-to-internal-bootloader-with-stm32f7/m-p/365587). That has the advantage of leveraging the cpu reset circuitry.

ryan-summers:
I suspect the key was that I was invoking the DFU bootloader from the USB task, which runs in an ISR binding. I like the idea of a boot flag in no-init RAM, since that's more future-proof (and far less code)

jordens:
Maybe also some code from the persistent crash info crate can be stolen for the flag handling.

ryan-summers:
@jordens Alright, this should now be functional. Thanks for the pointers!

jordens (changes requested):

CHANGELOG.md
```diff
@@ -5,6 +5,16 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## [Unreleased](https://github.com/quartiq/stabilizer/compare/v0.9.0...main)
+
+### Added
+* Support added for pounder v1.2
```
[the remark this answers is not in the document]
ryan-summers: Copy-paste fail. Thanks

book/src/setup.md
```diff
@@ -124,15 +123,24 @@ and applying power again.
...
+> **Note:** If there is already newer firmware running on Stabilizer that supports the USB serial
+> interface, there is no need to remove Stabilizer from the crate or disconnect any existing
+> connectors/power supplies or to jumper the BOOT0 pin. Instead, open the serial port on Stabilizer
+> and request it to enter DFU mode:
+> ```bash
+> python -m serial <serial-port>
+> > platform dfu
+> ```
+>
+> After the device is in DFU mode, use the `dfu-util` command specified in the instructions below,
```
jordens, line 138: The instruction flow here could be simplified and clarified a lot. But let's leave it.

src/hardware/flash.rs
```diff
@@ -126,12 +127,15 @@ impl serial_settings::Platform for SerialSettingsPlatform {
...
             "reboot" => cortex_m::peripheral::SCB::sys_reset(),
-            _ => writeln!(
-                self.interface_mut(),
-                "Invalid platform command `{cmd}`"
-            ),
+            "dfu" => platform::start_dfu_reboot(),
+            _ => {
+                writeln!(
+                    self.interface_mut(),
+                    "Invalid platform command `{cmd}`"
```
jordens, line 134:
```suggestion
                    "Invalid platform command: `{cmd}` not in [`dfu`, `reboot`]"
```

jordens (changes requested):
Just that one tweak adding supported platform commands to the error message. Then good to go.

jordens (approved):

Issue #822, opened 2023-12-01
support changing applications settings via serial-settings and load/store in flash


jordens:
* On boot the initial settings are loaded from flash, or `Default::default()` if invalid
* Then they are altered from retained, republished, and altered from non-retained pub messages on mqtt as currently done
* serial terminal allows accessing them (get and set) as well
* serial terminal allows comitting them to flash

Impact/feasibility to be determined.
This allows full standalone operation without network/broker.

ryan-summers:
Theoretically this is quite easy to implement. The main hurdle that I foresee is that the `Settings` structure is owned by the Miniconf client and the serial interface respectively, so there's some thought that would need to go into how to maintain synchronization between these two structures.

If we could somehow create a singleton of the settings that was referenced by the MQTT client and the serial interface, I believe this would become relatively trivial to implement.

jordens:
I would only ever modify stored settings over serial and only load settings from flash on boot (like broker and IP currently). That solves repub.

ryan-summers:
Ah so the intent here would be to make the serial interface effectively only touch flash, as opposed to modifying the run-time state of the device? It seems like that would be a bit... unexpected? I don't see anything blocking us from allowing the serial interface to affect the run-time state as well outside of software complexity.

jordens:
As for broker and IP. Those also need a reboot. If we call the flash settings to be the boot settings then all is well defined.
And as you say, repub races against potential host retained settings. Hard to clear all corner cases. Better leave simple settings precedence rules (default < flash < mqtt)

ryan-summers:
I agree w.r.t. the broker and IP settings being boot-time configuration, but the application run-time settings is what this issue is about, correct?

It seems like there would be limited usefulness to configuring run-time settings over serial if you had to save them and reboot the machine constantly to have them take effect.

I would expect that:
1. Runtime settings would immediately apply to the device operational state when configured via Serial
2. There would be an option for me to "save" the operational state via the serial console when I'm happy with the configuration
   * Maybe this would be automatic, similar to how we auto-save the broker and IP currently
3. I could harmoniously use the MQTT interface to configure as well, but these values would not be auto-saved to flash
4. There would be some MQTT command to save the device configuration to flash so they're used on the next boot

Am I missing something here?

jordens:
* There would quite some explaining, misunderstanding and potential frustration about why broker/ip need a reboot and e.g. streaming target doesn't.
[...]

PR #862, opened 2024-03-27
Add support for saving the active configuration to flash

### LINKED ISSUE #859: Add ability to save active settings to flash
Add functionality to save the currently-operational settings into device flash memory via the USB interface. This was identified in https://github.com/quartiq/stabilizer/pull/857#discussion_r1539080660

ryan-summers:
This PR fixes #859 by adding a platform command to save the active running settings to flash.

Below is output during testing. CH0 is configured as a PID with Kp=2 via MQTT and then the `save-active` command was used to update the flash-based settings.
```
> list
/dual_iir/afe/0: "G1" [default: "G1"]
/dual_iir/afe/1: "G1" [default: "G1"]
/dual_iir/iir_ch/0/0: {"ba":[1.0,0.0,0.0,0.0,0.0],"u":0.0,"min":-32767.0,"max":32767.0} [default: {"ba":[1.0,0.0,0.0,0.0,0.0],"u":0.0,"min":-32767.0,"max":32767.0}]
/dual_iir/iir_ch/1/0: {"ba":[1.0,0.0,0.0,0.0,0.0],"u":0.0,"min":-32767.0,"max":32767.0} [default: {"ba":[1.0,0.0,0.0,0.0,0.0],"u":0.0,"min":-32767.0,"max":32767.0}]
/dual_iir/allow_hold: false [default: false]
/dual_iir/force_hold: false [default: false]
/dual_iir/telemetry_period: 10 [default: 10]
/dual_iir/stream_target: {"ip":[0,0,0,0],"port":0} [default: {"ip":[0,0,0,0],"port":0}]
/dual_iir/signal_generator/0/signal: "Cosine" [default: "Cosine"]
/dual_iir/signal_generator/0/frequency: 1000.0 [default: 1000.0]
/dual_iir/signal_generator/0/symmetry: 0.5 [default: 0.5]
/dual_iir/signal_generator/0/amplitude: 0.0 [default: 0.0]
/dual_iir/signal_generator/0/phase: 0.0 [default: 0.0]
/dual_iir/signal_generator/1/signal: "Cosine" [default: "Cosine"]
/dual_iir/signal_generator/1/frequency: 1000.0 [default: 1000.0]
/dual_iir/signal_generator/1/symmetry: 0.5 [default: 0.5]
/dual_iir/signal_generator/1/amplitude: 0.0 [default: 0.0]
/dual_iir/signal_generator/1/phase: 0.0 [default: 0.0]
/net/broker: "mqtt" [default: "mqtt"]
/net/id: "04-91-62-d2-a8-6f" [default: "04-91-62-d2-a8-6f"]
/net/ip: "0.0.0.0" [default: "0.0.0.0"]

> platform save-active
Currently active settings saved.

> list
/dual_iir/afe/0: "G1" [default: "G1"]
/dual_iir/afe/1: "G1" [default: "G1"]
/dual_iir/iir_ch/0/0: {"ba":[2.0,0.0,0.0,-0.0,-0.0],"u":0.0,"min":-32767.0,"max":32767.0} [default: {"ba":[1.0,0.0,0.0,0.0,0.0],"u":0.0,"min":-32767.0,"max":32767.0}]
/dual_iir/iir_ch/1/0: {"ba":[1.0,0.0,0.0,0.0,0.0],"u":0.0,"min":-32767.0,"max":32767.0} [default: {"ba":[1.0,0.0,0.0,0.0,0.0],"u":0.0,"min":-32767.0,"max":32767.0}]
/dual_iir/allow_hold: false [default: false]
/dual_iir/force_hold: false [default: false]
/dual_iir/telemetry_period: 10 [default: 10]
/dual_iir/stream_target: {"ip":[0,0,0,0],"port":0} [default: {"ip":[0,0,0,0],"port":0}]
/dual_iir/signal_generator/0/signal: "Cosine" [default: "Cosine"]
/dual_iir/signal_generator/0/frequency: 1000.0 [default: 1000.0]
/dual_iir/signal_generator/0/symmetry: 0.5 [default: 0.5]
/dual_iir/signal_generator/0/amplitude: 0.0 [default: 0.0]
/dual_iir/signal_generator/0/phase: 0.0 [default: 0.0]
/dual_iir/signal_generator/1/signal: "Cosine" [default: "Cosine"]
/dual_iir/signal_generator/1/frequency: 1000.0 [default: 1000.0]
/dual_iir/signal_generator/1/symmetry: 0.5 [default: 0.5]
/dual_iir/signal_generator/1/amplitude: 0.0 [default: 0.0]
/dual_iir/signal_generator/1/phase: 0.0 [default: 0.0]
/net/broker: "mqtt" [default: "mqtt"]
/net/id: "04-91-62-d2-a8-6f" [default: "04-91-62-d2-a8-6f"]
/net/ip: "0.0.0.0" [default: "0.0.0.0"]
```

### DIFF
--- a/src/settings.rs
+++ b/src/settings.rs
@@ -153,6 +153,8 @@ pub struct SerialSettingsPlatform<C, const Y: usize> {
     /// The storage mechanism used to persist settings to between boots.
     pub storage: Flash,
 
+    pub active_settings: C,
+
     /// Metadata associated with the application
     pub metadata: &'static ApplicationMetadata,
 }
@@ -221,10 +223,19 @@ where
         Ok(())
     }
 
-    fn cmd(&mut self, cmd: &str) {
+    fn cmd(&mut self, cmd: &str, buf: &mut [u8]) -> Result<(), Self::Error> {
         match cmd {
             "reboot" => cortex_m::peripheral::SCB::sys_reset(),
             "dfu" => platform::start_dfu_reboot(),
+            "save-active" => {
+                self.settings = self.active_settings.clone();
+                self.save(buf)?;
+                writeln!(
+                    &mut self.interface,
+                    "Currently active settings saved."
+                )
+                .unwrap();
+            }
             "service" => {
                 writeln!(
                     &mut self.interface,
@@ -262,11 +273,13 @@ where
             _ => {
                 writeln!(
                     self.interface_mut(),
-                    "Invalid platform command: `{cmd}` not in [`dfu`, `reboot`, `service`]"
+                    "Invalid platform command: `{cmd}` not in [`dfu`, `reboot`, `service`, `save-active`]"
                 )
                 .ok();
             }
         }
+
+        Ok(())
     }
 
     fn settings(&self) -> &Self::Settings {
--- a/book/src/setup.md
+++ b/book/src/setup.md
@@ -23,7 +23,7 @@ Power Stabilizer through **exactly one** of the following mechanisms.
 
 ## USB Configuration
 
-The USB port can be used to bootstrap Stabilizer and configure all internal settings. This is useful
+The USB port can be used to bootstrap Stabilizer and configure initial settings. This is useful
 either when first configuring the MQTT connection or when operating Stabilizer in standalone mode
 (i.e. without an ethernet connection or an MQTT broker).
 
@@ -33,20 +33,19 @@ provides a simple, easy-to-use terminal emulator:
 python -m serial <port>
 ```
 
-Once you have opened the port, you can use the provided menu to update any of Stabilizers runtime
-settings. All settings configured via the USB interface are only applied when Stabilizer first boots
-up. Any modifications that occur after boot over the MQTT interface will occur after the initial
-settings configured via USB.
+Once you have opened the port, you can use the provided menu to update any of Stabilizers settings.
+All settings configured via the USB interface are only applied when Stabilizer first boots up.
 
-> **Note:** Settings configured via USB do not take immediat effect. Instead, they will apply after
+> **Note:** Settings configured via USB do not take immediate effect. Instead, they will apply after
 > the device is rebooted.
 
 > **Note:** Settings configured via USB are only initial settings. Any modifications
 > that occur after startup (i.e. via MQTT) will take precedent over those configured via USB. MQTT
 > settings can only be persisted or retained via the MQTT broker.
 
-> **Note:** Saving settings via the USB port only saves the initial settings. It does not take into
-> account the current operational settings that may have been applied via MQTT.
+> **Note:** Saving settings through the USB interface only saves the initial settings. It does not
+> take into account the current operational settings that may have been applied via MQTT. To save
+> the active settings, use the `platform save-active` command.
 
 ## Network and DHCP
 
--- a/book/src/setup.md
+++ b/book/src/setup.md
@@ -43,9 +43,9 @@ All settings configured via the USB interface are only applied when Stabilizer f
 > that occur after startup (i.e. via MQTT) will take precedent over those configured via USB. MQTT
 > settings can only be persisted or retained via the MQTT broker.
 
-> **Note:** Saving settings through the USB interface only saves the initial settings. It does not
-> take into account the current operational settings that may have been applied via MQTT. To save
-> the active settings, use the `platform save-active` command.
+> **Note:** Modifying settings through the USB interface only modifies the initial settings. It does
+> not take into account the current operational settings that may have been applied via MQTT. To
+> save the active settings, use the `platform save-active` command.
 
 ## Network and DHCP
 
--- a/book/src/usage.md
+++ b/book/src/usage.md
@@ -116,3 +116,10 @@ Livestreamed data is sent with "best effort" - it's possible that data may be lo
 network congestion or by Stabilizer.
 
 Refer to the the respective [application documentation](overview.md#applications) for more information.
+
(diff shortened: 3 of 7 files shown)


src/settings.rs
```diff
@@ -153,6 +153,8 @@ pub struct SerialSettingsPlatform<C, const Y: usize> {
     /// The storage mechanism used to persist settings to between boots.
     pub storage: Flash,
 
+    pub active_settings: C,
```
jordens, line 156: Meh. There are at least four places where we keep the setting now...

I'd rather have the poll call into serial settings return some sort of flag that the app then uses to push the settings. Or save that flag in the platform.

jordens (commented):
I think we're hitting the limits of the design here.

What if we drop the `AppSettings` in serial settings, as well as the backing copy in the Miniconf MQTT client, as well as this `active_settings`. The original requirement for multiple settings copies  (IIRC only) arose because serialization/deserialization could take a while and locking the settings for that time would have starved the low latency process task. Maybe that doesn't scale.

If we can split that into only two settings structs, we might cover all bases. One active low latency that has very short lock durations, and a single "slower" backing one for serialization/deserialization/update tasks in (but not owned by) miniconf and serial settings/flash. The latter would be passed into the slow poll calls as `&mut` and flipped onto the low latency active.

For serial settings that would mean the operations actually operate on the single backing buffer (and implicitly on the low latency one). We'd have to make `save` explicit. It would also remove lots of docs and explanations around what which op actually affects.

For the mqtt client, a wrinkle would be around wanting to republish a setting when changed via USB. Or foregoing that for now and instead documenting the behavior. It would affect the `miniconf` crate as well.

In general passing `&mut Settings` into the different "frontends" does sound better than having them each own theirs.

There is also an uncertainty around `handled_update` in miniconf-mqtt. I wonder if we actually use that feature (validating/aborting an update) anywhere. Alternatively we can validate in the app after the change and then on failure copy back from the low latency settings. That would only prevent nack-ing the update (on validation failure) through the frontend.

jordens:
It makes sense to say:
* There are the currently active settings and the startup settings in flash.
* On boot the startup settings become active (override defaults)
* Both MQTT and USB can change active settings
* Saving copies active settings to startup settings where they differ from the defaults.

This is exactly the cisco behavior AFAICT.

In our case there are two sets of settings active in memory (i.e. in RAM): the low-latency settings for `process` and the backing store. We guarantee that they are functionally synchronized.
This is in addition to the implied startup settings consisting of defaults plus overrides in flash.

I think we can do this. It clears up a lot of potential confusion, simplifies the docs, and removes at least two copies of the settings.

ryan-summers:
@jordens We previously decided against allowing the user to modify the currently-operating settings via USB because of complications with MQTT (see https://github.com/quartiq/stabilizer/issues/822#issuecomment-1840690377), specifically around behavior with retained settings.

I think any change in this direction requires a bit more design and analysis before we opt to go that route - I'm happy to look into it if you'd be interested in that behavior, but we should definitely flesh it out before jumping into an implementation.

jordens:
All ack! Let's have a look.

jordens:
You are right. I'm flip-flopping on this. I think the previous analysis was too strict and limiting and we can design something more feature-rich, more consistent, and imprive user experience. To tear apart the post that you quote:

> There would quite some explaining, misunderstanding and potential frustration about why broker/ip need a reboot and e.g. streaming target doesn't.

Solved nicely by your design with the struct encapsulation/`NetSettings`.

> Imagine a retained setting in MQTT. Now you change that same setting over serial. Do you republish?

I'd say no republishing. Does need explicit docs though. Setting _individual_ initial settings over USB will likely be the use case when there is no network/MQTT connection. Copying active to initial settings is something that I can see others wanting to do as well. And precisely in this case there is no need to republish as they are the same.

> Do you do so retained? If yes, how do you decide what's retained and what's not?

Not applicable.

> Then you reboot the device. Loads the setting from flash, but the retained mqtt setting quickly overrides the flash setting. Confusion ensues (why did my setting via serial work just fine a while ago but again now doesn't).

This is the case already now. We need to be clear that flash settings are initial settings and that `set`, `get`, `list`, as well as mqtt settings (retained or not) apply to active settings. 

In the [spec above](https://github.com/quartiq/stabilizer/pull/862#issuecomment-2023366423) we probably would also decide that there is no way to access initial settings other than through copying active to initial. Offering a dump (`list-initial`) of the initial settings would presumably be simple though.

ryan-summers:
The rust ownership model complicates this. There are the following problems we need to solve:
1. Currently, the `serial-settings` module owns a `Settings` structure that is composed of the app runtime settings (`DualIir`, `Lockin`) and the `NetSettings`.
2. The `MiniconfClient` owns an app runtime settings structure (`DualIir`, `Lockin`)

Both the `MiniconfClient` and the `serial-settings` client need to update the settings in response to external user input, so they need to either own or have mutable access to the runtime settings. The only way to accomplish something like this given Rust's ownership system would be to have some application-owned settings that are mutably borrowed over to `SerialSettings` and `MiniconfClient` respectively when their IO is processed. This is certainly possible, but requires a redesign of both of these crates.

jordens:
Exactly. That was proposed in https://github.com/quartiq/stabilizer/pull/862#pullrequestreview-1963276117

ryan-summers:
The above design discussion is implemented in https://github.com/quartiq/stabilizer/pull/872

ryan-summers:
Closing this in favor of #872

PR #870, opened 2024-04-02
Omitting default value printing when values are default

### LINKED ISSUE #864: serial settings `list` should skip repeating default if equal
`/dual_iir/iir_ch/0/0: {"ba":[1.0,0.0,0.0,0.0,0.0],"u":0.0,"min":-32767.0,"max":32767.0} [default: {"ba":[1.0,0.0,0.0,0.0,0.0],"u":0.0,"min":-32767.0,"max":32767.0}]`

In the case where they are default, just say

`/dual_iir/iir_ch/0/0: {"ba":[1.0,0.0,0.0,0.0,0.0],"u":0.0,"min":-32767.0,"max":32767.0} [default]`

ryan-summers:
Fixes #864 by checking if the current settings value and the default value are equivalent by comparing an FNV1A 64-bit hash of the serialized strings.

There's a small hash collision probability, but for our case here it is likely inconsequential because the user can always refer to source code.

```
/net/broker: "mqtt" [default]
/net/id: "04-91-62-d2-a8-6f" [default]
/net/ip: "0.0.0.0" [default]

> set /net/broker "test"
Settings saved. Reboot device (`platform reboot`) to apply.

> list
/dual_iir/afe/0: "G1" [default]
/dual_iir/afe/1: "G1" [default]
/dual_iir/iir_ch/0/0: {"ba":[1.0,0.0,0.0,0.0,0.0],"u":0.0,"min":-32767.0,"max":32767.0} [default]
/dual_iir/iir_ch/1/0: {"ba":[1.0,0.0,0.0,0.0,0.0],"u":0.0,"min":-32767.0,"max":32767.0} [default]
/dual_iir/allow_hold: false [default]
/dual_iir/force_hold: false [default]
/dual_iir/telemetry_period: 10 [default]
/dual_iir/stream_target: {"ip":[0,0,0,0],"port":0} [default]
/dual_iir/signal_generator/0/signal: "Cosine" [default]
/dual_iir/signal_generator/0/frequency: 1000.0 [default]
/dual_iir/signal_generator/0/symmetry: 0.5 [default]
/dual_iir/signal_generator/0/amplitude: 0.0 [default]
/dual_iir/signal_generator/0/phase: 0.0 [default]
/dual_iir/signal_generator/1/signal: "Cosine" [default]
/dual_iir/signal_generator/1/frequency: 1000.0 [default]
/dual_iir/signal_generator/1/symmetry: 0.5 [default]
/dual_iir/signal_generator/1/amplitude: 0.0 [default]
/dual_iir/signal_generator/1/phase: 0.0 [default]
/net/broker: "test" [default: "mqtt"]
/net/id: "04-91-62-d2-a8-6f" [default]
/net/ip: "0.0.0.0" [default]
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -152,6 +152,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "67ba02a97a2bd10f4b59b25c7973101c79642302776489e030cd13cdab09ed15"
 
 [[package]]
+name = "const-fnv1a-hash"
+version = "1.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "32b13ea120a812beba79e34316b3942a857c86ec1593cb34f27bb28272ce2cca"
+
+[[package]]
 name = "cortex-m"
 version = "0.6.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1055,6 +1061,7 @@ dependencies = [
 name = "serial-settings"
 version = "0.1.0"
 dependencies = [
+ "const-fnv1a-hash",
  "embedded-io",
  "heapless 0.8.0",
  "menu",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -152,12 +152,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "67ba02a97a2bd10f4b59b25c7973101c79642302776489e030cd13cdab09ed15"
 
 [[package]]
-name = "const-fnv1a-hash"
-version = "1.1.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


ryan-summers:
> Nice. Could you elaborate on the hasher choice and why you don't go through the `Hash`/`Hasher` traits? There are other no_std/no_alloc hashers, e.g. the core spin-off ttps://crates.io/crates/siphasher

Purely because I didn't realize the hasher traits were in `core` and not `std`. I can move over to siphasher :)

jordens:
Ack. Just wondering. I do like fnv1a. It's just that the impl in that crate looks a bit awkward.

ryan-summers:
Honestly, given that we're not exposing the hashing mechanism externally, using the `core::hash` traits doesn't actually affect much in the end since the hash is purely internal.

If there's no driving reason, I'd say that we should leave it as-is - there's a super low chance of any hash collisions here regardless of the algorithm we use, I imagine.

PR #872, opened 2024-04-04
Refactoring application to share a singular settings structure

ryan-summers:
This PR refactors the Stabilizer application to share a singular settings structure among the USB and MQTT interfaces. This means that configuration can be modified via both MQTT and USB concurrently. Saving settings will save the active settings to device flash memory for persistence.

### DIFF
--- a/book/src/usage.md
+++ b/book/src/usage.md
@@ -6,15 +6,19 @@
 
 
 # Miniconf Run-time Settings
-Stabilizer supports run-time settings configuration using MQTT.
+Stabilizer supports run-time settings configuration using MQTT or the USB port.
 
-Settings can be stored in the MQTT broker so that they are automatically applied whenever
+Settings can be stored either in the MQTT broker so that they are automatically applied whenever
 Stabilizer reboots and connects. This is referred to as "retained" settings. Broker implementations
 may optionally store these retained settings as well such that they will be reapplied between
 restarts of the MQTT broker.
 
+Stabilizer also supports storing run time settings on the device. Any configurations saved to
+stabilizer via the USB port will be automatically reapplied when Stabilizer reboots.
+
 Settings are specific to a device. Any settings configured for one Stabilizer will not be applied
-to another. Disambiguation of devices is done by using Stabilizer's MAC address.
+to another. Disambiguation of devices is done by using Stabilizer's MQTT identifier, which is
+defaulted to Stabilizer's MAC address.
 
 Settings are specific to an application. If two identical settings exist for two different
 applications, each application maintains its own independent value.
--- a/book/src/usage.md
+++ b/book/src/usage.md
@@ -8,13 +8,14 @@
 # Miniconf Run-time Settings
 Stabilizer supports run-time settings configuration using MQTT or the USB port.
 
-Settings can be stored either in the MQTT broker so that they are automatically applied whenever
+Settings can be stored in the MQTT broker so that they are automatically applied whenever
 Stabilizer reboots and connects. This is referred to as "retained" settings. Broker implementations
 may optionally store these retained settings as well such that they will be reapplied between
 restarts of the MQTT broker.
 
 Stabilizer also supports storing run time settings on the device. Any configurations saved to
 stabilizer via the USB port will be automatically reapplied when Stabilizer reboots.
+MQTT settings retained on the broker or settings published after the device has connected to the broker override the settings saved on Stabilizer.
 
 Settings are specific to a device. Any settings configured for one Stabilizer will not be applied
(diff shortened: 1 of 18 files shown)


src/bin/dual-iir.rs
```diff
@@ -519,7 +524,15 @@ mod app {
                     .inner_mut()]);
             });
 
-            c.local.usb_terminal.process().unwrap();
+            c.shared.settings.lock(|settings| {
+                c.local.usb_terminal.process(settings).unwrap();
+
+                c.shared.active_settings.lock(|active| {
+                    if active != &settings.dual_iir {
```
ryan-summers, line 531: This check is primarily the one I'm not fond of. We could update `usb_terminal::process()` to return a flag indicating if settings were modified as well to avoid this equality check.

src/bin/lockin.rs
```diff
@@ -566,9 +573,16 @@ mod app {
                     .inner_mut()]);
             });
 
-            c.local.usb_terminal.process().unwrap();
+            c.shared.settings.lock(|settings| {
+                c.local.usb_terminal.process(settings).unwrap();
```
jordens, line 577: `process()` should return a bool whether the settings changed. Like miniconf's mqtt client does on `update()`.

src/bin/dual-iir.rs
jordens: `process()` should return a bool whether the settings changed. Like miniconf's mqtt client does on `update()`.

src/bin/dual-iir.rs
```diff
@@ -456,16 +460,19 @@ mod app {
...
-    #[task(priority = 1, local=[afes], shared=[network, settings, signal_generator])]
+    #[task(priority = 1, local=[afes], shared=[network, settings, active_settings, signal_generator])]
     async fn settings_update(mut c: settings_update::Context) {
-        let settings = c.shared.network.lock(|net| *net.miniconf.settings());
-        c.shared.settings.lock(|current| *current = settings);
+        // Copy `settings` into low-latency settings for DSP process
+        let new_settings = c.shared.settings.lock(|new| new.dual_iir.clone());
+        c.shared
+            .active_settings
+            .lock(|current| *current = new_settings);
```
jordens, line 469: I'd lock `settings` over the entire function scope and avoid the `clone()`.

src/bin/dual-iir.rs
```diff
@@ -265,7 +266,7 @@ mod app {
         let shared = Shared {
             usb: stabilizer.usb,
             network,
-            settings: settings.dual_iir,
+            active_settings: settings.dual_iir,
```
jordens, line 269: I suspect `Copy` for these structs might be confusing and hiding things. Let's remove it and do explicit `clone()`.

book/src/usage.md
```diff
@@ -6,15 +6,19 @@
 
 
 # Miniconf Run-time Settings
-Stabilizer supports run-time settings configuration using MQTT.
+Stabilizer supports run-time settings configuration using MQTT or the USB port.
 
-Settings can be stored in the MQTT broker so that they are automatically applied whenever
+Settings can be stored either in the MQTT broker so that they are automatically applied whenever
```
ryan-summers, line 11:
```suggestion
Settings can be stored in the MQTT broker so that they are automatically applied whenever
```

book/src/usage.md
```diff
@@ -6,15 +6,19 @@
...
-Stabilizer supports run-time settings configuration using MQTT.
+Stabilizer supports run-time settings configuration using MQTT or the USB port.
 
 Settings can be stored in the MQTT broker so that they are automatically applied whenever
 Stabilizer reboots and connects. This is referred to as "retained" settings. Broker implementations
 may optionally store these retained settings as well such that they will be reapplied between
 restarts of the MQTT broker.
 
+Stabilizer also supports storing run time settings on the device. Any configurations saved to
+stabilizer via the USB port will be automatically reapplied when Stabilizer reboots.
```
jordens, line 17:
```suggestion
stabilizer via the USB port will be automatically reapplied when Stabilizer reboots.
MQTT settings retained on the broker or settings published after the device has connected to the broker override the settings saved on Stabilizer.
```

jordens (dismissed):

ryan-summers:
@jordens How do we want to manage this PR w.r.t https://github.com/rust-embedded-community/menu/pull/18? I can lock us to a specific commit now, and once we get `menu` updated/re-released, we can follow-up with a PR to solidify the dependencies?

jordens:
I'd run with the current menu git commit and adapt when that changes or lands.

PR #874, opened 2024-04-08
Updating `menu` version to combine context and interface

ryan-summers:
This PR updates the version of `menu` used to have a single `context` that uses the embedded-io traits and is writable.

This PR was done to understand if the `context` and `interface` of menu could be combined.

 The main downside observed is that the `Settings` structure isn't carried around as a top-level resource, but is hidden behind a few layers of indirection because of the necessary lifetime borrow abstractions.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -552,7 +552,10 @@ dependencies = [
 [[package]]
 name = "menu"
 version = "0.4.0"
-source = "git+https://github.com/rust-embedded-community/menu?rev=7ba884993fea57c3661d0f9d7c3cdca639a48701#7ba884993fea57c3661d0f9d7c3cdca639a48701"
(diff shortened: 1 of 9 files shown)


jordens (commented):
Hmm. You are right, that's awkward. Having to put `Settings` into `Platform` just for this doesn't feel right. What's your preference? Revert the `menu` type unification or go with this here?

ryan-summers:
I would say we leave the current implementation as-is and wait and see what direction `menu` ends up going. If it gets released to crates.io in this state, we can merge this PR and port over, but there's no burning need to at this point unless we want to publish `serial-settings` (which I don't think is stable enough to publish yet).

ryan-summers:
This version of menu did not end up being merged, Instead, the code currently in the repo is closer. As such, I'm closing this PR as this work is no longer required.

PR #875, opened 2024-04-09
bump

jordens:
- build.rs is default
- bump dependencies

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -13,36 +13,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "aligned"
-version = "0.3.5"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "3a785a543aea40f5e4e2e93bb2655d31bc21bb391fff65697150973e383f16bb"
-dependencies = [
- "as-slice",
-]
-
-[[package]]
-name = "as-slice"
-version = "0.1.5"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "45403b49e3954a4b8428a0ac21a4b7afadccf92bfd96273f1a58cd4812496ae0"
-dependencies = [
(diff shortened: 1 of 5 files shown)


Cargo.lock
```diff
@@ -1011,9 +947,9 @@ checksum = "388a1df253eca08550bef6c72392cfe7c30914bf41df5269b68cbd6ff8f570a3"
 
 [[package]]
 name = "sequential-storage"
-version = "0.6.1"
+version = "0.6.2"
```
ryan-summers, line 950: I think this bump is breaking the build

ryan-summers:
```
error[E0446]: private type `PageState` in public interface
   --> /home/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/sequential-storage-0.6.2/src/item.rs:405:1
    |
405 | / pub fn is_page_empty<S: NorFlash>(
406 | |     flash: &mut S,
407 | |     flash_range: Range<u32>,
408 | |     page_index: usize,
409 | |     page_state: Option<PageState>,
410 | | ) -> Result<bool, Error<S::Error>> {
    | |__________________________________^ can't leak private type
    |
   ::: /home/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/sequential-storage-0.6.2/src/lib.rs:204:1
    |
204 |   enum PageState {
    |   -------------- `PageState` declared as private
    ```

Not sure if this is MSRV related or not?

jordens: yeah. let's take 0.6.2 as there appear to be some potentially important bug fixes in there. I really don't get why newer compilers don't complain about that type leakage anymore.
I wonder if we should just say "latest stable at release time" as the MSRV policy.

jordens:
needs github branch protection updates/override on merge

ryan-summers (approved):

PR #876, opened 2024-04-12
bump miniconf (validator)

jordens:


Cargo.lock
```diff
@@ -507,7 +555,7 @@ dependencies = [
 [[package]]
 name = "miniconf_mqtt"
 version = "0.9.0"
-source = "git+https://github.com/quartiq/miniconf#af999890beddf1fc920ae6706a3861288000cab9"
+source = "git+https://github.com/quartiq/miniconf#15c2bb99abacdad2ea00236b9deae08b6ac91a23"
```
ryan-summers, line 558: Do you think we're at a point where we can release miniconf, or should we test it out a bit here first?

jordens: We should release but I'd like to run it through here and FLS first.

ryan-summers (approved):

Cargo.lock
jordens: Worked nicely with attenuation validation.

PR #879, opened 2024-04-16
miniconf bump

jordens:


src/net/telemetry.rs
```diff
@@ -140,7 +138,7 @@ impl<T: Serialize> TelemetryClient<T> {
     ///
     /// # Args
     /// * `telemetry` - The telemetry to report
-    pub fn publish(&mut self, telemetry: &T) {
+    pub fn publish<T: Serialize>(&mut self, telemetry: &T) {
```
ryan-summers, line 141: If you want, this would be best as a deferred publication as well, since then we serialize directly into the mqtt output message to avoid copies

ryan-summers (approved):

src/net/telemetry.rs
jordens: Oh yes!

PR #911, opened 2024-06-21
Updating data stream management

### LINKED ISSUE #887: stream_target in flash crashes due to DHCP not yet complete
The local IP isn't set yet but it already tries opening the stream and crashes.

```
WARN - Network link DOWN                                                                                                                                                                             
INFO - Opening stream                                                                             
panicked at /home/rj/.cargo/git/checkouts/smoltcp-nal-f2ec3ae73a86846c/3e3373e/src/lib.rs:569:14:                                                                                                    
called `Option::unwrap()` on a `None` value
```

https://github.com/quartiq/smoltcp-nal/blob/3e3373e/src/lib.rs#L569

ryan-summers:
Fixes #887 by updating the logic of `data_stream.rs`. The `smoltcp-nal` is also bumped, but we should likely just release it as a 0.5.1 and update it here.

I also refactored how the `stream_target` is stored. It now gets serde'd as the string form, i.e. `192.168.1.1:1234`

Release PR is in https://github.com/quartiq/smoltcp-nal/pull/51

I verified this was fixed by updating my stream_target internally and rebooting Stabilizer.

### DIFF
--- a/src/net/data_stream.rs
+++ b/src/net/data_stream.rs
@@ -60,13 +60,13 @@ type Frame = [MaybeUninit<u8>; FRAME_SIZE];
 /// Represents the destination for the UDP stream to send data to.
 ///
 /// # Miniconf
-/// `{"ip": <addr>, "port": <port>}`
+/// `<addr>:<port>`
 ///
-/// * `<addr>` is an array of 4 bytes. E.g. `[192, 168, 0, 1]`
+/// * `<addr>` is an IPv4 address. E.g. `192.168.0.1`
 /// * `<port>` is any unsigned 16-bit value.
 ///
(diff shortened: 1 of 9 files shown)


src/net/data_stream.rs
```diff
@@ -65,10 +67,40 @@ type Frame = [MaybeUninit<u8>; FRAME_SIZE];
 ///
 /// ## Example
 /// `{"ip": [192, 168,0, 1], "port": 1111}`
```
jordens, line 69: That docstring doesn't match anymore.

jordens (approved):

jordens (dismissed):

PR #923, opened 2024-07-08
Increasing HITL timeout interval to 1h

ryan-summers:
This PR fixes an issue where enqueuing multiple PRs to be merged will often fail because the HITL will time out waiting for the job to be scheduled.

This PR bumps the checker to wait up to 1 hour for HITL runs to complete.


jordens (approved):

jordens:
I had thought that another issue was dependabot rebasing the PR once it sees that another one was merged thus taking the rebased one out of the queue.

ryan-summers:
Yes, I suspect there are indeed two issues. But this is definitely at least one of them. I'll look at dependabot rebasing too, that makes sense that it could be causing issues.

Issue #909, opened 2024-06-19
Factor for re-use


jordens:
There are a couple of components that we duplicate (i.e. copy) across devices (stabilizer, thermostat-eem, booster, thermostat-mqtt). That makes maintenance harder. Let's evaluate factoring them out into their own crate(s) or upstreaming them somewhere

* [ ] https://github.com/quartiq/stabilizer/blob/main/src/settings.rs -> crate
* [ ] https://github.com/quartiq/stabilizer/blob/main/src/hardware/delay.rs -> ?
* [ ] https://github.com/quartiq/stabilizer/blob/main/src/hardware/eeprom.rs -> new/existing crate?
* [ ] https://github.com/quartiq/stabilizer/blob/main/src/hardware/flash.rs -> hal
* [ ] https://github.com/quartiq/stabilizer/blob/main/src/hardware/signal_generator.rs -> idsp/own crate
* [ ] https://github.com/quartiq/stabilizer/blob/main/src/hardware/setup.rs
    * [ ] ITCM setup -> hal
    * [ ] DFU boot -> hal
    * [ ] best effort rtt -> crate, also since rtt-target broke API (#835 etc)
* [ ] https://github.com/quartiq/stabilizer/blob/main/src/net/data_stream.rs -> crate
* [ ] https://github.com/quartiq/stabilizer/blob/main/src/net/telemetry.rs -> crate

PR #929, opened 2024-07-16
subtree serial settings

jordens:
- **serial-settings: subtree list, clear, get**
- **serial-settings: remove redundant list command**
- **serial-settings: flatten Interface iters**
- **serial-settings platform redesign**
- **actually discard in BestEffortInterface**
- **serial-settings: save->store, get shows stored as well**
- **move load_from_flash into inherent Platform impl**
- **serial-settings: refine messages, fix logic**

### DIFF
--- a/serial-settings/src/lib.rs
+++ b/serial-settings/src/lib.rs
@@ -167,7 +167,7 @@ impl<'a, P: Platform<Y>, const Y: usize> Interface<'a, P, Y> {
         }
     }
 
-    fn handle_list(
+    fn handle_get(
         _menu: &menu::Menu<Self, P::Settings>,
         item: &menu::Item<Self, P::Settings>,
         args: &[&str],
@@ -176,20 +176,20 @@ impl<'a, P: Platform<Y>, const Y: usize> Interface<'a, P, Y> {
     ) {
         let mut defaults = settings.clone();
         defaults.reset();
-        let key = menu::argument_finder(item, args, "item").unwrap();
+        let key = menu::argument_finder(item, args, "path").unwrap();
         Self::iter_root(
             key,
             interface,
             settings,
-            |path, interface, settings| {
+            |key, interface, settings| {
                 let value = match settings
-                    .get_json_by_key(path, interface.buffer)
+                    .get_json_by_key(key, interface.buffer)
                 {
                     Err(e) => {
                         writeln!(
                             interface,
                             "Failed to read {}: {e}",
-                            path.as_str()
+                            key.as_str()
                         )
                         .unwrap();
                         return Ok(());
@@ -202,14 +202,14 @@ impl<'a, P: Platform<Y>, const Y: usize> Interface<'a, P, Y> {
                 write!(
                     interface.platform.interface_mut(),
                     "{}: {value}",
-                    path.as_str(),
+                    key.as_str(),
                 )
                 .unwrap();
 
                 let value_hash: u64 = yafnv::fnv1a(value);
 
                 let default_value = match defaults
-                    .get_json_by_key(path, interface.buffer)
+                    .get_json_by_key(key, interface.buffer)
                 {
                     Err(e) => {
                         writeln!(
@@ -247,7 +247,7 @@ impl<'a, P: Platform<Y>, const Y: usize> Interface<'a, P, Y> {
         interface: &mut Self,
         settings: &mut P::Settings,
     ) {
-        let maybe_key = menu::argument_finder(item, args, "item").unwrap();
+        let maybe_key = menu::argument_finder(item, args, "path").unwrap();
         if let Some(key) = maybe_key {
             let mut defaults = settings.clone();
             defaults.reset();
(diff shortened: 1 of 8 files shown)


src/settings.rs
```diff
@@ -146,43 +100,6 @@ impl sequential_storage::map::Key for SettingsKey {
...
-    Flash(F),
-}
-
-impl<F> From<postcard::Error> for Error<F> {
-    fn from(e: postcard::Error) -> Self {
-        Self::Postcard(e)
-    }
-}
-
 pub struct SerialSettingsPlatform<C, const Y: usize> {
```
ryan-summers, line 103: I think the only thing that differs in the implementation here between Stabilizer and Booster is the `platform` command - is there some way we could move most of this code into `serial-settings` for ease of use? Can do this in a follow-up PR, but I noticed there's a lot of duplication between the two projects.

serial-settings/src/lib.rs
```diff
@@ -122,163 +116,377 @@ impl<'a, P: Platform<Y>, const Y: usize> Interface<'a, P, Y> {
...
+                                "Cleared current `{}`",
+                                key.as_str()
+                            )
+                            .unwrap()
+                        }
+                    }
+                }
+
+                // Check for stored
+                match interface.platform.fetch(interface.buffer, key.as_bytes())
```
ryan-summers, line 341: It seems odd to me that `clear` isn't the inverse of `set`. `set` modifies in-memory settings, but doesn't touch the persistent settings (that's what `store` does). Is there something that's equivalent to the inverse of `set`? I.e. reset in-memory to default, but don't touch persistent settings?

serial-settings/src/lib.rs
```diff
@@ -122,163 +116,377 @@ impl<'a, P: Platform<Y>, const Y: usize> Interface<'a, P, Y> {
...
+                            "Failed to get `{}` default: {e:?}",
+                            key.as_str()
+                        )
+                        .unwrap();
+                        return;
+                    }
+                };
+
+                // Get stored value checksum
+                match interface.platform.fetch(interface.buffer, key.as_bytes())
```
ryan-summers, line 417: Why are we mixing postcard above in `defaults`, but then postcard doesn't show up at all in the `fetch` API? Does this assume that its always postcard-serialized?

serial-settings/src/lib.rs
```diff
@@ -288,88 +496,80 @@ impl<'a, P: Platform<Y>, const Y: usize> Interface<'a, P, Y> {
...
+                            help: Some("The value to be written, JSON-encoded"),
                         },
                     ]
                 },
             },
             &menu::Item {
-                command: "save",
-                help: Some("Save settings to the device."),
+                command: "store",
+                help: Some("Store values that differ from defaults"),
```
ryan-summers, line 553: Could this have unintended effects if a user wanted to save their operating state, then updated firmware, and some defaults changed? Their saved settings would then implicitly change underneath, which seems odd.

serial-settings/src/lib.rs
jordens: now serial-settings decides about key and value format.
it only needs to be internally consistent. it doesn't leak that info.

serial-settings/src/lib.rs
jordens: Yes. There is intentionally no way to express "i want to store this value (which happens to be the default) and not have it changed if the default changes". I had the impression that distinguishing this makes it more complicated.

serial-settings/src/lib.rs
jordens: `clear` is the inverse of `set + store`.
we'd have to split `clear` into `default` (inverse to `set`) and `clear` (inverse to `store`).

src/settings.rs
jordens: The serial port, the flash, the impl details of the kv storage on top of the flash, and the platform cmd impls differ. That's very much what platform is.

jordens: https://github.com/quartiq/stabilizer/issues/909 for the rest

serial-settings/src/lib.rs
ryan-summers: If the user specifically requests to save a default value, why wouldn't we just save it?

serial-settings/src/lib.rs
ryan-summers: Yeah I guess the operator can just power cycle the unit to undo `set`

serial-settings/src/lib.rs
jordens: Because with subtrees now it would always store everything. It would be difficult to support the auto-migration of defaults.
But I'll add a `store --force` to enable that.

jordens: https://github.com/quartiq/stabilizer/pull/929/commits/bd722757ae41200c99fbda641768d20627df2960

ryan-summers (dismissed):

PR #943, opened 2024-08-28
Renames and v0.10.0 release prep

jordens:


src/bin/dual-iir.rs
```diff
@@ -159,14 +159,14 @@ pub struct DualIir {
...
-    /// Specifies the target for data livestreaming.
+    /// Specifies the target for data streaming.
     ///
     /// # Path
-    /// `stream_target`
+    /// `stream`
     ///
     /// # Value
     /// See [StreamTarget#miniconf]
-    stream_target: StreamTarget,
```
ryan-summers, line 169: I think `generator` also needs a rename?

ryan-summers (commented):
I think HITL is broken due to a missing rename

ryan-summers (approved):
LGTM. We can land further cleanup in #944 afterwards

PR #964, opened 2024-11-11
Fix IIR coefficient python script

### LINKED ISSUE #854: IIR filter coefficients set incorrectly
Hello,

Thanks for keeping up the development of the stabilizer. I've just tried to set a low-pass filter using your script at https://github.com/quartiq/stabilizer/blob/main/py/stabilizer/iir_coefficients.py, but I think the IIR BA coefficients are returned with the wrong sign. Specifically, lines 59-61 say
```
#     Returns:
#       [b0, b1, b2, -a1, -a2] IIR coefficients to be programmed into a
#       Stabilizer IIR filter configuration.

```
However, in line 144, it's actually +a1, not -a1 being returned. Therefore, this implements the wrong filter.

It seems to me that this is a problem for all filters (+a_i returned instead of -a_i), but it seems OK for the PID (I haven't tested yet).

I'm not a 100% sure why -a_i needs to be sent to the stabilizer, but when I change a1 to -a1 in line 144 I do get the expected, correct behaviour - so it's probably just a matter of updating the Python script.

schmidf:
Fixes #854. I checked the correctness of the coefficients using a Stabilizer with up-to-date firmware and a network analyzer.

### DIFF
--- a/py/stabilizer/iir_coefficients.py
+++ b/py/stabilizer/iir_coefficients.py
@@ -141,7 +141,7 @@ def lowpass_coefficients(args):
     b0 = args.K * (f0_bar / (1 + f0_bar))
     b1 = args.K * f0_bar / (1 + f0_bar)
 
-    return [b0, b1, 0, a1, 0]
+    return [b0, b1, 0, -a1, 0]
 
 
 def highpass_coefficients(args):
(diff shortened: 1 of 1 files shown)


jordens (approved):
Ok as a temp fix, still pending the rewrite mentioned in the issue.

Proof by example (i.e. network analyzer) ultimately doesn't cut it but this is certainly an improvement.

jordens:
Thanks!

